PR #30, opened 2026-02-26
Add Chinese (Simplified) localization

Kobe9312:
- Translate all UI text to Chinese
- Add Chinese README
- Add Chinese notification categorization docs
- Include menu items, dialogs, and window titles


Kobe9312:
Hi! I have addressed all the review comments:

- Reverted implementation comments back to English (keeping only UI strings localized)
- Fixed ProviderSummaryText fallback from n/a to 无 in the code-behind
- Localized on/off display strings to 开/关 in menu items while keeping action payload values stable

Please take another look, thanks!

PR #80, opened 2026-03-20
fix: stabilize tray pairing and reconnect behavior

andyeskridge:
# Summary
This updates the Windows tray node pairing flow to match current OpenClaw gateway behavior and fixes reconnect recovery after gateway restarts.

## Why
The tray previously assumed a node was only paired if `hello-ok` returned `auth.deviceToken` and that token was persisted locally.

In practice, current gateway behavior for Windows WS nodes looks like this:

- unapproved device -> `NOT_PAIRED`
- approved device -> `hello-ok`
- node commands work normally after approval
- `auth.deviceToken` is not always returned
- That left the tray in incorrect states like false `Pending`, false `Unknown`, and failed reconnect recovery after gateway restarts.

## Changes

- treat `NOT_PAIRED` as a real pairing-needed state instead of a generic registration error
- only show pending approval when the gateway explicitly indicates pairing is required
- treat an approved `hello-ok` as paired even when `auth.deviceToken` is absent
- fix websocket reconnect so the tray keeps retrying until the gateway is back
- suppress repeated pairing/connect notifications and duplicate activity noise on routine reconnects
- tighten pairing event matching so node pairing events use the actual node identity when available

## Validation
Tested locally against a real gateway:

- fresh tray state enters `Pending` on `NOT_PAIRED`
- approving the device in OpenClaw transitions the node to `Connected`
- node requests succeed after approval (`system.notify` verified)
- gateway restart disconnects the tray, retries through startup errors, and reconnects automatically

### DIFF
--- a/src/OpenClaw.Shared/WebSocketClientBase.cs
+++ b/src/OpenClaw.Shared/WebSocketClientBase.cs
@@ -186,12 +186,18 @@ public abstract class WebSocketClientBase : IDisposable
 
     protected async Task ReconnectWithBackoffAsync()
     {
+        var emittedConnecting = false;
+
         while (!_disposed)
         {
             var delay = BackoffMs[Math.Min(_reconnectAttempts, BackoffMs.Length - 1)];
             _reconnectAttempts++;
             _logger.Warn($"{ClientRole} reconnecting in {delay}ms (attempt {_reconnectAttempts})");
-            RaiseStatusChanged(ConnectionStatus.Connecting);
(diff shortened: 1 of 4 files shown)


andyeskridge:
Pushed some polish to address the feedback.

andyeskridge:
No problem @shanselman! I'm glad you were able to get some smaller fixes in already. I'll push up a draft with the remaining diffs from this branch based on current master.

PR #159, opened 2026-04-08
feat: add screen recording capability

AlexAlves87:
## Summary
Adds `screen.record`, `screen.record.start`, and `screen.record.stop` to `ScreenCapability`. `screen.record` does a fixed-duration capture and returns a base64 MP4. The start/stop pair allows session-based recording with a `recordingId` so the caller controls when to stop.

## What changed
- `ScreenCapability`: three new commands with events and arg parsing
- `ScreenRecordingService`: WinRT capture (D3D11CaptureFramePool), BGRA→NV12 conversion, MediaTranscoder with hardware and software fallback
- `NodeService`: wired to the new capability events

## Testing
- `./build.ps1`
- Manual: called `screen.record` from a connected agent, received valid MP4

## Notes
OpenClaw.Shared.Tests currently has 8 pre-existing failures on this branch related to culture-sensitive number formatting. These are unrelated to this change and are covered by a separate fix.


AlexAlves87:
Addressing the Repo Assist observations — all three issues are resolved in commit f4dbc52:

**🚨 BuildRawVideoStream .Wait()** — Fixed. `BuildRawVideoStream` no longer exists. The entire encoding pipeline was replaced with `MediaStreamSource` + `PrepareMediaStreamSourceTranscodeAsync`, which also fixed the root cause of the `Transcode failed: Unknown` error that was blocking screen recording entirely.

**⚠️ Memory footprint** — Fixed. Added `MaxFrameBufferBytes = 500 MB` constant with an early-stop guard in both capture loops (`RecordAsync` and `ActiveSession.RunAsync`). When the cap is hit, capture stops gracefully and a warning is logged. Limits documented in XML docs on `ScreenRecordArgs` and `ScreenRecordStartArgs`.

**🧹 Fps float → int** — Fixed. `ScreenRecordResult.Fps` is now `int`.

**🔴 Pre-existing culture test failures** — Resolved by syncing with upstream. Commit `6933239` (`fix: use invariant culture in numeric display formatting`, merged in #158) fixes those 8 failures. CI on this PR is now fully green.

shanselman:
Hi @AlexAlves87 — **Copilot here replying on Scott's behalf.**

Thanks for the work on this. The feature looks useful and the CI result is encouraging, but on a deeper engineering pass we found a few things we'd like tightened up before we can merge it.

1. **Handle the no-monitor / bad-index case explicitly** so this returns a controlled error instead of blowing up on monitors[screenIndex].

```
private static GraphicsCaptureItem CreateCaptureItem(int screenIndex)
{
    var monitors = GetMonitorHandles();
    if (monitors.Count == 0)
        throw new InvalidOperationException("No screens available");

    if (screenIndex < 0 || screenIndex >= monitors.Count)
        throw new ArgumentOutOfRangeException(nameof(screenIndex), "Invalid monitor index");

    ...
}
```

2. **Own and dispose the D3D device together with the session/pool.** Right now the device lifetime looks too loose for long-running capture sessions.

```
private readonly IDirect3DDevice _device;

public ActiveSession(..., IDirect3DDevice device, ...)
{
    _device = device;
}

public void Dispose()
{
    _cts.Cancel();
    try { _captureTask.GetAwaiter().GetResult(); } catch { }
    try { _session.Dispose(); } catch { }
    try { _pool.Dispose(); } catch { }
    try { _device.Dispose(); } catch { }
}
```

3. **Please stop/clear active recording sessions on node disconnect**, so an abandoned screen.record.start session can't keep holding memory/resources after the connection goes away.

4. **This one feels especially important:** dispose the per-frame SoftwareBitmap immediately after extracting bytes. At the current scale this can retain a *lot* of native memory during recordings.

```
using var bmp = await SoftwareBitmap.CreateCopyFromSurfaceAsync(frame.Surface);
frames.Add(ExtractBitmapBytes(bmp));
```

and similarly in the encoder:

```
using var dw = new DataWriter();
dw.WriteBytes(nv12);
var sample = MediaStreamSample.CreateFromBuffer(dw.DetachBuffer(), ts);
```

One other thing we'd like cleaned up: **temp MP4 retention.** Right now each recording writes a file under %TEMP%\openclaw\openclaw-screen-record-*.mp4 with no obvious cleanup policy. Even if we keep returning ilePath, we'd want those files expired/cleaned aggressively or made opt-in for privacy/hygiene.

```
private static void CleanupOldTempRecordings()
{
    var dir = Path.Combine(Path.GetTempPath(), "openclaw");
    if (!Directory.Exists(dir)) return;

    foreach (var file in Directory.EnumerateFiles(dir, "openclaw-screen-record-*.mp4"))
    {
        var info = new FileInfo(file);
        if (info.CreationTimeUtc < DateTime.UtcNow.AddHours(-24))
        {
            try { File.Delete(file); } catch (IOException) { }
            catch (UnauthorizedAccessException) { }
        }
    }
}
```

If you can tidy up those areas — especially **#4** — we'd be happy to take another look. Thanks again for pushing this forward.

AlexAlves87:
Hi @shanselman / @copilot — all five points are already addressed in the commits on this branch:

1. **No-monitor / bad-index** (`5267bb6`) — `CreateCaptureItem` now throws `InvalidOperationException` if no monitors found and `ArgumentOutOfRangeException` for an out-of-range index.
2. **D3D device lifetime** (`9008cb6`) — `IDirect3DDevice` is now owned by `ActiveSession` and disposed in both `RecordAsync`'s `finally` and `ActiveSession.Dispose()`.
3. **Stop sessions on disconnect** (`e272e9f`) — `NodeService.OnDisconnected` now calls `StopAllSessions()`, so no session survives a node drop.
4. **SoftwareBitmap disposal** (`8b69827`) — each frame bitmap is `using`-wrapped; `DataWriter` is also disposed immediately after writing.
5. **Temp MP4 cleanup** (`d962f4a`) — `CleanupOldTempRecordings()` runs on every new recording and deletes files older than 24h.

Let me know if you'd like anything squashed before merge.

AlexAlves87:
Updated this PR with `master` in `2ec81e4` and resolved the merge conflicts caused by upstream screen capability changes.

The branch now keeps the upstream `screen.snapshot` rename while preserving the recording work from this PR, plus the small follow-up fix needed to get the build back into shape.

Sorry for the confusion/delay here.

shanselman:
Thanks @AlexAlves87 for doing the screen recording work here. The Windows.Graphics.Capture + MP4 encoding implementation in this PR gave us the practical path for closing the Windows/Mac screen-recording parity gap.

I integrated this on `master` in a parity-safe form in dba5342 (`feat: add screen recording parity`) rather than merging this branch as-is, because the current Mac/gateway command surface supports exactly:

- `screen.snapshot`
- `screen.record`

So the final Windows implementation intentionally keeps only the fixed-duration `screen.record` API and does **not** advertise or support the extra non-Mac commands from this PR (`screen.list`, `screen.record.start`, `screen.record.stop`). That keeps Windows aligned with the Mac node and avoids expanding the gateway command surface beyond what upstream currently defines.

What landed:

- Added `screen.record` alongside existing `screen.snapshot` in `ScreenCapability`.
- Wired Windows node handling through a new `ScreenRecordingService` using Windows.Graphics.Capture and MP4/base64 output.
- Matched the Mac-compatible params/payload shape: `durationMs`, `fps`, `screenIndex`, `format: "mp4"`, `includeAudio`, and response `format`, `base64`, `durationMs`, `fps`, `screenIndex`, `hasAudio`.
- Rejected unsupported formats instead of accepting non-parity outputs.
- Removed the hidden temp-file persistence from the Windows service so recordings are returned to the gateway without leaving an extra local MP4 copy on disk.
- Hardened native interop failures so they surface as command errors instead of risking process crashes.
- Added tests locking the parity command surface: `screen.record` is supported, while `screen.capture`, `screen.list`, `screen.record.start`, and `screen.record.stop` remain unsupported.

Validation completed after integration:

- `./build.ps1`
- Live integration test through the Mac gateway CLI to the Windows node: `screen.record` succeeded for a 3s/10fps recording and returned MP4 metadata; the temporary Mac-side output file was deleted after validation.

Closing this PR because the useful implementation direction has been incorporated on `master` with the adjusted parity-safe command surface. Thanks again for the contribution.

shanselman:
Closing because the parity-safe screen.record implementation has landed on master in dba5342. Thanks again for the contribution!

AlexAlves87:
Thanks @shanselman, I really appreciate this.
I hadn’t seen these messages yet, so this was a nice surprise. Glad the recording work helped get screen.record landed on Windows in a parity-safe way.
Thanks again for integrating it

PR #216, opened 2026-04-26
feat: add a stable routing seam for system.run exec approvals

AlexAlves87:
I have a small, fully tested slice prepared for the exec approvals work. In practice, it only adds a stable routing seam for `system.run`, without changing the current production behavior.

The goal here is not to open a "half-finished" PR, but to sanity-check whether this area makes sense to evolve through small, well-scoped changes that can be reviewed and merged independently while keeping the current legacy path unchanged by default. If that direction is a good fit for the repo, it would also move Windows closer to the existing macOS reference.

Before pushing further in that direction, I wanted to put this first slice in front of you with real code and validate whether the approach itself makes sense here, or whether you'd prefer changes in exec approvals to stay within the current legacy flow.

## What this PR includes

- a typed result model for exec approval outcomes with 6 stable codes
- a UI-free handler seam for `system.run`
- a null handler that keeps the new seam inert by default
- the routing branch in `SystemCapability.HandleRunAsync`
- minimum observability at the routing boundary:
  - correlation ID
  - selected path
  - decision
  - reason code

## What does not change

- `system.run` still uses the current legacy path by default
- no production wiring activates the new seam in this PR
- this PR does not add validation, executable resolution, prompting, or persistence behavior yet

## Validation

- `./build.ps1`


shanselman:
Closing because this was manually rebased/cherry-picked onto master as 24a4c15, with required build and test validation passing.

AlexAlves87:
Thanks for cherry-picking the routing seam into `master`, @shanselman.

I put together and published the rest of the exec approvals design work that builds on that base: https://github.com/AlexAlves87/windows-exec-approvals-design-notes

It covers the architecture, an incremental implementation path, and the key design decisions for the Windows side. I'm sharing it in case it's useful context. I'd also appreciate a signal on whether the incremental approach described there fits the direction you want for this area, or whether you'd prefer a different path for any follow-up work.

PR #239, opened 2026-04-28
feat: native WinUI A2UI renderer + MCP/security hardening

codemonkeychris:
## Summary

Brings the Windows tray to feature parity with the Android client on A2UI v0.8 and works through a security review of the resulting attack surface.

- **Native WinUI A2UI renderer.** Replaces the Lit/WebView2 path with a XAML-native renderer hosting all 18 v0.8 components (Row, Column, List, Card, Tabs, Modal, Divider, Text, Image, Icon, Video, AudioPlayer, Button, CheckBox, TextField, DateTimeInput, MultipleChoice, Slider). Catalog-strict registry, JSONL envelope routing, A2UIValue tagged-union resolution, surface-scoped data model with subscriptions, action.context dataBinding scoping, secret-path redaction. SVG support added on top of D2D, with DOCTYPE strip + decode budgeting.
- **Local MCP server.** Bearer-token auth, loopback-only bind, Origin/Host/Content-Type CSRF gate, per-request deadline + handler concurrency cap with drain-on-Dispose. Black-box integration tests cover every advertised tool. Tray Settings surfaces the endpoint, masked token, and a reset button.
- **canvas.navigate routes to the OS default browser** (no longer requires an embedded WebView2 surface). High-risk URLs (loopback / private IP / link-local / no-dot intranet / non-HTTPS / IP literal / Windows Internet Security Manager Local-Machine or Intranet zone) require user confirmation via a Win32 MessageBox prompt; the agent always receives the same success response immediately and the prompt+launch run fire-and-forget so the user's decision and decision time aren't exposed via timing.
- **Media safety.** Image / Video / AudioPlayer route through MediaResolver with HTTPS-only allowlist, DNS resolution + non-public-IP rejection (defends against DNS rebinding), data: URL size cap, Content-Length and stream-side byte caps for image fetches, and per-image cancellation tokens.
- **jsonlPath TOCTOU.** canvas.a2ui.push jsonlPath rejects unresolved reparse points, opens the file once with FileStream, re-validates the resolved path with GetFinalPathNameByHandle, and enforces the size cap on the open stream.
- **Token sanitization.** TokenSanitizer redacts Authorization: Bearer, JSON token/secret/bearer fields, and bare 43-char base64url tokens from log + jsonl diagnostics + malformed-line logs.

All Win32/COM APIs in the URL security path (`MessageBoxW`, `IInternetSecurityManager::MapUrlToZone` via documented CLSID/IID) are public and shipped in the current Windows SDK.

## Test plan

- [x] `OpenClaw.Shared.Tests` — 1039 passed
- [x] `OpenClaw.Tray.Tests` — 219 passed
- [x] `OpenClaw.Tray.IntegrationTests` (`OPENCLAW_RUN_INTEGRATION=1`, real tray spawned per fixture) — 18 passed
- [x] `OpenClaw.Tray.UITests` — 62 passed
- [x] Manual: `canvas.navigate http://www.example.com` from MCP client — confirmation prompt appears, "Yes" launches default browser, "No" silently blocks, agent gets identical success in both cases
- [x] Manual: rich A2UI surface push round-trips through native renderer with `canvas.a2ui.dump` returning the redacted state snapshot
- [x] Manual: MCP token reset from Settings — old token invalidated, new token hot-swapped without restarting the listener

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,8 +48,9 @@ jobs:
     - name: Build Tests
       run: |
         dotnet build tests/OpenClaw.Shared.Tests -c Debug --no-restore
-        dotnet build tests/OpenClaw.Tray.Tests -c Debug --no-restore
-        dotnet build tests/OpenClaw.Tray.IntegrationTests -c Debug --no-restore
+        dotnet build tests/OpenClaw.Tray.Tests -c Debug -r win-x64 --no-restore
+        dotnet build tests/OpenClaw.Tray.IntegrationTests -c Debug -r win-x64 --no-restore
+        dotnet build tests/OpenClaw.Tray.UITests -c Debug -r win-x64 --no-restore
 
     - name: Run Shared Tests
       env:
@@ -68,10 +69,41 @@ jobs:
         dotnet test tests/OpenClaw.Tray.Tests
         --no-build
         -c Debug
+        -r win-x64
         --verbosity normal
         --collect:"XPlat Code Coverage"
         --results-directory TestResults\Tray
         --logger "trx;LogFileName=OpenClaw.Tray.Tests.trx"
+
+    # Tray integration tests gate on OPENCLAW_RUN_INTEGRATION; set it so the
+    # MCP-server / capability tests actually run. M3 (tool-name drift) and
+    # similar would have been caught here if these had ever been wired.
+    - name: Run Tray Integration Tests
+      env:
+        OPENCLAW_RUN_INTEGRATION: 1
+      run: >
+        dotnet test tests/OpenClaw.Tray.IntegrationTests
+        --no-build
+        -c Debug
+        -r win-x64
+        --verbosity normal
+        --collect:"XPlat Code Coverage"
+        --results-directory TestResults\TrayIntegration
+        --logger "trx;LogFileName=OpenClaw.Tray.IntegrationTests.trx"
+
+    # UI tests need a real visual tree; on the GitHub windows-latest runner this
+    # is a Window Station with WindowsAppSDK available, which is enough for
+    # WinUI 3 UI tests. Failures here block PRs.
+    - name: Run Tray UI Tests
+      run: >
+        dotnet test tests/OpenClaw.Tray.UITests
+        --no-build
+        -c Debug
+        -r win-x64
+        --verbosity normal
+        --collect:"XPlat Code Coverage"
+        --results-directory TestResults\TrayUI
+        --logger "trx;LogFileName=OpenClaw.Tray.UITests.trx"
 
     - name: Upload Test Results
       if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,7 @@ jobs:
       run: |
         dotnet build tests/OpenClaw.Shared.Tests -c Debug --no-restore
         dotnet build tests/OpenClaw.Tray.Tests -c Debug --no-restore
+        dotnet build tests/OpenClaw.Tray.IntegrationTests -c Debug --no-restore
 
     - name: Run Shared Tests
       env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,9 +105,20 @@ jobs:
         --results-directory TestResults\TrayIntegration
         --logger trx;LogFileName=OpenClaw.Tray.IntegrationTests.trx"
 
-    # UI tests need a real visual tree; on the GitHub windows-latest runner this
-    # is a Window Station with WindowsAppSDK available, which is enough for
-    # WinUI 3 UI tests. Failures here block PRs.
+    # UI tests need a real visual tree AND a system-registered WindowsAppRuntime
+    # framework MSIX — the test fixture calls Bootstrap.Initialize(1.8, stable),
+    # which looks up the framework package by identity. The hosted windows-2025
+    # runner image doesn't preinstall it, so we install it explicitly here.
+    # Version pinned to match Microsoft.WindowsAppSDK 1.8.260101001 in the csprojs.
+    - name: Install WindowsAppRuntime 1.8
+      shell: pwsh
+      run: |
+        $url = "https://aka.ms/windowsappsdk/1.8/1.8.260101001/windowsappruntimeinstall-x64.exe"
+        $exe = "$env:RUNNER_TEMP\WindowsAppRuntimeInstall.exe"
+        Invoke-WebRequest -Uri $url -OutFile $exe
+        & $exe --quiet
+        if ($LASTEXITCODE -ne 0) { throw "WindowsAppRuntimeInstall failed with exit code $LASTEXITCODE" }
+
     - name: Run Tray UI Tests
       run: >
         dotnet-coverage collect
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,13 @@ jobs:
     - name: Restore dependencies
       run: dotnet restore
 
+    # dotnet-coverage replaces coverlet because the integration tests spawn the
+    # tray exe out-of-process; coverlet only instruments the in-proc test
+    # assembly. Installing once at the job level lets every test step wrap its
+    # `dotnet test` invocation in `dotnet-coverage collect`.
+    - name: Install dotnet-coverage
+      run: dotnet tool install --global dotnet-coverage
+
     - name: Build Shared Library
       run: dotnet build src/OpenClaw.Shared -c Debug --no-restore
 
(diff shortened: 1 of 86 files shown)


codemonkeychris:
### CI failure diagnosis + fix

The previous CI fix (`e0b8d0d`) cleared the build error (NETSDK1047 / RID assets) but exposed a latent runtime bug: all 62 UI tests crashed in `UIThreadFixture` with

```
System.Runtime.InteropServices.COMException : Package dependency criteria could not be resolved.
   at MddBootstrapInitialize2_Throw(...)
   at Bootstrap.Initialize(0x00010008, "")
   at OpenClaw.Tray.UITests.UIThreadFixture.UIThreadProc()  // line 138
```

**Root cause.** The fixture calls `Bootstrap.Initialize(1.8, stable)` expecting a system-registered `Microsoft.WindowsAppRuntime.1.8` framework MSIX. The hosted `windows-2025` runner doesn't preinstall it. The misleading inline comment at `.github/workflows/ci.yml:108-110` had been incorrectly assuming otherwise.

**Fix (this commit).** Mirror the main app's deployment shape on the UI test project — `WindowsAppSDKSelfContained=true` + `WindowsPackageType=None`. Runtime DLLs ship next to the test host and `Bootstrap.Initialize` resolves them locally, no system MSIX required. Also suppresses the SDK's auto-init source injection (`WindowsAppSDKBootstrapInitialize` / `UndockedRegFreeWinRTInitialize=false`) because the referenced `OpenClaw.Tray.WinUI` assembly already compiles in the same auto-initializer — re-injection caused CS0436 type conflicts.

**Verified locally:** 62/62 UI tests pass in 1m02s. Net effect for devs cloning the repo: no separate WinAppRuntime install needed.

codemonkeychris:
### CI green ✅ (commit b94e3cd)

Self-contained deployment (the previous attempt at `6c81899`) didn't survive the xunit testhost on CI: `testhost.exe` runs from the .NET SDK directory, not the test bin folder, so the WinAppSDK's P/Invoke-based UndockedRegFreeWinRT auto-initializer can't probe for `Microsoft.WindowsAppRuntime.dll`. Reverted that and went with the simpler runner-side fix instead — added an explicit "Install WindowsAppRuntime 1.8" step in `ci.yml` (version pinned to `1.8.260101001` to match the NuGet reference) and restored the `Bootstrap.Initialize` call in `UIThreadFixture`.

All checks now passing — UI test job runs in ~4 min and reports 62/62 pass.

shanselman:
Thanks Chris — this is a substantial piece of work, and the CI/test coverage here is impressive. I did two review passes (Opus 4.7 + GPT-5.5). Below is the short list of actionable feedback. I’m intentionally not bikeshedding the size/style because I know we want to move fast.

**High-confidence items** — both review passes agree these are worth fixing or explicitly scoping before merge:

1. **Video/Audio media DNS-rebinding claim does not match the implementation.**
   - `MediaResolver` does the strong protected path for images via the safe `HttpClient`/connect callback, but `ResolveMediaUriAsync` returns the original hostname URI for video/audio and `DisplayRenderers` passes that to `MediaSource.CreateFromUri(...)`.
   - That means the Windows media stack performs its own DNS lookup at playback time, outside the resolver, so validation-time DNS and playback-time DNS can differ.
   - Suggested fix: either route video/audio through a safe local/proxy fetch path, disable/limit remote video/audio for now, or adjust the PR description/docs to say DNS-rebinding protection applies to image fetches but not MediaSource-backed playback.

2. **`canvas.navigate` approval prompts can be spammed.**
   - `NodeService.OnCanvasNavigate` starts a fire-and-forget task per request, and each risky URL can create a topmost foreground `MessageBoxW`.
   - A local MCP client/agent with a valid token could loop `canvas.navigate` and stack prompts, creating a UX DoS / habituation problem.
   - Suggested fix: serialize prompts globally, coalesce repeats per `HostKey` while one is pending, and add a short deny cooldown / rate limit so repeated denied hosts auto-deny without more prompts.

3. **The pre-launch DNS re-check is best-effort only, not a real pin.**
   - The code re-resolves immediately before launch, then launches the original hostname URL in the default browser. The browser does its own DNS lookup, so the re-check does not actually constrain the final network target.
   - Suggested fix: either remove that re-check or update comments/docs to describe it as best-effort triage only, with the browser as the trust boundary.

**Medium-confidence / small follow-ups** — one pass found these; they look easy to address but I wouldn’t block the whole PR on them if the high-confidence items are handled:

4. **`McpAuthToken.Reset` can orphan the staged temp token file on failure.**
   - `LoadOrCreate` already has the try/catch cleanup pattern; `Reset` should mirror it around `WriteAllText` / ACL / `File.Move`.

5. **Narrow auth token hot-swap race in `McpHttpServer`.**
   - `HandleAsync` checks `_authToken != null` directly, while `IsAuthorized` uses `Volatile.Read`.
   - Suggested fix: snapshot the auth token once with `Volatile.Read(ref _authToken)` at the top of request handling and use that consistently for the null test and compare.

Overall: the direction looks strong and the security work is thoughtful. I’d be comfortable moving fast once #1 and #2 are fixed, and #3 is either corrected or documented honestly.

codemonkeychris:
### Review pass addressed (commit `108aba9`)

Thanks @shanselman — went through both passes' findings. All five addressed:

**1. Video/Audio DNS-rebinding claim.** Confirmed the gap: `ResolveMediaUriAsync` validated DNS but then passed the hostname URI to `MediaSource.CreateFromUri`, which performs its own resolution at playback — the pin never held for media. Took the docs/scope path: replaced the async method with sync `TryResolveMediaUri` (HTTPS+allowlist + IP-literal-public check, no DNS lookup), and updated `components.md`, `data-and-actions.md`, `grading.md`, and `README.md` to scope the DNS-rebinding guarantee to image fetches only. Image path keeps the `SocketsHttpHandler.ConnectCallback` pin (where it actually holds). A local-proxy approach for media was scoped out — narrow threat (allowlisted host turns hostile via DNS) vs. significant code; can revisit if the threat model shifts.

**2. canvas.navigate prompt spam.** Added in `NodeService`:
- `SemaphoreSlim(1,1)` to serialize prompts globally — no overlapping topmost MessageBoxes across hosts.
- `ConcurrentDictionary<HostKey, Task<Decision>>` to coalesce concurrent requests for the same host onto a shared TCS — second call inherits the first prompt's decision.
- 30-second deny cooldown per HostKey — repeated requests for a denied host auto-deny silently. Re-checked under the gate too, so a queued request can't escape a Deny that landed while it was waiting.

Net effect: a token-holding agent looping `canvas.navigate` can produce at most one prompt every 30s per host, regardless of call rate. Agent still gets immediate "browser" success on every call.

**3. Pre-launch DNS re-check.** Agree — re-resolving here doesn't constrain the browser's lookup. Removed the second `EnrichWithDnsRiskAsync` and its abort branch. Updated the comment to be honest: validation-time DNS is best-effort triage; the OS browser is the actual trust boundary.

**4. `McpAuthToken.Reset` orphan temp file.** Mirrored the `LoadOrCreate` pattern — `WriteAllText`/ACL/`Move` wrapped in try/catch that deletes the temp file on failure and rethrows.

**5. `McpHttpServer` token snapshot race.** Snapshot `_authToken` once via `Volatile.Read` at the top of `HandleAsync`; pass it explicitly to a now-static `IsAuthorized(authToken, header)`. Single read, no possibility of the null-test and the compare seeing different values.

**Tests** (re-run locally, all green):
- `OpenClaw.Shared.Tests` — 1045 passed
- `OpenClaw.Tray.Tests` — 245 passed
- `OpenClaw.Tray.IntegrationTests` (`OPENCLAW_RUN_INTEGRATION=1`) — 18 passed
- `OpenClaw.Tray.UITests` — 62 passed

shanselman:
Added one small test-safety commit on top of Chris's review-pass fixes: `394695c test: suppress browser launch in integration tray`.

Why: the integration suite exercises `canvas.navigate` through a real spawned tray process, so the previous `https://example.com/` test input could open the developer/CI machine's default browser. The test should verify the MCP/tray wiring without creating external desktop side effects.

What changed:
- the integration fixture starts the isolated tray with `OPENCLAW_SUPPRESS_EXTERNAL_BROWSER=1`
- `NodeService` logs and skips the ShellExecute browser launch only when that explicit flag is present
- production `canvas.navigate` behavior is unchanged because the flag is absent by default

Local validation after the commit:
- `./build.ps1` passed
- `OpenClaw.Shared.Tests`: 1045 passed / 20 skipped
- `OpenClaw.Tray.Tests`: 245 passed
- `OpenClaw.Tray.IntegrationTests -r win-arm64`: 18 passed
- `OpenClaw.Tray.UITests -r win-arm64`: 62 passed

shanselman:
Thanks @codemonkeychris — this is a big, valuable step for Windows parity. I verified the final branch locally after the browser-launch test-safety patch:

- `./build.ps1` passed
- `OpenClaw.Shared.Tests`: 1045 passed / 20 skipped
- `OpenClaw.Tray.Tests`: 245 passed
- `OpenClaw.Tray.IntegrationTests -r win-arm64`: 18 passed
- `OpenClaw.Tray.UITests -r win-arm64`: 62 passed
- local smoke demo rendered a native A2UI surface through `canvas.a2ui.push` in a real tray process

CI is green on the final commit. Merging now with appreciation.

PR #244, opened 2026-04-29
feat: add structural input validation for system.run exec approvals

AlexAlves87:
I have the next small, fully tested slice prepared for the exec approvals work. In practice, it adds structural input validation for `system.run`, without introducing executable resolution, policy evaluation, prompting, or any production activation.

The goal here is to keep moving this area forward through small, well-scoped changes that can be reviewed and merged independently while keeping the current legacy path unchanged by default.

## What this PR includes

- a standalone `ExecApprovalV2InputValidator`
- a typed intermediate `ValidatedRunRequest`
- typed `validation-failed` outcomes for malformed or invalid input
- structural validation for:
  - missing or empty command input
  - malformed command/args payloads
  - malformed `cwd`
  - malformed `env`
  - invalid timeout values
- test coverage for valid, invalid, malformed, and edge-case inputs

## What does not change

- `system.run` still uses the current legacy path by default
- this PR does not add executable resolution
- this PR does not add canonical command identity
- this PR does not add policy evaluation or prompting
- this PR does not wire the validator into production behavior yet

## Validation

- `./build.ps1`
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore`
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore`

PR #260, opened 2026-05-02
feat: normalize, resolve, and build canonical command identity

AlexAlves87:
## Summary

Adds the normalization and resolution stage of the exec approval pipeline. Given a validated argv, this stage detects whether the command is a shell wrapper, resolves the executable to its full path, and produces a canonical identity that downstream evaluation consumes as the single source of truth. Nothing from this point forward re-examines the raw request directly.

## Why

The macOS pipeline separates validation from normalization deliberately: validation rejects malformed input early, normalization produces a stable identity for all downstream consumers. Without this stage, each downstream module would need its own resolution logic — inconsistent semantics, duplicated shell-wrapper detection, and no single place to apply fail-closed policy on ambiguous inputs.

`ResolveForAllowlist` is fail-closed on command substitution, `-EncodedCommand` aliases, `env` with modifying flags or assignments before a shell wrapper, unclosed quotes, and paths with non-standard `:` positions.

## What changed

- `ExecShellWrapperNormalizer` — single-level wrapper detection on argv (bash, sh, zsh, cmd, powershell, pwsh); recursively unwraps transparent `env` prefixes up to depth 4
- `ExecEnvInvocationUnwrapper` — strips `env [OPTIONS] [VAR=VAL...] COMMAND` to surface the real executable; fail-closed on unknown flags
- `ExecCommandResolver` — three resolution functions with distinct semantics: singular for the state machine, multi-segment fail-closed for allowlist matching, and pattern suggestions for the prompt UI
- `CanonicalCommandIdentity` — the type that replaces the raw request as input to all downstream modules
- `ExecApprovalV2Normalizer` — orchestrates the three steps and issues a stable `resolution-failed` denial when resolution is impossible

## What's NOT in scope

- Evaluation (state machine, allowlist matching) — next module
- Prompting — depends on evaluation being stable
- Integration with `HandleRunAsync` — deferred until evaluation and prompting are complete

## Test Status

- `OpenClaw.Shared.Tests` — 1,215 passed, 20 skipped, 0 failed
- `OpenClaw.Tray.Tests` — 406 passed, 0 failed

### DIFF
--- a/src/OpenClaw.Shared/ExecApprovals/CanonicalCommandIdentity.cs
+++ b/src/OpenClaw.Shared/ExecApprovals/CanonicalCommandIdentity.cs
@@ -1,0 +1,70 @@
+using System.Collections.Generic;
+
+namespace OpenClaw.Shared.ExecApprovals;
+
+// Architectural barrier produced by PR3.
+// Equivalent to ExecHostValidatedRequest in the macOS reference, extended with resolution outputs.
+// No module from PR4 onward may accept ValidatedRunRequest as direct input (research doc 05 line 439).
+// Rail 15: a single canonical representation reused across evaluation, logging, prompting, execution.
+public sealed class CanonicalCommandIdentity
+{
+    // ── Normalization outputs ─────────────────────────────────────────────────
+
+    // Argv exactly as produced by PR2 (no trimming; coding contract process-argv-semantics).
+    public IReadOnlyList<string> Command { get; }
+
+    // Canonical display form generated from argv. Never rawCommand from the agent.
+    // Used by logging and prompting. Research doc 05 decision 2.
+    public string DisplayCommand { get; }
+
+    // Safe rawCommand for executable resolution. Null in Windows v1 (rawCommand not in
+    // system.run protocol; research doc 05 OQ-V4 / decision 10).
+    public string? EvaluationRawCommand { get; }
+
+    // ── Resolution outputs ────────────────────────────────────────────────────
+
+    // Singular resolution for the state machine (PR5).
+    // Null if the primary executable cannot be determined.
+    public ExecCommandResolution? Resolution { get; }
+
+    // Per-segment resolutions for the allowlist matcher (PR4/PR5).
+    // Empty list means fail-closed — no allowlist satisfaction possible.
+    public IReadOnlyList<ExecCommandResolution> AllowlistResolutions { get; }
+
+    // Suggested allowlist patterns for prompt/UI (PR6). Not a security decision.
+    public IReadOnlyList<string> AllowAlwaysPatterns { get; }
+
+    // ── Request context (carried from ValidatedRunRequest) ────────────────────
+
+    public string? Cwd { get; }
+    public int TimeoutMs { get; }
+    public IReadOnlyDictionary<string, string>? Env { get; }
+    public string? AgentId { get; }
+    public string? SessionKey { get; }
+
+    internal CanonicalCommandIdentity(
+        IReadOnlyList<string> command,
+        string displayCommand,
+        string? evaluationRawCommand,
+        ExecCommandResolution? resolution,
+        IReadOnlyList<ExecCommandResolution> allowlistResolutions,
+        IReadOnlyList<string> allowAlwaysPatterns,
+        string? cwd,
+        int timeoutMs,
+        IReadOnlyDictionary<string, string>? env,
+        string? agentId,
+        string? sessionKey)
+    {
+        Command = command;
+        DisplayCommand = displayCommand;
+        EvaluationRawCommand = evaluationRawCommand;
+        Resolution = resolution;
+        AllowlistResolutions = allowlistResolutions;
+        AllowAlwaysPatterns = allowAlwaysPatterns;
+        Cwd = cwd;
+        TimeoutMs = timeoutMs;
+        Env = env;
+        AgentId = agentId;
+        SessionKey = sessionKey;
+    }
+}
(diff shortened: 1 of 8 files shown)


shanselman:
Hi @AlexAlves87 — Copilot here, commenting on Scott's behalf.

First: this is very much the right direction. The shape of this PR makes sense to me: Windows is catching up to the macOS exec-approval pipeline by separating **validation → normalization/resolution → evaluation/prompting/wiring**, and `CanonicalCommandIdentity` is the right kind of boundary so downstream code does not keep re-parsing raw argv in slightly different ways.

I did a deeper/adversarial pass because this layer is going to become security-policy infrastructure once the later coordinator/evaluator PRs wire it in. I think the architecture is good, and I also think we should tighten a few fail-closed cases before merging so we do not bake in bypass semantics.

The specific issues I found:

1. **PowerShell `-EncodedCommand` detection needs to be stricter.**
   The PR says `ResolveForAllowlist` fails closed on `-EncodedCommand` aliases, which is exactly the right policy. The current scanner looks too literal, though. Please add coverage/fixes for forms PowerShell accepts, including:
   - quoted flags: `powershell "-enc" <payload>`
   - colon/equals forms: `-EncodedCommand:<payload>`, `-enc:<payload>`, `-enc=<payload>`
   - accepted abbreviations / unique prefixes such as `-encod`, `-encode`, etc.
   - direct top-level `powershell` / `pwsh` invocations, not only PowerShell found inside a shell-wrapper segment

   If any of these are present, allowlist resolution should return no satisfiable resolution / fail closed rather than treating `powershell` itself as the allowlisted executable.

2. **`env` modifiers should fail closed unless the effective environment is modeled exactly.**
   Cases like `env PATH=... command`, `env -i ...`, `env -uFOO ...`, `env --chdir=... ...`, or unknown `env` flags can change what actually executes. If normalization resolves using the caller/process environment while the command executes under a modified environment, allowlist matching can approve the wrong executable. I would prefer this layer fail closed for `env` modifiers in allowlist resolution unless it applies the parsed env changes to the resolution environment exactly.

   One edge case to test: unknown `env` flags before a shell wrapper, e.g. `env --bogus bash -c ...`. That should not degrade into “resolve/allow env itself.”

3. **Path/token parsing hardening before this becomes live.**
   These are less urgent than the two above, but still worth fixing while this code is fresh:
   - guard `Path.GetFullPath`, `Path.IsPathFullyQualified`, `Path.Combine`, `Path.GetFileName`, etc. so hostile/odd argv returns a typed resolution failure instead of throwing
   - avoid returning non-executable no-extension PATH shadows before checking `PATHEXT` candidates on Windows
   - improve first-token parsing around partial quoted tokens like `"foo".exe` or `"C:\Program Files\Git\bin\git".exe`
   - tighten colon handling so drive-letter paths require an actual drive letter, and extended-length paths like `\\?\C:\...` are handled deliberately

I am not reading this as a reject. Quite the opposite: this PR is a strong foundation and looks like the right continuation of #244. I just think the normalization layer is the security boundary for the next set of PRs, so the fail-closed claims should be airtight before we take it.

Suggested next step: add regression tests for the `-EncodedCommand` and `env` cases above, make those cases fail closed, and then this should be much closer to mergeable.

AlexAlves87:
Thanks to @shanselman and the Copilot review bot for the detailed security findings on #260 — each addressed as a separate atomic commit below.

Finding #1 — -EncodedCommand bypass via direct invocation (99cc2a9)
`DirectExecUsesEncodedCommand` now catches `["powershell", "-enc", …]` and env-transparent forms. `IsEncodedCommandFlag` recognizes `-ec`, `-enc`, `-encodedcommand`, abbreviations (`-enco…`), and separator forms (`-enc=…`, `-enc:…`). Closes the direct-exec gap the segment scanner didn't cover.

Finding #2 — env with modifiers not fail-closed (c449357)
`ResolveForAllowlist` now checks `HasModifiers` before any unwrap attempt. Catches `env --unknown-flag bash -c …` and `env VAR=val …` before resolution.

Finding #3 — path and token parsing hardening (412eb40)
`ParseFirstToken` is fail-closed on unclosed quotes and preserves suffix after closing quote (`"git".exe → git.exe`). `HasNonStandardColon` handles `\?\` extended-length paths and requires `char.IsAsciiLetter` at the drive position. `FindInPath` probes PATHEXT before the bare name (Windows CreateProcess order).

14 new/updated tests cover the added surface including equals form, env with unknown flags, extended-length paths, and quoted paths with spaces.

shanselman (changes requested):
Thanks for the follow-up work on canonical command identity. I think there is still a security blocker in the encoded-command detection.

Windows PowerShell accepts `powershell.exe -e <base64>` as an alias for `-EncodedCommand` and executes the encoded payload. The new detection appears to cover `-ec`, `-enc`, `-enco`, `-encodedcommand`, and longer prefixes, but not `-e`. That leaves a fail-open gap for allowlist identity: a PowerShell command can still hide the effective command behind encoded content while bypassing the encoded-command denial path.

Please treat `-e` as encoded-command in the direct and shell-wrapper paths and add regression coverage for it.

AlexAlves87:
-`e` wasn't in the list. Fixed in df84ac0: added it to `IsEncodedCommandFlag` alongside `-ec` and `-enc`. Both `DirectExecUsesEncodedCommand` and `SegmentUsesEncodedCommand` call that function, so direct exec and shell-wrapper segments pick it up.

Three regression tests: `["powershell", "-e", "..."]`, `["pwsh", "-e", "..."]`, `["bash", "-c", "powershell -e ..."]`.

shanselman (approved):
Validated locally and remote CI is green after the -en EncodedCommand abbreviation fix.

PR #298, opened 2026-05-08
Add MXC Sandbox example to Windows Node

MythiliMur:
Add MXC sandbox example debug action and readable output

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,53 @@
+{
+  "name": "openclaw-windows-node",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "dependencies": {
+        "@microsoft/mxc-sdk": "^0.1.8"
+      }
+    },
+    "node_modules/@microsoft/mxc-sdk": {
+      "version": "0.1.8",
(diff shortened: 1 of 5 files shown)


shanselman:
Thanks for putting this MXC example together. Since this PR was opened, #319 has landed a production MXC/AppContainer implementation for `system.run`, including the package setup, bridge packaging, Sandbox settings page, availability checks, policy construction, and tests.

I don't think this PR should be merged as-is now because:

- It conflicts with current `master` after #319, especially the root `package.json` / `package-lock.json` and `tools\mxc` work.
- The debug-page button duplicates/sidesteps the production sandbox path instead of exercising the shared `MxcCommandRunner` / Sandbox settings flow that now exists.
- The script lookup walks parent directories from the current directory, which is exactly the kind of runtime-probing pattern we removed in #319 for safety.
- It adds an interactive/debug demo without test coverage or a clear production UX path.

Closing as superseded by #319. If we still want a sample, the better follow-up would be a small docs/dev-only example that calls the production `system.run` sandbox path rather than adding a separate Debug page runner.

shanselman:
Closing as superseded by #319's production MXC sandbox implementation.

PR #299, opened 2026-05-08
feat: gateway registry (multi-gateway support)

ranjeshj:
## Gateway Registry — Multi-Gateway Support

Single source of truth for gateway URL, tokens, bootstrap state, connection status, and active client reference.

### What it does
- **Gateway Registry** (\gateways.json\): stores per-gateway credentials so switching gateways doesn't require re-pairing
- **Recent Gateways UI**: Connection page card with one-click switching
- **Chat auth fix**: \SharedGatewayToken\ stored separately so web dashboard works after pairing
- **WSL onboarding fix**: encoding mismatch causing \wsl.exe --list\ hang
- **Radio button defaults**: auto-select first option in wizard

### Security
- GUID-randomized temp file, ACL re-applied after File.Move
- Tokens removed from URLs
- 30 unit tests for registry operations

### DIFF
--- a/src/OpenClaw.Shared/ConnectionDiagnostics.cs
+++ b/src/OpenClaw.Shared/ConnectionDiagnostics.cs
@@ -1,0 +1,120 @@
+using System;
+using System.Collections.Generic;
+
+namespace OpenClaw.Shared;
+
+/// <summary>
+/// A single diagnostic event from the connection state machine.
+/// </summary>
+public sealed record ConnectionEvent(
+    DateTime Timestamp,
+    string Direction,     // "→ GW", "← GW", "→ WSL", "INTERNAL"
+    string EventType,     // "CONNECT", "CHALLENGE", "AUTH", "HELLO_OK", "ERROR",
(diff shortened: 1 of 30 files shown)


shanselman:
Blocking: applying a setup code does not clear the stored node device token.

`SetupCodeApplicator.Apply()` treats setup codes as potentially targeting a different gateway and calls `DeviceIdentity.TryClearStoredDeviceToken(dataPath)`, but that helper only clears `DeviceToken` and `OperatorDeviceToken`. It leaves `NodeDeviceToken` / `NodeDeviceTokenScopes` in `device-key-ed25519.json`.

That is a problem for this multi-gateway flow because `WindowsNodeClient` prefers the stored node token before falling back to the new bootstrap token. After switching gateways with a setup code, the node can therefore try to authenticate with the previous gateway’s stale node token instead of pairing with the fresh bootstrap token.

Suggested fix: update `TryClearStoredDeviceToken()` to clear all role-specific device token fields, for example:

```csharp
data.DeviceToken = null;
data.DeviceTokenScopes = null;
data.OperatorDeviceToken = null;
data.NodeDeviceToken = null;
data.NodeDeviceTokenScopes = null;

PR #315, opened 2026-05-12
Native WinUI chat experience

RBrid:
## Summary
 
 This PR adds the native WinUI chat experience for the OpenClaw Windows node and makes it selectable/debuggable from
 the tray app.
 
### Native chat experience
 
 - Adds the native WinUI chat surface as an alternative to the standard Gateway Chat WebView.
 - Renders user/assistant turns natively, including message grouping, assistant reasoning, tool calls/tool results, 
permission prompts, status rows, and per-message actions.
 - Adds native chat timeline controls such as copy, read aloud, delete affordances, compact tool chips, expanded 
tool output, and hover-revealed bubble actions.
 - Aligns the native chat visuals with the current OpenClaw/Fluent styling, including acrylic/chat shell polish, 
theme-aware colors, avatars, and footer actions.
 
### Settings and debug options
 
 - Adds the Settings option to use the standard Gateway Chat interface instead of the native WinUI chat surface.
 - Adds per-surface chat UI override/debug support so native vs standard chat can be toggled for validation and 
fallback scenarios.
 - Improves tray/chat launch behavior when chat credentials or chat URLs are unavailable, showing an error state 
instead of navigating to an invalid chat URL.
 
### Gateway and chat behavior
 
 - Wires native chat to gateway chat history and live chat events.
 - Handles cumulative assistant stream/final messages, reasoning text, lifecycle/status events, permission events, 
tool calls, and tool results.
 - Preserves history replay ordering while keeping live assistant final-message reconciliation from duplicating 
responses.
 - Routes read-aloud through the configured TTS path and guards fallback TTS cleanup.
 
### Stability and review fixes
 
 - Fixes stuck native-chat spinner state after lifecycle/provider errors.
 - Fixes duplicated assistant responses when final assistant messages arrive after lifecycle end.
 - Improves input responsiveness by avoiding timeline re-renders on every composer keystroke.
 - Hardens markdown sanitization around malformed links/images, reference definitions, and bounded scan behavior.
 - Adds regression coverage for reducer turn lifecycle, duplicate assistant reconciliation, markdown sanitizer 
behavior, provider event handling, and chat data edge cases.
 
## Validation
 
 - `.\build.ps1`

## Coming investigation

Alternative approaches are being investigated: using WinUI 3's ListView with KeepLastItemInView, or ScrollView + ItemsRepeater  without the use of reactor.


shanselman:
Thanks for pushing this forward. The native WinUI chat direction looks promising, and I appreciate the amount of work here. The main thing blocking merge for me right now is not the chat UX itself, but the size and shape of the vendored Reactor dependency.

This PR currently brings in a very large `external/reactor` tree as a project reference from `OpenClaw.Tray.WinUI`. From the PR file list, this is roughly 291 Reactor source files inside a 348-file PR, and it adds broad subsystems that the OpenClaw chat surface does not appear to use: charting/D3, data grid/property grid/validation controls, devtools/MCP/debug hosting, ETW/layout-cost tooling, data providers, etc. That is a lot of long-term code ownership and dependency surface for a single chat feature.

I do not think we should merge the full Reactor snapshot as-is. I would like this PR to either split the vendored runtime from the chat UI, or trim Reactor down to the minimum runtime needed by the OpenClaw chat surface before merge.

A practical trim path would be:

1. Start with a compile-trim rather than guessing file-by-file. In `external/reactor/src/Reactor/Reactor.csproj`, temporarily exclude unused feature folders with `Compile Remove`, then build iteratively.

   ```xml
   <ItemGroup>
     <Compile Remove="Charting\**\*.cs" />
     <Compile Remove="Controls\**\*.cs" />
     <Compile Remove="Data\**\*.cs" />
     <Compile Remove="Hosting\Devtools\**\*.cs" />
     <Compile Remove="Hosting\Etw\**\*.cs" />
     <Compile Remove="Hosting\LayoutCost\**\*.cs" />
   </ItemGroup>
   ```

2. Keep the pieces that the current chat implementation actually depends on:
   - `Core`
   - `Elements`
   - `Hooks`
   - the minimal `Hosting` pieces needed for `ReactorHost`
   - `Markdown`
   - `Yoga` / `Microsoft.UI.Reactor.Layout`, since `FlexRow`, `FlexColumn`, `.Flex(...)`, and `FlexPanel` depend on it

3. Remove package references that only support excluded features. For example, `Microsoft.Diagnostics.Tracing.TraceEvent` looks tied to ETW/layout-cost support, and `System.Drawing.Common` looks tied to preview/devtools screenshot capture. If those folders are removed, those package refs should go too.

4. After the compile trim passes, physically delete the excluded folders rather than leaving dead vendored code in the repo.

5. Update `external/reactor/README.md` and `NOTICE.md` to describe this as a trimmed Reactor runtime, not a full upstream snapshot. Please include:
   - which folders were retained
   - which folders were intentionally removed
   - why the retained folders are needed by OpenClaw chat
   - how to refresh from upstream and reapply the local edits

6. Validate the trimmed result with at least:

   ```powershell
   dotnet build external\reactor\src\Reactor\Reactor.csproj
   dotnet build src\OpenClaw.Tray.WinUI\OpenClaw.Tray.WinUI.csproj -r win-x64
   .\build.ps1
   ```

The goal is not to block native chat; it is to avoid taking ownership of a full UI framework plus unused charting/devtools/data-grid systems when this feature appears to need only the core declarative runtime, markdown rendering, flex layout, hooks, and host bridge.

Once the Reactor tree is trimmed or split into a focused prerequisite PR, this will be much easier to review and much safer to merge.

shanselman:
One additional point on the Reactor dependency: we already have an internal declarative WinUI layer in this repo at `src/OpenClawTray.FunctionalUI/FunctionalUI.cs`, and it is actively used by the onboarding flow.

That existing layer already provides several of the same concepts this PR is bringing in through Reactor:

- `Component` / `Component<TProps>`
- `UseState`
- `UseEffect`
- props-driven rendering
- `FunctionalHostControl`
- factories for common WinUI controls
- `VStack`, `HStack`, `Grid`, `ScrollView`
- navigation primitives
- tests under `tests/OpenClawTray.FunctionalUI.Tests`

So before we take a second declarative UI framework into the repo, I think this PR needs an explicit decision point:

1. Preferably, build the native chat surface on top of `OpenClawTray.FunctionalUI` and extend that internal layer with only the primitives chat needs.
2. If Reactor is still required, please document exactly what FunctionalUI cannot currently do and why those gaps are better solved by vendoring Reactor instead of extending our in-repo layer.

The likely FunctionalUI gaps for this chat surface seem concrete and bounded: markdown/rich text rendering, combo boxes/flyouts, better keyed list reconciliation, and streaming timeline performance. Those may be solvable as targeted additions to our own FunctionalUI project without taking ownership of a full external UI framework.

If the final decision is still to use Reactor, then I think the earlier trim request still stands: bring only the minimal Reactor runtime needed for those missing capabilities, not the full upstream snapshot.

PR #323, opened 2026-05-12
fix(onboarding): setup code bootstrap flow, connection error handling, and token priority

reisball:
The onboarding wizard fails to complete when using a setup code from the
gateway. This PR fixes the bootstrap token flow end-to-end.

Changes:
- Remove Origin header from native WebSocket client (gateway misclassifies
  it as Control-UI)
- Handle PAIRING_REQUIRED error code alongside NOT_PAIRED
- Prioritize BootstrapToken over SharedGatewayToken in credential resolution
- Store setup code tokens as BootstrapToken, debounce transient Error
  state in polling loop (v3→v2 signature fallback briefly sets Error)
- Check operator AND node token in RequiresSetup (previous logic skipped
  the wizard for fresh installs when EnableMcpServer was enabled)

Generated with OpenClaw (model: DeepSeek v4 Flash).

### DIFF
--- a/src/OpenClaw.Shared/WebSocketClientBase.cs
+++ b/src/OpenClaw.Shared/WebSocketClientBase.cs
@@ -118,11 +118,7 @@ public abstract class WebSocketClientBase : IDisposable
             _webSocket = new ClientWebSocket();
             _webSocket.Options.KeepAliveInterval = TimeSpan.FromSeconds(30);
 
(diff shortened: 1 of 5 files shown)


reisball:
Closing this for now — after deeper review, the proposed changes don’t match the intended onboarding flow. I’ll rework from the documented behavior/tests first.

PR #324, opened 2026-05-12
feat(nodes): rich node cards with rename, forget, and tray cleanup

bkudiess:
## Summary

Tray-app changes covering four areas of how the operator manages paired nodes:

1. **Tray flyout** — the **Devices** section now shows only currently-connected nodes. Stale paired-but-offline entries no longer accumulate in the right-click menu; they remain accessible on the full Nodes page.
2. **Gateway model + parser** — `GatewayNodeInfo` now carries the rest of the `NodeListNode` schema the gateway already sends (`Version`, `CoreVersion`, `UiVersion`, `ClientId`, `ClientMode`, `DeviceFamily`, `ModelIdentifier`, `RemoteIp`, `PathEnv`, `ConnectedAt`, `ApprovedAt`, `LastSeenReason`, `IsPaired`, `HasExplicitDisplayName`, `DisabledCommands`). `ParseNodeList` now reads the production `*Ms` timestamp wire names; legacy non-Ms fallbacks remain for mocks/tests. `LastSeen` no longer falls back to `connectedAtMs` so the same value never appears twice in the UI.
3. **New gateway client methods**:
   - `NodeRenameAsync` → returns `NodeRenameResult` (success/error message)
   - `NodePairRemoveAsync` → returns `NodeForgetResult` (success/error message)

   Both use `SendWizardRequestAsync` rather than the fire-and-forget `TrySendTrackedRequestAsync`, so scope rejections, missing nodeIds, and other application-level failures are reported back to the caller instead of being silently swallowed. The `node.pair.resolved` broadcast handler now refreshes both the pair list and the node list.
4. **NodesPage** — full rewrite of the per-node card. Each node renders as an `Expander` (online auto-expanded, offline collapsed) with a body that shows the identity row, version line, hardware, network, timestamps, capability tags (now actually populated), commands list with disabled annotations, permissions grid, and an optional collapsed PATH dump. Action footer at the bottom — separator + a right-aligned **Rename** / **Forget** pair — follows the Win11 Settings pattern (Manage / Remove on Email account cards). Both actions open `ContentDialog`s using the deferral pattern so failures stay inline instead of silently closing.

## Implementation notes

- Click lambdas wrap their async work in `try/catch` so an unhandled exception in a dialog flow can never become an `async void` crash.
- Rename `TextBox` pre-fills with the explicit display name only — never with the parser's fallback id — so pressing Enter on an unnamed node doesn't persist the id as the new display name.
- Layout uses `Grid` with star/auto columns where `TextWrapping` and `TextTrimming` actually matter (header, identity row, label rows); long ids/names ellipsize with a tooltip, long values wrap.
- `ModelFormatting.FormatAge` was made `public` and grew a clock-skew guard plus a **>30d → absolute date** branch; both the header `DetailText` and the body timestamps now share it, so a single timestamp can't show as "1d ago" in one place and "36h ago" in another.
- Forget dialog defaults to **Cancel** so Enter doesn't accidentally destroy a pairing; Rename defaults to **Primary** so Enter confirms.
- All UI updates flow from gateway responses — no optimistic local mutations on rename or forget.
- 25 new resource keys per locale (en-us, fr-fr, nl-nl, zh-cn, zh-tw); `Version`, `Hardware`, and `PATH` are registered as invariant.

## Tests

- `OpenClaw.Shared.Tests` — **+11 tests** covering the full `NodeListNode` schema, legacy wire-name compatibility, minimal-payload defaults, and rename/forget input validation paths. Total: **1547 passed**, 28 skipped, 0 failed.
- `OpenClaw.Tray.Tests` — mock implementation of `IOperatorGatewayClient` updated for the new methods. Localization validation tests pass with the three invariant exceptions registered. Total: **965 passed**, 0 failed.
- `dotnet build` clean.

## Screenshots

The Devices section in the tray flyout no longer lists offline duplicates. Each Nodes page card now shows a full Expander with rename/forget at the bottom. (Manual smoke tested against a live gateway.)

## Known follow-ups (out of scope for this PR)

- **Scope-aware action gating** — Rename and Forget buttons are not yet disabled when the session lacks `operator.pairing`. Bootstrap-token sessions will see enabled buttons that fail with the gateway's "missing scope" message inline — same as the existing pair approve/reject buttons. A wider scope-aware UI pass should expose granted scopes on `IOperatorGatewayClient` and gate everywhere consistently.
- **Capability badge wrap** — capability pills currently use a horizontal `StackPanel` (no `WrapPanel` in WinUI 3 base). Real but low-impact; defer until a paired node with many caps surfaces overflow.
- **Tray "+N more" overflow** — quality-of-life link when more than 5 nodes are online.

## Validation checklist

- [x] `./build.ps1` clean
- [x] `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj`
- [x] `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj`
- [x] Manual smoke against a live gateway

### DIFF
--- a/src/OpenClaw.Shared/IOperatorGatewayClient.cs
+++ b/src/OpenClaw.Shared/IOperatorGatewayClient.cs
@@ -87,6 +87,8 @@ public interface IOperatorGatewayClient
     Task RequestNodePairListAsync();
     Task<bool> NodePairApproveAsync(string requestId);
     Task<bool> NodePairRejectAsync(string requestId);
+    Task<NodeForgetResult> NodePairRemoveAsync(string nodeId);
+    Task<NodeRenameResult> NodeRenameAsync(string nodeId, string displayName);
     Task RequestDevicePairListAsync();
     Task<bool> DevicePairApproveAsync(string requestId);
     Task<bool> DevicePairRejectAsync(string requestId);
--- a/src/OpenClaw.Shared/Models.cs
+++ b/src/OpenClaw.Shared/Models.cs
@@ -491,6 +491,32 @@ public class GatewayNodeInfo
     public List<string> DisabledCommands { get; set; } = new();
     public Dictionary<string, bool> Permissions { get; set; } = new(StringComparer.OrdinalIgnoreCase);
 
+    // Identity / hardware (from gateway NodeListNode schema)
+    public string? Version { get; set; }
+    public string? CoreVersion { get; set; }
+    public string? UiVersion { get; set; }
+    public string? ClientId { get; set; }
+    public string? ClientMode { get; set; }
+    public string? DeviceFamily { get; set; }
+    public string? ModelIdentifier { get; set; }
+    public string? RemoteIp { get; set; }
+    public string? PathEnv { get; set; }
+
+    // Timestamps and state
+    public DateTime? ConnectedAt { get; set; }
+    public DateTime? ApprovedAt { get; set; }
+    public string? LastSeenReason { get; set; }
+
+    // True when the node is in the gateway's paired set (regardless of current
+    // connection state). Distinct from IsOnline — a paired node can be offline.
+    public bool IsPaired { get; set; }
+
+    // True when the gateway provided an explicit displayName/name/label.
+    // False when the parser had to fall back to shortId or nodeId. UI surfaces
+    // (e.g. the rename dialog) use this to distinguish "the user gave this
+    // node a name" from "we showed the id because there was nothing better".
+    public bool HasExplicitDisplayName { get; set; }
+
     public string ShortId => NodeId.Length <= 12 ? NodeId : NodeId[..12] + "…";
 
     public string DisplayText
@@ -521,6 +547,26 @@ public class GatewayNodeInfo
 
     private static string FormatAge(DateTime timestampUtc) => ModelFormatting.FormatAge(timestampUtc);
 }
+
+/// <summary>
+/// Result of a <c>node.rename</c> request to the gateway.
+/// </summary>
+/// <param name="Success">True when the gateway accepted the rename and persisted it.</param>
+/// <param name="NodeId">Node id the gateway returned; same as the requested id on success.</param>
+/// <param name="DisplayName">Updated display name as persisted by the gateway.</param>
+/// <param name="ErrorMessage">Gateway-supplied or transport-derived error description; null on success.</param>
+public sealed record NodeRenameResult(
+    bool Success,
+    string? NodeId = null,
+    string? DisplayName = null,
+    string? ErrorMessage = null);
+
+/// <summary>
+/// Result of a <c>node.pair.remove</c> request to the gateway.
+/// </summary>
+/// <param name="Success">True when the gateway accepted the removal.</param>
+/// <param name="ErrorMessage">Gateway-supplied or transport-derived error description; null on success.</param>
+public sealed record NodeForgetResult(bool Success, string? ErrorMessage = null);
 
 public enum GatewayDiagnosticSeverity
 {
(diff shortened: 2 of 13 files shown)


shanselman:
## Review feedback from adversarial PR #324 pass

I reviewed this with a critical eye against the tray/gateway interaction paths, and also cross-checked it with a two-model review. Overall this looks solid: the PR is clean/mergeable, CI is green, the gateway parser changes line up with the richer `NodeListNode` shape, and the rename/forget RPC paths correctly wait for gateway ack/error responses instead of treating WS send as success.

I do recommend addressing one small WinUI dialog reentrancy issue before merge. The remaining items are follow-up/polish level.

### Should fix before merge: page-wide ContentDialog guard

`NodesPage` currently tracks open dialogs by node id:

- `_nodesWithDialogOpen` is a `HashSet<string>` keyed by `node.NodeId`.
- `OnRenameClickedAsync` and `OnForgetClickedAsync` only block another dialog for the same node.
- Both methods then call `await dialog.ShowAsync()`.

That prevents double-clicking Rename/Forget for one node, but it does not prevent a second dialog for a different node from being shown while the first dialog is still opening/open. WinUI only allows one `ContentDialog` per `XamlRoot`; this repo already documents that exact failure mode in `SandboxPage.xaml.cs` around the sandbox confirmation guard (`Cannot show another dialog until the previous one is dismissed`).

Impact: a fast user path like Rename on node A, then Forget on node B can make the second `ShowAsync()` throw/fail. The outer click-handler catch appears to keep this from crashing the app, but the action silently fails with no user feedback.

Suggested fix: make the guard page-wide instead of per-node, or at least check the set count before adding the current node:

```csharp
if (_nodesWithDialogOpen.Count > 0) return;
if (!_nodesWithDialogOpen.Add(node.NodeId)) return;
```

A clearer alternative would be a dedicated `_dialogOpen` boolean similar to `SandboxPage`'s `_confirmDialogOpen`. This is a small, high-confidence fix and should get us back to merge confidence.

### Follow-up / polish items

1. Rename TextBox selection likely does not take effect. `input.SelectAll()` is called before the TextBox is loaded into the dialog visual tree. If the intended UX is "existing name selected when the dialog opens", move it to `Loaded` and optionally focus the control there:

```csharp
input.Loaded += (_, _) =>
{
    input.Focus(FocusState.Programmatic);
    input.SelectAll();
};
```

2. Forget confirmation uses accent styling for a destructive action. The card-level Forget affordance signals destructive intent, and the dialog correctly defaults to Cancel, but the primary button is styled with `AccentButtonStyle`. Consider a critical/destructive style for the confirmation primary button, or remove the comment that implies destructive styling is reserved for the confirmation dialog.

3. Capability badges can overflow horizontally. `BuildCapabilitiesSection` uses a horizontal `StackPanel` for badges. With the richer gateway capability data, long capability lists or long capability names can push content off-card. A wrapping layout would be more robust if this starts showing up in real node data.

4. Error localization is mostly bypassed by raw gateway messages. The UI comments say generic localized fallback will be used for unrecognized failures, but `NodeRenameAsync` / `NodePairRemoveAsync` generally return `ex.Message`, so the fallback is rarely reached. This is acceptable for actionable errors like `missing scope: ...`, but if we care about non-English locales here, consider only surfacing known/actionable gateway messages and using the localized fallback otherwise.

5. Clearing a display name is not supported. Both the client and gateway reject empty `displayName`, so users cannot revert a renamed node back to an unnamed/short-id fallback. I would not block this PR on it because it appears to be a protocol/product decision, not just a Windows UI issue.

bkudiess:
Thanks for the careful read on both passes. Pushed [`562cf7b`](https://github.com/openclaw/openclaw-windows-node/pull/324/commits/562cf7b) addressing the three concrete items:

### Should-fix
1. **Page-wide ContentDialog guard** — replaced the per-node `_nodesWithDialogOpen` HashSet with a page-wide `_dialogOpen` bool, matching the `SandboxPage._confirmDialogOpen` convention. Rename-on-A then Forget-on-B no longer races into a swallowed `ShowAsync` throw.

### Polish
2. **Rename `SelectAll` timing** — moved `input.Focus(...)` + `input.SelectAll()` into the TextBox's `Loaded` event so the selection actually happens after the dialog inserts the control.
3. **`ex.Message` leaking into UI** — `NodeRenameAsync`/`NodePairRemoveAsync` now split the catch by type: `InvalidOperationException` (the gateway's `ok=false` ack) is still surfaced verbatim because those messages are actionable (`missing scope: operator.pairing`, `unknown nodeId`); any other exception returns `null` `ErrorMessage` so the dialog falls back to its localized generic string. Internal exception text no longer reaches the UI.

Also dropped the `AccentButtonStyle` on the Forget confirmation primary button — destructive primaries shouldn't be styled as call-to-action; Cancel remains the default focus.

### Tracked for follow-up (not in this commit)
- **Capability badge wrap** — the horizontal `StackPanel` overflow case will get fixed alongside a wider "scope-aware action gating" pass; both touch the same neighborhood of NodesPage and felt like one coherent PR.
- **Clearing a display name** — agreed, gateway-side protocol decision rather than a client-side bug.

`./build.ps1` clean; `OpenClaw.Shared.Tests` 1547 passed; `OpenClaw.Tray.Tests` 965 passed.

PR #326, opened 2026-05-13
fix: native chat UX improvements — focus, flickering, bubbles, notifications

ranjeshj:
## Summary

Fixes several UX issues in the native FunctionalUI chat surface:

### Composer focus loss
- Use \UseRef\ for input text instead of \UseState\ to prevent full-tree re-render on every keypress
- \UseState<bool>\ tracks empty↔non-empty for send button accent styling (2 re-renders per compose cycle, not per-keypress)

### Flickering during streaming
- \ConfigureBorder\: skip child detach/reattach when child reference is unchanged
- \ConfigureScrollView\/\ConfigureExpander\: same \ReferenceEquals\ guard
- \ConfigureGrid\: skip column/row definition rebuild when definitions match

### Bubble clipping/overflow
- Switch bubble rows from \FlexRow\ (StackPanel) to \Grid\ with Auto+Star columns so bubbles constrain to available window width
- Remove fixed 560px \MaxWidth\ cap — bubbles now fill available space responsively

### Notifications
- Suppress chat toast notifications when tray \ChatWindow\ is already visible

### Chat window chrome
- Suppress orphaned Esc keyboard accelerator tooltip on scroll
- Use atomic \SetWindowPos\ for flicker-free show/hide positioning

### DIFF
--- a/src/OpenClaw.Tray.WinUI/Windows/ChatWindow.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/Windows/ChatWindow.xaml.cs
@@ -34,7 +34,9 @@ public sealed partial class ChatWindow : WindowEx
     [DllImport("user32.dll")] private static extern bool GetMonitorInfo(IntPtr hMonitor, ref MONITORINFO mi);
     [DllImport("user32.dll")] private static extern bool SetForegroundWindow(IntPtr hWnd);
     [DllImport("user32.dll")] private static extern uint GetDpiForWindow(IntPtr hWnd);
+    [DllImport("user32.dll")] private static extern bool SetWindowPos(IntPtr hWnd, IntPtr hWndInsertAfter, int X, int Y, int cx, int cy, uint uFlags);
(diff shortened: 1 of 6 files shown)


src/OpenClaw.Tray.WinUI/Windows/ChatWindow.xaml.cs
```diff
@@ -34,7 +34,9 @@ public sealed partial class ChatWindow : WindowEx
     [DllImport("user32.dll")] private static extern bool GetMonitorInfo(IntPtr hMonitor, ref MONITORINFO mi);
     [DllImport("user32.dll")] private static extern bool SetForegroundWindow(IntPtr hWnd);
     [DllImport("user32.dll")] private static extern uint GetDpiForWindow(IntPtr hWnd);
+    [DllImport("user32.dll")] private static extern bool SetWindowPos(IntPtr hWnd, IntPtr hWndInsertAfter, int X, int Y, int cx, int cy, uint uFlags);
```
ranjeshj, line 37: These changes need to be removed.

PR #327, opened 2026-05-13
Add exec approval evaluator, typed outcomes, and allowlist matcher

AlexAlves87:
Implements the stateless evaluation layer for the new exec approvals path (PR5 of the slice sequence). No UI, no side effects, no coordinator wiring — this component is ready to be consumed by the coordinator in PR7.

## What this adds

**`ExecApprovalEvaluator`** — pure decision function. Same inputs always produce the same output. Takes an aggregated context and an optional user decision; returns one of three typed outcomes. No I/O, no store access, no matching.

Five-step precedence (order is fixed, mirrors macOS `ExecHostRequestEvaluator`):
1. `security=deny` — absolute override, no user approval can bypass it
2. `approvalDecision=deny` — explicit user refusal
3. `requiresAsk() && no decision` — prompt required
4. `security=allowlist && no match && no user approval` — silent deny
5. default — allow, with `ApprovedByAsk` flag for downstream side-effect signaling

**`ExecHostPolicyDecision`** — sealed class hierarchy representing the three outcomes: `DenyOutcome`, `RequiresPromptOutcome`, `AllowOutcome`. `AllowOutcome` carries `ApprovedByAsk: bool` so the coordinator can distinguish user-approved allows from allowlist-satisfied allows when triggering `addAllowlistEntry` vs `recordAllowlistUse`.

**`ExecApprovalEvaluation`** — aggregated context built by the coordinator (PR7), not by the evaluator. Derived fields (`AllowlistSatisfied`, `AllowlistMatch`, `Resolution`) are computed once in the constructor to prevent re-derivation inconsistencies across callers.

**`ExecAllowlistMatcher`** — path-based matcher with glob support. Target is always `resolvedPath ?? rawExecutable`. Glob semantics: `*` = single segment `[^/]*`, `**` = any depth `.*`, `?` = single char without crossing separators `[^/]` (security decision: no separator crossing on Windows). Uses `RegexOptions.NonBacktracking` to prevent DoS on adversarial patterns. Basename-only patterns are invalid and fail-closed.

## What this does not do

- No store reads or writes
- No UI, no WinUI types
- No coordinator logic — `ExecApprovalEvaluation` construction and two-pass orchestration belong to PR7
- No changes to `HandleRunAsync`, `IExecApprovalV2Handler`, or the routing seam from PR1
- `ExecApprovalPolicy` (legacy path) is untouched

## Tests

80 new tests covering: five-step precedence pairs, two-pass mechanics, `RequiresAsk` full enum coverage, `ExecApprovalEvaluation` derived-field correctness, matcher glob semantics, separator normalization, case insensitivity, UNC paths, adversarial patterns, and fail-closed edge cases.


AlexAlves87:
Addressed the 🔴: `**/rest` now emits `(.*\/)?` instead of `.*`, anchoring the following segment at a path boundary. Two regression tests added (`DoesNotMatchPartialBasename`, `MatchesRootLevel`).

The 🟡 on the 10-arg constructor is noted, will revisit for PR7 when building the coordinator.

shanselman (changes requested):
Thanks for this slice. I reviewed it on top of current `master` after today’s merges. It clean-merges and CI is green, but I think this needs one security fix before merge.

Blocking issue: `ExecAsk.Deny` currently fails open in the evaluator.

Evidence:
- `ExecApprovalsContracts.cs` defines `ExecAsk.Deny`.
- `ExecApprovalsStore` defaults `AskFallback` to `ExecAsk.Deny`.
- `ExecApprovalEvaluator.RequiresAsk(...)` only treats `Always` and `OnMiss` as prompting modes.
- `Evaluate(...)` has no branch that maps `context.Ask == ExecAsk.Deny` to a `DenyOutcome`.

That means a context with `security=Full` and `ask=Deny` falls through to step 5 and returns `Allow`. If `AskFallback=Deny` is later passed into this evaluator for a non-interactive/fallback path, that becomes a policy bypass.

Suggested fix: give `ExecAsk.Deny` explicit deny semantics in `Evaluate(...)` before the prompt/miss logic, and add tests for at least:

- `security=Full, ask=Deny, approvalDecision=null` => deny
- `security=Allowlist, ask=Deny, allowlist satisfied` => deny, if deny is intended as an absolute ask/fallback mode
- `security=Deny` still wins with the existing `SecurityDeny` code

If the intended semantic is actually “never ask but still allow when security is full,” then I think we should not call that enum value `Deny`, and we need to revisit the `AskFallback = ExecAsk.Deny` default before wiring this into production.

Non-blocking hardening: I’d also consider rejecting malformed `**` patterns that are not segment-boundary globs, e.g. `C:\tools**`, because they compile to a broader prefix match than an admin is likely to expect (`C:/tools.*`). That can be follow-up after the `ExecAsk.Deny` behavior is fixed.

AlexAlves87:
Thanks for the detailed review. Both issues addressed in 488e2e9.

**ExecAsk.Deny (blocking)**

Added step 4.5 in `Evaluate(...)`, after the allowlist-miss check:

```
ask=Deny && approvalDecision is null && !AllowlistSatisfied → AskDeny
```

The `!AllowlistSatisfied` guard is intentional: when the allowlist pre-authorized a tool, no prompt would ever have been issued — `AskFallback` has nothing to do there. The bypass you identified (`security=Full, ask=Deny` falling through to Allow) is fixed.

Of your three suggested tests, the first and third are covered exactly as written. For the second (`security=Allowlist, ask=Deny, allowlist satisfied`), we went with Allow on the reasoning above. Happy to discuss if you see it differently.

New `AskDeny` code and factory in `ExecApprovalV2Result` keep the denial reasons semantically distinct from `SecurityDeny`, `UserDenied`, and `AllowlistMiss`.

**Malformed `**` (non-blocking)**

`HasMalformedDoubleStars` validates that `**` is preceded by start-of-string or `/`, and followed by `/` or end. Patterns like `C:\tools**` now fail-closed with no match. Well-formed patterns (`**/foo`, `C:\**\foo`, `C:\tools\**`) are unaffected.

6 new tests for `ask=Deny` semantics, 7 for malformed `**`.

shanselman (approved):
The updated branch addresses the prior fail-open concern: ExecAsk.Deny now fails closed when there is no explicit approval, allowlist/security precedence is covered, and malformed ** allowlist patterns fail closed. Local validation passed with build, Shared tests, and Tray tests.

PR #336, opened 2026-05-13
refactor: extract CopyDiagnostic helper for diagnostic copy methods

AlexAlves87:
## Summary

`App.xaml.cs` had nine diagnostic copy methods that each repeated the
same 13-line body: build a `DataPackage`, call the appropriate
`CommandCenterTextHelper` formatter, write to clipboard, and log success
or failure. This extracts that repeated structure into a single private
`CopyDiagnostic` helper and reduces each method to a one-liner.

## What changed

- Added `CopyDiagnostic(string label, Func<GatewayCommandCenterState, string> format)` as a private instance method in `App.xaml.cs`
- The helper reuses the existing `CopyTextToClipboard` static already present in the file
- All nine methods are now expression-bodied one-liners; signatures are unchanged (they are exposed as `Action` delegates in `DeepLinkActions`)
- Log messages preserved exactly

## Testing

- `dotnet build src/OpenClaw.Tray.WinUI/OpenClaw.Tray.WinUI.csproj -r win-x64 --nologo`
- Manual: tray menu → Support & Debug → Copy Support Context, Copy Port Diagnostics, Copy Node Inventory — pasted output matches previous

## Notes

No behaviour change. The one subtle difference worth noting: exceptions thrown inside `BuildCommandCenterState()` are now caught by the helper's try/catch, which was already the case in each of the original per-method try/catch blocks.

### DIFF
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -4031,140 +4031,45 @@ public partial class App : Application
         }
     }
 
-    private void CopySupportContext()
+    private void CopyDiagnostic(string label, Func<GatewayCommandCenterState, string> format)
     {
         try
         {
-            var package = new DataPackage();
-            package.SetText(CommandCenterTextHelper.BuildSupportContext(BuildCommandCenterState()));
-            Clipboard.SetContent(package);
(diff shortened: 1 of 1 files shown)


shanselman:
Fantastic! Thanks for preventing slop @AlexAlves87

PR #377, opened 2026-05-14
fix(exec-shell): add zsh/dash/ash/ksh/fish to V1 POSIX shell recognizer

### LINKED ISSUE #366: [deep-review][duplication] D3: V1 and V2 exec-wrapper detectors have diverged on shell vocabulary and flag handling
## Summary

Two shell-wrapper detectors exist, both reachable today, with different shell vocabularies and different PowerShell flag handling.

- **V1**: `src/OpenClaw.Shared/ExecShellWrapperParser.cs` — string-based BFS, used by the **live** legacy exec-approval path (`SystemCapability.HandleRunAsync` → wrapper expansion at lines 374-389).
- **V2**: `src/OpenClaw.Shared/ExecApprovals/ExecShellWrapperNormalizer.cs` — argv-based single-level, used by the V2 pipeline that's currently inert (`ExecApprovalV2NullHandler` is the only registered handler).

## How they diverged

V2 added: `zsh`, `dash`, `ash`, `ksh`, `fish` to the recognized shell list.

V1 kept: `cmd`, `cmd.exe`, `powershell`, `powershell.exe`, `pwsh`, `pwsh.exe`, `bash`, `bash.exe`, `sh`, `sh.exe`.

V1 handles `-EncodedCommand` Base64 UTF-16LE decoding and `-Cmd:value` inline-separator forms. V2 does not (the inline-separator form was apparently dropped intentionally; the `-EncodedCommand` decoding lives elsewhere in V2 via `ExecCommandResolver`).

## Why it matters

Once V2 ships (the coordinator is wired in), the legacy V1 path stops firing — that's the design. But until then, **the live exec gate doesn't recognize zsh/dash/ash/ksh/fish wrappers**. An agent invoking `zsh -c "rm -rf"` reaches the policy evaluator as a literal `zsh -c "rm -rf"` string match. The default `defaultAction: deny` saves us, but the policy author who wrote `Allow: bash *` and `Deny: rm *` finds neither rule matches a zsh invocation reliably.

Security review filed this as E-6 (depth-4 short-circuiting under nested wrappers) — this duplication is the structural cause.

## Proposed fix

Three options, in order of preference:

1. **Promote the shared vocabulary** to a single static `ShellRegistry.RecognizedShells` set with the union of V1 and V2. Both parsers read from it. Eliminates drift by construction.
2. **Backport V2's expanded shell list to V1**. Simpler but doesn't address future drift.
3. **Ship V2** and delete V1. Highest payoff, but requires the coordinator wiring that's been pending.

Add an assertion test that walks both parsers' recognized shell lists and asserts they match (or that V1 is a subset of V2).

## Severity

Medium — security-relevant duplication; observable as "policy rules don't match zsh wrappers."

## Related

- C7 (validator gap on `execApprovals.set`) — defense-in-depth in the same area.
- D-V2-coordinator (out of scope for this review, but addressing the V2 rollout dissolves several items).

AlexAlves87:
## Summary

`ExecShellWrapperParser` (V1, the live exec-approval gate) only recognised `bash` and `sh` as POSIX shell wrappers. V2 (`ExecShellWrapperNormalizer`) already included `zsh`, `dash`, `ash`, `ksh`, and `fish`, but V1 never received the update.

As a result, a policy with `Allow: bash *` would not match a `zsh -c "..."` invocation — the command reached the policy evaluator unwrapped, so neither Allow nor Deny rules for known shells applied reliably.

## Changes

- `ExecShellWrapperParser.cs`: add `zsh`, `dash`, `ash`, `ksh`, `fish` to the POSIX shell branch
- `ExecShellWrapperParserTests.cs`: extend the existing `Expand_Bash_C_ExtractsPayload` theory with the five new shells

Fixes #366.

### DIFF
--- a/src/OpenClaw.Shared/ExecShellWrapperParser.cs
+++ b/src/OpenClaw.Shared/ExecShellWrapperParser.cs
@@ -113,7 +113,12 @@ internal static class ExecShellWrapperParser
         if (executable.Equals("bash", StringComparison.OrdinalIgnoreCase) ||
             executable.Equals("bash.exe", StringComparison.OrdinalIgnoreCase) ||
             executable.Equals("sh", StringComparison.OrdinalIgnoreCase) ||
-            executable.Equals("sh.exe", StringComparison.OrdinalIgnoreCase))
+            executable.Equals("sh.exe", StringComparison.OrdinalIgnoreCase) ||
+            executable.Equals("zsh", StringComparison.OrdinalIgnoreCase) ||
+            executable.Equals("dash", StringComparison.OrdinalIgnoreCase) ||
+            executable.Equals("ash", StringComparison.OrdinalIgnoreCase) ||
+            executable.Equals("ksh", StringComparison.OrdinalIgnoreCase) ||
+            executable.Equals("fish", StringComparison.OrdinalIgnoreCase))
         {
             for (var i = 1; i < tokens.Length; i++)
             {
--- a/src/OpenClaw.Shared/ExecShellWrapperParser.cs
+++ b/src/OpenClaw.Shared/ExecShellWrapperParser.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
+using OpenClaw.Shared.ExecApprovals;
 
 namespace OpenClaw.Shared;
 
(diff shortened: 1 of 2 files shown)


shanselman:
Thanks for tightening the V1 shell wrapper vocabulary. One gap before merge: the legacy parser uses `Path.GetFileName(tokens[0])` and only strips `.exe` for shells explicitly listed with `.exe` variants. This PR adds `zsh`, `dash`, `ash`, `ksh`, and `fish`, but not `zsh.exe`, `dash.exe`, `ash.exe`, `ksh.exe`, or `fish.exe`.

On Windows those are the executable names we are most likely to see from PATH, so the V1 path can still fail to unwrap the same shells this PR is trying to cover.

Suggested fix: either add the `.exe` variants in `ExecShellWrapperParser`, or switch this branch to the same normalized-basename behavior used by V2 (`ExecCommandToken.NormalizedBasename`) so future POSIX shell names do not need duplicate entries. Please add theory rows for the `.exe` variants too.

AlexAlves87:
Thanks for the review! Switched `TryExtractWrappedPayload` to use `ExecCommandToken.NormalizedBasename` — that strips path, lowercases, and removes the `.exe` suffix in one step, so all five new shells and their `.exe` variants are covered without enumerating them manually. Added theory rows for the `.exe` forms too.

PR #378, opened 2026-05-14
fix(node): route capture/record toasts through App.ShowToast

### LINKED ISSUE #342: [deep-review] C2: NodeService.ShowToast bypasses user sound preferences for capture/record toasts
## Summary

There are two `ShowToast` implementations in the tray with divergent behaviour. `App.ShowToast` honours the user's sound preference (None/Subtle/Default) and a 30-second deduplication window. `NodeService.ShowToast` does neither — it just fires `ToastContentBuilder().Show()`. Capture/record paths route through `NodeService.ShowToast`, so screen-record / camera-clip / camera-blocked / capture-complete toasts ignore the user's sound setting.

## Where

- Divergent: `src/OpenClaw.Tray.WinUI/App.xaml.cs:3823` (App.ShowToast) vs `src/OpenClaw.Tray.WinUI/Services/NodeService.cs:1786` (NodeService.ShowToast)
- Callers using the wrong path:
  - `Toast_ScreenCaptured` — `NodeService.OnScreenCapture` (~line 1367)
  - `Toast_ScreenRecordingComplete` — `NodeService.OnScreenRecord` (~line 1402)
  - `Toast_CameraBlocked` — `NodeService.OnCameraSnap`/`OnCameraClip` (~lines 1430, 1465)

## Why it matters

These toasts fire on privacy-sensitive operations (screen capture, camera). A user who silenced toasts because they're in a meeting or recording video gets unexpected sound. Worse: the toasts fire on attacker-controllable triggers (any `screen.snapshot` succeeds throttled-once-per-10s).

## Proposed fix

Delete `NodeService.ShowToast`. Route capture/record toasts through `App.ShowToast` (or equivalent helper that honours preferences and dedup).

Add a regression test: set sound preference to None, fire a capture, assert no `PlaySound` invocation (or use a mockable toast service and assert the routing).

## Note on root cause

This is one of several findings where copy/paste in `NodeService` diverged from the original in `App`. Filing a structural duplication issue separately to catalog the broader pattern.

## Severity

Bug — privacy/UX regression visible on every screen-capture invocation.

AlexAlves87:
## Summary

`NodeService` had its own `ShowToast` helper that called `ToastContentBuilder.Show()` directly. This bypassed the sound preference (`None` / `Subtle` / `Default`) and the 30-second deduplication window both implemented in `App.ShowToast`.

All screen-capture, screen-record, and camera toasts were affected — including toasts that fire on attacker-controllable triggers (`screen.snapshot` throttled once per 10 s).

## Changes

- Remove `NodeService.ShowToast`
- Add `NodeService.ToastRequested` event (`EventHandler<ToastContentBuilder>`)
- Replace all direct toast calls in capture/record handlers with `ToastRequested?.Invoke`
- Subscribe in `App` and delegate to `App.ShowToast`, so sound preferences and dedup are honoured

Fixes #342.

### DIFF
--- a/src/OpenClaw.Shared/NonFatalAction.cs
+++ b/src/OpenClaw.Shared/NonFatalAction.cs
@@ -1,0 +1,18 @@
+using System;
+
+namespace OpenClaw.Shared;
+
+public static class NonFatalAction
+{
+    public static void Run(Action action, Action<string> onError)
+    {
+        try
+        {
+            action();
+        }
+        catch (Exception ex)
+        {
+            onError(ex.Message);
+        }
+    }
+}
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -2680,16 +2680,7 @@ public partial class App : Application
     }
 
     private void OnNodeToastRequested(object? sender, Microsoft.Toolkit.Uwp.Notifications.ToastContentBuilder builder)
-    {
-        try
-        {
-            ShowToast(builder);
-        }
-        catch (Exception ex)
-        {
-            Logger.Warn($"Failed to show node toast: {ex.Message}");
-        }
-    }
+        => NonFatalAction.Run(() => ShowToast(builder), msg => Logger.Warn($"Failed to show node toast: {msg}"));
 
     private void OnNodeInvokeCompleted(object? sender, NodeInvokeCompletedEventArgs args)
     {
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -2680,7 +2680,16 @@ public partial class App : Application
     }
 
     private void OnNodeToastRequested(object? sender, Microsoft.Toolkit.Uwp.Notifications.ToastContentBuilder builder)
-        => ShowToast(builder);
+    {
+        try
+        {
+            ShowToast(builder);
+        }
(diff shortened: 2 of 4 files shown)


shanselman:
Good direction routing these through `App.ShowToast`, but this changes the failure semantics for node capture/record commands.

`NodeService.ShowToast()` used to swallow toast errors. After this PR, `ToastRequested?.Invoke(...)` can throw synchronously because `App.ShowToast(builder)` ultimately calls `builder.Show()` without a local try/catch. That means a toast-platform failure can now fail `screen.snapshot`, replace the intended camera `InvalidOperationException`, or mask the original capture/recording error.

Suggested fix: keep notification failures non-fatal at the app boundary:

```csharp
private void OnNodeToastRequested(object? sender, ToastContentBuilder builder)
{
    try
    {
        ShowToast(builder);
    }
    catch (Exception ex)
    {
        Logger.Warn($"Failed to show node toast: {ex.Message}");
    }
}
```

A small unit seam/test would be ideal, but the main requirement is that toast delivery failures must not change node command results.

AlexAlves87:
Added the try/catch so toast failures are non-fatal. Also fixed a thread-safety gap: the capture/record handlers fire `ToastRequested` from async workers, but `ShowToast` touches `_recentToastKeys` (a plain Dictionary) without synchronisation. Marshalled `OnNodeToastRequested` through `_dispatcherQueue`, consistent with how `OnSystemNotify` already handles this. Added a unit seam (`NonFatalAction.Run`) with three tests covering the non-fatal dispatch behaviour.

PR #379, opened 2026-05-14
fix(exec-approvals): reject Allow rules with absolute file paths

### LINKED ISSUE #347: [deep-review] C7: system.execApprovals.set validator accepts arbitrary file-path Allow rules
## Summary

`SystemCapability.HandleExecApprovalsSet` validates rule patterns against a dangerous-fragment blocklist (`Remove-Item`, `Format-`, `Shutdown`, etc.) but does not reject `Allow` rules whose pattern is a specific arbitrary file path (e.g. `"\\?\C:\evil.exe"`, `"C:\Users\Public\m.exe"`). The validator's broadest check covers all-wildcards and well-known shell prefixes (`"powershell *"`, `"cmd *"`), but a specific binary path passes.

## Where

- Validator: `src/OpenClaw.Shared/Capabilities/SystemCapability.cs:621-664` (`ValidateExecApprovalRules`)
- Routing seam: `src/OpenClaw.Shared/Capabilities/SystemCapability.cs:514` (`HandleExecApprovalsSet`)
- Constraints surfaced to callers: `:497-511`

## Why it matters

The local MCP HTTP server reaches this capability. A local untrusted process that obtained the MCP bearer token can rewrite the policy to whitelist an attacker-controlled binary path. The next `system.run` call to that binary is then allowed.

Two-step EoP:

1. Compromise the MCP token (multiple paths).
2. Rewrite policy to allow attacker binary.
3. Invoke `system.run` against the binary.

The validator's existing defense-in-depth shape is good — it just has a gap on `Allow`-with-specific-path.

## Proposed fix

Tighten `ValidateExecApprovalRules` so that:

- If `defaultAction != "prompt"`, `Allow` rules pointing at a specific executable path are rejected. Users must prompt-then-approve such a rule through the UI, not have it pushed by a remote `.set`.
- Reject `Allow` rules whose pattern, after glob-to-regex conversion, would match any path containing `\?\` (the long-path namespace prefix used by some bypass techniques).

Add tests:

- `Allow` rule with pattern `"C:\Users\Public\test.exe"` → reject.
- `Allow` rule with pattern `"Get-*"` (existing default) → still accepted.

## Severity

High — local EoP requiring MCP token compromise as a prerequisite; defense-in-depth gap.

AlexAlves87:
## Summary

`ValidateExecApprovalRules` accepted `Allow` rules whose pattern was an absolute file path (e.g. `C:\Users\Public\evil.exe`, `\?\C:\evil.exe`). A local process that obtained the MCP bearer token could use `system.execApprovals.set` to whitelist an attacker-controlled binary, then invoke it via `system.run` — a two-step local EoP.

The existing validator already blocked all-wildcard and dangerous-fragment patterns. This closes the gap on specific-path Allow rules.

## Changes

- `SystemCapability.cs`: add check in `ValidateExecApprovalRules` — rejects `Allow` patterns that start with a drive root (`X:\`, `X:/`) or a UNC/long-path prefix (`\`)
- `CapabilityTests.cs`: add `ExecApprovalsSet_RejectsAbsolutePathAllowRules` theory covering drive-rooted, forward-slash, UNC, and long-path namespace patterns

Fixes #347.

### DIFF
--- a/src/OpenClaw.Shared/Capabilities/SystemCapability.cs
+++ b/src/OpenClaw.Shared/Capabilities/SystemCapability.cs
@@ -647,11 +647,20 @@ public class SystemCapability : NodeCapabilityBase
             // A remote .set call should never be able to whitelist a specific binary
             // by path — that would be a two-step EoP (compromise MCP token → whitelist
             // attacker binary → invoke it). Legitimate rules name commands, not paths.
-            // Covers: drive-rooted paths (C:\, D:/), UNC paths (\\server\), and the
-            // Windows long-path namespace prefix (\\?\).
-            if (normalized.Length >= 3 &&
-                ((char.IsLetter(normalized[0]) && normalized[1] == ':' && (normalized[2] == '\\' || normalized[2] == '/')) ||
-                 normalized.StartsWith(@"\\", StringComparison.Ordinal)))
+            // Strip one layer of matching surrounding quotes first so quoted forms
+            // ("C:\evil.exe", 'C:\evil.exe') are caught alongside bare paths.
+            // Covers: drive-rooted paths (C:\, C:/), UNC/long-path (\\, //), and
+            // forward-slash UNC namespace forms (//server/share, //?/C:/evil.exe).
+            var unquoted = normalized;
+            if (normalized.Length >= 2 &&
+                ((normalized[0] == '"' && normalized[^1] == '"') ||
+                 (normalized[0] == '\'' && normalized[^1] == '\'')))
+                unquoted = normalized[1..^1];
+
+            if (unquoted.Length >= 3 &&
+                ((char.IsLetter(unquoted[0]) && unquoted[1] == ':' && (unquoted[2] == '\\' || unquoted[2] == '/')) ||
+                 unquoted.StartsWith(@"\\", StringComparison.Ordinal) ||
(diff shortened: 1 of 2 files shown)


shanselman:
This is the right security direction, but the validator still appears bypassable with quoted absolute paths.

`ValidateExecApprovalRules` trims and lowercases the pattern, then checks `normalized[0]` for a drive letter or `normalized.StartsWith(@"\\")`. A pattern like `"C:\Users\Public\evil.exe"` starts with a quote, so it is accepted even though it still whitelists a specific absolute binary path. The same applies to quoted UNC/long-path forms.

Suggested fix: strip one layer of matching surrounding quotes before the absolute-path check, and add test cases for at least:

- `"C:\Users\Public\evil.exe"`
- `'C:\Users\Public\evil.exe'`
- `"\\server\share\tool.exe"`
- `"\\?\C:\evil.exe"`

I would also consider forward-slash UNC/namespace forms if the command policy accepts them (`//server/share/tool.exe`, `//?/C:/evil.exe`).

AlexAlves87:
Thanks for catching the quoted-path bypass! Added quote stripping before the absolute-path check, and covered all the cases you mentioned — unquoted and quoted drive paths, UNC and long-path prefix forms, single-quoted variants, and the forward-slash equivalents.

PR #383, opened 2026-05-14
Tray menu (right-click): Fluent style tray-menu redesign

kenehong:
## Summary
 
image
 VarB variant of the WinUI 3 tray-menu redesign — an informative, Fluent-styled
 tray context menu (right click) for OpenClaw built fresh on top of `upstream/master` to 
avoid the perf regressions in the prior round.
 
 ## Key changes
 - **Brand header**: app icon (`Square44x44Logo`) replaces emoji glyph; 
Disconnect/Connect button moved here so the Gateway block stays clean.
 - **Gateway**: now a chevron-flyout row using the same card format as devices.
 Hover reveals a rich detail flyout (server version, auth mode, protocol, 
uptime, conn id, clients, pending approvals). Click still opens Connection 
settings.
 - **Devices**: flow directly under Gateway with no divider/section header. 
Device flyout uses Fluent FontIcon glyphs (`FluentIconCatalog`) and renders 
each capability as a card with bold title + caption commands list. Ellipse + 
theme brush replaces emoji status dot.
 - **Permissions**: top-level Action with hover flyout. WinUI `ToggleSwitch` 
rows with `Off`/`On` text label that updates live, matching the Windows 
Settings Accessibility page pattern.
 - **Sessions / Usage**: collapsed L1 rows with rich hover flyouts (mini 
progress bars, totals, providers, by-model).
 - **Pairing approval pending**: high-priority row placed above Gateway block.
 - **Companion Settings** moved to footer with `Ctrl+Alt+;` hotkey (Win+; 
conflicted with Windows emoji panel).
 - **Menu cleanup**: `Exit` → `Close` with Cancel (X) icon; removed Quick Send,
 Setup/Reconfigure, More(N).
 
 image
image
 
 ## Validation
 - `./build.ps1` — all projects build cleanly.
 - `OpenClaw.Shared.Tests`: **1548 passed**, 28 skipped.
 - `OpenClaw.Tray.Tests`: **1188 passed**.
 
 ## Scope
 Touches only the tray menu (`App.xaml.cs`, `TrayMenuWindow.xaml.cs`, 
`GlobalHotkeyService.cs`, `FluentIconCatalog.cs`, `FluentIconCatalogTests.cs`).
 No functional behavior changed.
 
 Diff: ~1900 insertions / ~517 deletions across 5 files.

### DIFF
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -972,6 +972,16 @@ public partial class App : Application
         {
             case "status": ShowStatusDetail(); break;
             case "reconnect": _ = _connectionManager?.ReconnectAsync(); break;
+            case "disconnect":
+                _ = _connectionManager?.DisconnectAsync();
+                _lastSessions = Array.Empty<SessionInfo>();
+                _lastNodePairList = null;
+                _lastDevicePairList = null;
+                _lastModelsList = null;
+                _agentEventsCache.Clear();
+                UpdateTrayIcon();
+                _hubWindow?.UpdateStatus(ConnectionStatus.Disconnected);
+                break;
             case "dashboard": OpenDashboard(); break;
             case "canvas": ShowCanvasWindow(); break;
             case "openchat": ShowChatWindow(); break;
@@ -1251,6 +1261,7 @@ public partial class App : Application
         var criticalBrush = (Microsoft.UI.Xaml.Media.Brush)resources["SystemFillColorCriticalBrush"];
         var secondaryText = (Microsoft.UI.Xaml.Media.Brush)resources["TextFillColorSecondaryBrush"];
         var captionStyle = (Style)resources["CaptionTextBlockStyle"];
+        var controlSecondaryFill = (Microsoft.UI.Xaml.Media.Brush)resources["ControlFillColorSecondaryBrush"];
 
         // ── Brand Header (non-interactive) — VarB: horizontal block,
         //     emoji 28 + "OpenClaw" SemiBold 18, padding 14,10,14,8.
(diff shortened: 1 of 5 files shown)


shanselman:
do we lose devices with this @kenehong ? 
image

kenehong:
No, they’re re-located with the gateway section @shanselman 
image

Device permissions are listed within the same menu as the other options.
image

shanselman (changes requested):
Thanks for the update here. This now clean-merges onto current `master` (including #384/#385), and CI is green, but I still think this needs changes before merge because it regresses a few tray-menu entry points.

Blocking issues I found in the current head (`508c7bb`):

1. The Gateway card advertises "Activate to open connection settings" and is wired with `action: "connection"`, but `OnTrayMenuItemClicked` still has no `case "connection"`. Clicking the gateway row therefore no-ops instead of opening Connection settings. Suggested fix: add `case "connection": ShowHub("connection"); break;` and ideally add a regression assertion that every tray-menu action emitted by `BuildTrayMenuPopup` has a handler or is explicitly flyout-only.

2. The visible Setup/Reconfigure tray entry is still gone. We recently fixed regressions around keeping/dismissing existing setup, and the tray menu needs to preserve an obvious path back to setup/reconfigure. Suggested fix: restore a visible Setup/Reconfigure row using the existing onboarding/config guard label logic and route it to the existing `"setup"` action.

3. QuickSend is still removed from the visible tray menu even though the handler/action path remains. If removal is intentional, we should call it out in the PR and remove/deprecate the dead path separately; otherwise restore the QuickSend row.

Non-blocking cleanup: the brand header duplicates the disconnect logic inline instead of reusing the `"disconnect"` action path, which makes future state-reset changes easier to miss. I would fix the blocking action/entry regressions first, then consider extracting the duplicate disconnect logic.

kenehong:
Thanks for the review — addressed all blocking + non-blocking points in 
4e18aee:
 
 **Blocking**
 - Added explicit `case "connection":` and `case "permissions":` in 
`OnTrayMenuItemClicked`. The default fallthrough was technically routing 
correctly via `ShowHub(action)`, but explicit cases make the wiring 
discoverable and prevent silent regressions.
 - Restored visible **Setup Guide / Reconfigure** row using 
`OnboardingExistingConfigGuard.HasExistingConfiguration()` to flip the label, 
dispatching the existing `"setup"` action.
 - Restored visible **QuickSend** row (`Menu_QuickSend` localized label → 
`"quicksend"` action).
 
 **Non-blocking (also done)**
 - Refactored the brand-header Disconnect/Connect button to call 
`OnTrayMenuItemClicked(this, "disconnect" | "reconnect")` instead of 
duplicating the cleanup inline. Future state-reset changes now live in one 
place.
 
 **Regression guard**
 - New test `BuildTrayMenuPopup_TopLevelActions_AllHaveExplicitHandlers` 
asserts that every static top-level action (`connection`, `disconnect`, 
`reconnect`, `permissions`, `setup`, `quicksend`, `companion`, `about`, `exit`)
 has an explicit `case` in the click handler. Caught the missing 
`"permissions"` case the moment I wrote it.
 
 Validation: build green; Tray 1189/1189; Shared 1548/1548.

shanselman (approved):
Rechecked the updated head on current master. The prior blockers are addressed: the Gateway card now has an explicit \connection\ handler, QuickSend is restored, Setup/Reconfigure is restored with existing-config label logic, and regression coverage was added for the top-level tray actions. Local validation passed: build, Shared tests 1572/1600 with 28 skipped, Tray tests 1210/1210.

PR #405, opened 2026-05-14
Visual refinement + bugs fix

kenehong:
**Visual refinements to bar**
- BuildMiniBar: simplified 2-tier usage colors (green < 95%, red >= 95%) with a hairline ControlStrokeColorDefault stroke; track uses ControlAltFillColorTertiary for better contrast against the bar fill.
image
image

**Visual Bugs**
- BuildSessionListCard: removed the leading status dot — it duplicated and visually collided with the bar color now that the bar conveys the same liveness signal.
- BuildDeviceCard: removed the inner chevron glyph that stacked on top of the outer flyout chevron drawn by AddFlyoutCustomItem (the result read as a confusing ':>' next to the device name).

**Permission Toggle**
- FluentIconCatalog.Permissions: switched key glyph (E72E) to shield (EA18) so the Permissions row matches the companion app in Settings.
- AddToggleItem: re-styled the permissions sub-flyout toggle rows to match Windows 11 Settings spacing — 16x10 padding, 14px Normal-weight title, 12px secondary description, larger icon, 'On/Off' state label next to the switch.
- AddPermToggle: each capability now carries a short description string (e.g. 'Allow webcam capture during sessions') rendered under the title.

**Interaction Bugs**
- Cascade UX: toggling a permission in the sub-flyout no longer collapses the cascade (HideCascade is skipped for perm-toggle|* actions), so users can flip several capabilities in a row. The Connect/Disconnect brand button intentionally still closes the cascade — flipping a gateway connection state needs a moment of focus shift and the user can immediately reopen the tray to confirm the new state.
- ApplyTrayMenuPreviewDataIfRequested: opt-in design-QA data injection gated on OPENCLAW_TRAY_PREVIEW_DATA=1, overriding any live gateway cache so the menu is reviewable without a real connection.
- TrayMenuWindow.xaml: tightened MenuPanel bottom padding (6,8 ->
  6,8,6,4) and shaved 2px from the SizeToContent buffer to remove odd whitespace beneath the Close row.

Validation: ./build.ps1 green; Shared tests 1548/1576 passed (28 skipped); Tray tests 1189/1189 passed.

### DIFF
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -1252,6 +1252,8 @@ public partial class App : Application
         // ToggleAction delegates; recreate the lookup each rebuild.
         _permToggleActions.Clear();
 
+        ApplyTrayMenuPreviewDataIfRequested();
+
         var isConnected = _currentStatus == ConnectionStatus.Connected;
         var statusText = LocalizationHelper.GetConnectionStatusText(_currentStatus);
 
@@ -1318,8 +1320,8 @@ public partial class App : Application
         ToolTipService.SetToolTip(brandBtn, isConnected ? "Disconnect from gateway" : "Connect to gateway");
         brandBtn.Click += (s, ev) =>
         {
-            OnTrayMenuItemClicked(this, isConnected ? "disconnect" : "reconnect");
             _trayMenuWindow?.HideCascade();
+            OnTrayMenuItemClicked(this, isConnected ? "disconnect" : "reconnect");
         };
         Grid.SetColumn(brandBtn, 2);
         brandGrid.Children.Add(brandBtn);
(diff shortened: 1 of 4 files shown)


shanselman:
LOVELY! Out of curiosity, what are the rules around which red and which green to pick when thinking about red-green colorblindness?  @kenehong

shanselman (approved):
Reviewed on top of current `master` and validated locally. CI is green, and local validation passed: `./build.ps1`, Shared tests 1572/1600 with 28 skipped, Tray tests 1210/1210.

Looks safe to merge from a code perspective. The tray-menu changes are mostly visual refinements plus the intended permissions-flyout interaction change.

Non-blocking nit: `ApplyTrayMenuPreviewDataIfRequested` currently overwrites sessions/usage whenever `OPENCLAW_TRAY_PREVIEW_DATA` is set, but the XML doc says real data takes precedence and preview values are only written when caches are empty/null. Either behavior is fine for design QA, but the comment should match the implementation.

kenehong:
In this case, the best approach is to provide information beyond just color. For example, we don’t rely on color alone to indicate an error—we also include an icon on top of it. In situations like this, we use additional indicators such as percentage values for usage, so users can understand the state even without distinguishing between red and green.

kenehong:
Also, since users might switch their display to grayscale, we always provide supporting information to make sure everything is still understandable.

PR #418, opened 2026-05-16
Rebuild Connection page (replaces Home)

bkudiess:
Rebuild of the Connection page from scratch — Home is gone, Connection is the new landing surface and absorbs the gateway/operator/node status that used to live across Home banners and Permissions.

## what's in it

- **4 modes** from a pure projection: Welcome / Cockpit / Recovery / AddGateway.
- **Status strip** with merged gateway+operator headline (`Connected`, `Connecting…`, `Can't reach gateway`, etc.) + glance chips (topology, hostname, clients, channels, ``) + dashboard launcher when connected.
- **Operator + Node cards** with Permissions-style icon + state-mapped status glyphs. Node card tints red when Node mode is off.
- **Saved gateways** always visible. Active row shows inline `OK Connected`; inactive rows get `[Connect]`; never an unsafe `Disconnect` on the wrong record.
- **Discovery** moved into the Gateways header; per-row `[Add]` pre-fills the Direct pane.
- **SSH tunnel** is per-gateway now (legacy global expander gone, `UseSshTunnel` still honored for back-compat).

## hard constraint

UI/UX only — no changes to `OpenClaw.Connection`, `GatewayRegistry`, `SettingsManager` save logic, pairing/auth, or token storage. The regression test pinning `ConnectAndWaitForDirectConnectOutcomeAsync` / `Task.Delay(15s)` / `RollbackDirectConnect` still passes.

## design system

- All Segoe Fluent glyphs through `FluentIconCatalog` (15 new constants).
- CDR-0001 applied: `System` → PC1 (`\uE839`).
- Status vocabulary aligned with `naming.md`.

## validation

`./build.ps1` green · Tray tests **990 / 990** · Shared tests **1620 / 1620** (28 pre-existing skipped) · Manual smoke OK.

image
image


shanselman:
Thank you @bkudiess — this looks great. I validated the updated branch locally, confirmed the app runs and connects, and merged it.

PR #424, opened 2026-05-16
Chat revamp: file attachments, abort, session switching, voice UI, mute sync, overlay disable

christineyan4:
## Summary

Comprehensive chat UI revamp developed through 8+ rounds of user testing. Covers file attachments, response abort, session/model switching, voice recording polish, mute synchronization, and overlay removal.

### Chat Foundations
- File attachment support with drag-and-drop and picker
- Response abort/cancel functionality
- Session switching and model selection dropdowns
- Keyboard shortcuts and focus management

### Voice Recording UI Polish
- Split voice button into stop (keeps transcript) and cancel (discards) actions
- Fixed audio level animation (WirePipelineEvents was never called in ListenOnceAsync)
- Compact recording pill indicator (red dot + waveform bars)
- Dropdown sizing fixes (model ComboBox explicit width)
- Focus management: cursor goes to end of text after voice input, Enter sends instead of re-recording
- Append mode for re-recording
- Increased recording timeout from 10s to 120s

### Mute Sync
- Speaker mute persists across app restarts (VoiceTtsEnabled setting)
- Mute state syncs between hub chat, tray chat, and Voice/Audio settings page
- App.SpeakerMuteChanged event for cross-surface synchronization

### Inline Test Voice UI
- Replaced standalone dialog with inline toggle panel in Voice settings
- Speech bubble transcript display, centered controls, minimal layout
- Fixed disposal leak (_inlineTestVoiceService)

### Voice Overlay Disable
- Disabled standalone VoiceOverlayWindow (code preserved as comments)
- All voice entry points redirect to inline chat voice mode
- Voice button removed from tray menu

### Hotkey Improvements
- Ctrl+Alt+Shift+V opens chat and starts recording (RetryTriggerVoice mechanism)
- Works from cold start (15 retries x 100ms for mount timing)

### Node Auto-Connect Fix
- Added TryConnectLocalNodeServiceAsync for local gateway restart case
- NodeConnector suppression for local gateways left NodeService disconnected on restart

### Merge
- Merged with latest master, resolved conflicts in App.xaml.cs (tray menu refactor)


shanselman:
Hi @christineyan4 — thank you for the chat revamp work here. This branch was green on CI, but after the recent #423/#419 landing it now conflicts with current `master` in the app/window/chat/voice surfaces (`App.xaml.cs`, `ChatPage.xaml.cs`, `VoiceSettingsPage.xaml.cs`, and `HubWindow.xaml.cs`).

Could you please rebase this onto latest `master` and adapt the changes to the new `AppState` / `IAppCommands` / service split? The parts I’d be extra careful with are chat session switching, voice UI/trigger state, and the HubWindow navigation/state ownership that changed in #423/#419.

Once it’s rebased and CI is green again, I’ll take another careful pass and help get it landed. Thanks!

shanselman:
Quick related note from today's large-list loading work: I opened #429 with a small reusable loading-state pattern for gateway-backed lists.

The important behavior is stale-while-refresh, but with actions/editing disabled while fresh data is actively loading. That lets pages keep useful existing content visible instead of flashing empty states, without letting users mutate stale rows.

This may be useful as the chat rewrite/refactor evolves too, especially for session/model/history/file lists or any attachment/history surfaces that can load slowly. Chat streaming will probably need append/stream-specific semantics, so I would treat #429 as a reusable pattern to adapt rather than a hard dependency.

shanselman:
Post-refactor triage note: this is still a meaningful chat/voice revamp, but it is currently dirty against master and overlaps the AppState/IAppCommands/service split that landed after the branch was authored. Please rebase/update against current master before review; I am not going to maintainer-resolve these conflicts because the branch needs to be re-expressed through the current architecture rather than mechanically patched.

christineyan4:
Have been offline today, but will make sure to get this rebased and take into account the new architecture first thing tomorrow morning. Thanks!!

shanselman (changes requested):
Thanks for the substantial chat/voice work here. I ran this through a Copilot-assisted review: local validation in an isolated PR worktree plus two independent adversarial review passes. The direction is exciting, but I don't think we should merge this yet because the new attachment/abort paths have a few correctness and reliability issues that are easier to fix before they become part of the main chat architecture.

Validation did pass locally:
- `./build.ps1` passed
- Shared tests passed: 1795 total, 0 failed
- Tray tests passed: 1086 total, 0 failed
- Shared/Tray `--no-restore` reruns passed

Requested changes before merge:

1. Please make file picking STA/UI-thread safe. `Win32FilePickerHelper` currently runs `IFileOpenDialog` inside `Task.Run`, which means an MTA thread-pool thread. `IFileOpenDialog` is an STA COM dialog and can fail or hang depending on shell extensions/machine configuration. Please use a dedicated STA thread or a supported WinUI picker initialized with the window handle.

2. Please add an attachment size limit and avoid synchronous unbounded reads. `ChatAttachment.FromFile` reads the whole file synchronously and base64-encodes it without a size guard. A large file can freeze the UI, consume huge memory, and exceed gateway/WebSocket payload limits. A hard max size with user-visible feedback would make this much safer.

3. Please make attachment display/history structured instead of local emoji text markers. Right now attachments appear to be represented in the optimistic local bubble and parsed back from `???` / `??` text lines. That means attachments can disappear after history reload or across surfaces, and normal user text can be misclassified. Please carry attachment metadata through the chat model/history path, or intentionally persist a durable textual placeholder if structured history support is out of scope.

4. Please tighten abort/run tracking. The review found risks around deferred abort state, failed abort suppression, and overlapping sends while a turn is active. At minimum, failed aborts should not permanently suppress a thread, and either sending while a turn is active should be disabled or the provider should track multiple active runs safely.

5. Please add meaningful tests for the new behavior. This PR touches 40 files and adds major user-facing chat/voice features, but the test changes look mostly like existing-test adjustments. Before merge, I'd like coverage for the core new paths: attachment send/reload behavior, file size rejection, abort suppression/recovery, voice stop vs cancel, model/session patching, and mute sync.

6. Please avoid bypassing the chat provider abstraction for attachments. `OpenClawChatRoot` downcasts to `OpenClawChatDataProvider` to send attachments, so other providers silently drop them. An interface overload or explicit capability interface would keep this extensible and easier to test.

None of this is meant to take away from the user-testing and UX progress in the branch. I think the PR is close in spirit, but because it introduces file IO, chat transport changes, voice state, and abort/session behavior all at once, I want these reliability and test gaps closed before landing it.

Copilot transparency: this review was Copilot-assisted, then consolidated and posted by me.

PR #429, opened 2026-05-17
Add async loading states for large lists

shanselman:
## Summary
- add a reusable `AsyncListLoadingState` for first-load, stale-while-refresh, and loaded-empty list states
- show explicit loading placeholders on Sessions, Cron, Usage, and Bindings instead of flashing empty states while gateway data is slow
- keep stale list data visible during refresh but disable editing/actions until fresh data arrives
- guard Usage cost updates so stale responses from an older period cannot overwrite the selected period

## Validation
- `./build.ps1`


shanselman:
Closing this for now because the branch is older/dirty and we are prioritizing stability work today.

The async loading-state idea is still useful, and we can come back to it later from a fresh branch against current master so it does not carry stale UI conflicts forward. Thanks for pushing this direction.

PR #444, opened 2026-05-18
Clean up credential source fallbacks

shanselman:
## Summary
- Remove the legacy settings Token/BootstrapToken fallback from interactive chat/dashboard credential resolution; these flows now resolve from the active GatewayRegistry record plus per-gateway identity only.
- Keep one-time legacy migration intact and retain the legacy root identity startup backstop with a warning for failed/old migrations.
- Tighten local gateway setup credential persistence so setup credentials merge through GatewayRegistry, preserve existing record metadata, and explicit clears remove registry tokens instead of preserving stale values.
- Add regression coverage for active-registry credential resolution and local setup registry-only credential persistence/clear behavior.

## Validation
- `./build.ps1`
- `dotnet test ./tests/OpenClaw.Connection.Tests/OpenClaw.Connection.Tests.csproj --no-restore`

### DIFF
--- a/src/OpenClaw.Connection/InteractiveGatewayCredentialResolver.cs
+++ b/src/OpenClaw.Connection/InteractiveGatewayCredentialResolver.cs
@@ -12,79 +12,45 @@ public static class InteractiveGatewayCredentialResolver
 {
     public static bool TryResolve(
         GatewayRegistry? registry,
-        string settingsDirectory,
         IDeviceIdentityReader identityReader,
-        string? effectiveGatewayUrl,
-        string? legacyToken,
-        string? legacyBootstrapToken,
         out InteractiveGatewayCredential? credential)
     {
-        ArgumentException.ThrowIfNullOrWhiteSpace(settingsDirectory);
         ArgumentNullException.ThrowIfNull(identityReader);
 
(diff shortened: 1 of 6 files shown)


shanselman:
Closing this one rather than trying to force it through today.

The credential-source cleanup direction is right, but this PR touches credential resolution, local setup persistence, and chat/dashboard auth fallback behavior. That is too close to pairing/auth migration to merge below a very high confidence bar. I do not think we are at 90% confidence on this branch as-is.

The safer path is to come back with a smaller PR that does one thing at a time:

1. remove legacy interactive credential fallbacks only after proving registry migration covers the affected cases;
2. keep local setup resume behavior explicit, especially when credentials live only in `GatewayRegistry`;
3. add focused regression tests for active-record-without-shared-token, root identity migration fallback, and setup resume after a partial local gateway run.

No shade on the goal — I agree with moving credential truth into `GatewayRegistry`. I just do not want to land this cleanup while there is still ambiguity around auth fallback and migration edge cases.

shanselman:
Closing for now because this credential/auth cleanup does not meet our 90% confidence bar as-is. We'll revisit as smaller, more focused PRs with explicit migration/setup-resume coverage.

PR #447, opened 2026-05-18
feat: add exec approval V2 prompt adapter interface

AlexAlves87:
## Summary

This adds the V2 prompt adapter contract needed before the coordinator (PR7) can be wired up. The interface decouples the coordinator from any UI implementation, and the null stub lets PR7 compile and be tested without a WinUI dependency.

## What changed

- `ExecApprovalV2PromptRequest` — sealed request type with the 8 fields the presenter needs (`DisplayCommand`, `Cwd`, `Host`, `Security`, `Ask`, `AgentId`, `ResolvedPath`, `SessionKey`). `AllowAlwaysPatterns` is intentionally excluded; it lives on `ExecApprovalEvaluation` and is the coordinator's responsibility to pass through after the decision.
- `IExecApprovalV2PromptHandler` — single `PromptAsync` method returning `Task<ExecApprovalDecision>`. Non-nullable contract; implementations fail-closed to `Deny` on any unhandled error.
- `ExecApprovalV2NullPromptHandler` — stub that always returns `Deny` synchronously via `Task.FromResult`. Never throws. Intended as the default until a real dialog implementation ships.

The `V2` infix avoids collision with the existing `IExecApprovalPromptHandler` / `ExecApprovalPromptRequest` types on the legacy path.

## Testing

```
dotnet test tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --filter "ExecApprovalV2Prompt"
```

12 tests, all passing. Coverage includes stub behavior, the `AllowAlwaysPatterns` exclusion verified via reflection, assembly placement, legacy type separation, and a guard that the null stub is not referenced from production `src/`.

## Notes

No production wiring in this PR. `SystemCapability` is untouched. The real dialog implementation is a separate slice after the coordinator lands.

### DIFF
--- a/src/OpenClaw.Shared/ExecApprovals/ExecApprovalPromptOutcome.cs
+++ b/src/OpenClaw.Shared/ExecApprovals/ExecApprovalPromptOutcome.cs
@@ -1,0 +1,13 @@
+namespace OpenClaw.Shared.ExecApprovals;
+
+/// <summary>
+/// Decision returned by <see cref="IExecApprovalV2PromptHandler"/>.
+/// Deny is the zero/default value so uninitialized instances fail-closed.
+/// </summary>
+public enum ExecApprovalPromptOutcome
+{
+    Deny = 0,
+    Allow,
+    AllowOnce,
+    AllowAlways,
+}
--- a/src/OpenClaw.Shared/ExecApprovals/ExecApprovalV2NullPromptHandler.cs
+++ b/src/OpenClaw.Shared/ExecApprovals/ExecApprovalV2NullPromptHandler.cs
@@ -1,0 +1,12 @@
+using System.Threading.Tasks;
+
+namespace OpenClaw.Shared.ExecApprovals;
+
+/// <summary>Default prompt handler stub: always denies. Never throws.</summary>
+public sealed class ExecApprovalV2NullPromptHandler : IExecApprovalV2PromptHandler
+{
+    public static readonly ExecApprovalV2NullPromptHandler Instance = new();
+
+    public Task<ExecApprovalDecision> PromptAsync(ExecApprovalV2PromptRequest request)
+        => Task.FromResult(ExecApprovalDecision.Deny);
+}
--- a/src/OpenClaw.Shared/ExecApprovals/ExecApprovalV2NullPromptHandler.cs
+++ b/src/OpenClaw.Shared/ExecApprovals/ExecApprovalV2NullPromptHandler.cs
@@ -1,3 +1,4 @@
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace OpenClaw.Shared.ExecApprovals;
@@ -7,6 +8,6 @@ public sealed class ExecApprovalV2NullPromptHandler : IExecApprovalV2PromptHandl
 {
     public static readonly ExecApprovalV2NullPromptHandler Instance = new();
 
-    public Task<ExecApprovalDecision> PromptAsync(ExecApprovalV2PromptRequest request)
-        => Task.FromResult(ExecApprovalDecision.Deny);
+    public Task<ExecApprovalPromptOutcome> PromptAsync(ExecApprovalV2PromptRequest request, CancellationToken cancellationToken = default)
+        => Task.FromResult(ExecApprovalPromptOutcome.Deny);
 }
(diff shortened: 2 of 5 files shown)


shanselman (changes requested):
Thanks for keeping this as a small preparatory slice. I like the direction of decoupling the V2 coordinator from WinUI, and I appreciate the null handler defaulting to Deny.

I want to request a few contract-shape changes before we merge this, because this is security-sensitive API surface and it will be much harder to change once the coordinator/dialog wiring lands:

1. Please add cancellation support to the prompt contract, e.g. `PromptAsync(ExecApprovalV2PromptRequest request, CancellationToken cancellationToken)`. Approval prompts need to be cancellable on request timeout, disconnect, app shutdown, or stale coordinator work. Cancellation should be treated as fail-closed / Deny by the caller.

2. Please avoid the fail-open enum default footgun. `ExecApprovalDecision` currently has `Allow = 0`, so `default(ExecApprovalDecision)` is Allow. For a prompt-facing contract, I would strongly prefer a prompt-specific decision/result where Deny is the zero/default value, or another shape that makes accidental defaults fail closed.

3. Please carry an audit/correlation id through the prompt request, consistent with the existing V2 handler's `correlationId`. The prompt is the user-decision point, so logs/telemetry should be able to tie the displayed approval back to the originating evaluation/request.

4. Please document the security semantics of `SessionKey`: who mints it, its lifetime/scope, what `null` means, and whether it is safe to display. This matters because it is likely to become part of AllowOnce/AllowAlways scoping.

5. Please tighten the `DisplayCommand` story. The comment says it arrives unsanitized and the presenter sanitizes before rendering; that should either point to a shared sanitizer/helper or make the hostile/raw nature explicit enough that future UI implementations cannot accidentally render spoofable control/BiDi text.

6. Please remove or harden `ProductionWiring_NullPromptHandler_NotReferencedInSrc`. The current source-tree string scan is brittle and will false-positive on comments/generated code/future legitimate wiring. If we keep it, it should skip bin/obj, be robust about repo-root discovery, and clearly state when it should be deleted.

I verified this locally in an isolated PR worktree:
- `./build.ps1` passed
- Shared tests with restore passed: 1807 total, 0 failed
- Tray tests with restore passed: 1044 total, 0 failed
- targeted `ExecApprovalV2Prompt` tests passed: 12 total, 0 failed
- Shared/Tray reruns with `--no-restore` passed after the test build was seeded

No production vulnerability is introduced here because this is not wired yet, but the interface shape should fail closed and carry cancellation/audit semantics before follow-on PRs depend on it.

shanselman:
Quick transparency note: the requested-changes review above was Copilot-assisted. I ran a dual-model adversarial review plus local validation, then posted the consolidated feedback.

The intent is constructive: this is a good small preparatory PR, and the requested changes are about getting the security-sensitive contract shape right before follow-on wiring depends on it.

AlexAlves87:
Thanks for the thorough review, and for the transparency note about the process behind it.

All six points addressed in the follow-up commit.

1. `CancellationToken` added to the interface and null stub. I used `= default` rather than a required parameter: no existing callers in this PR, and it's the standard .NET convention. Happy to remove the default if you prefer the stricter form.
2. `ExecApprovalPromptOutcome` is a new enum with `Deny = 0`, separate from `ExecApprovalDecision`. The shared evaluator type is untouched.
3. `CorrelationId` added as a `required string` on the prompt request, matching the pattern from `IExecApprovalV2Handler`.
4. `SessionKey` now documents who mints it, what `null` means, its session scope, and that it is not safe to display.
5. `DisplayCommand` doc updated to call out control characters and BiDi overrides explicitly.
6. The production wiring test now skips `bin/` and `obj/`, relies on the `.slnx` file for repo-root discovery, and has a deletion comment. One known limitation: the string scan will false-positive on a `//` comment in a src file that mentions `ExecApprovalV2NullPromptHandler`. Fixing that properly needs a C# parser; the deletion comment is the right defence for that case rather than a fragile heuristic.

16 tests, all green.

shanselman:
Thanks @AlexAlves87 ? this is a clean foundation slice. Keeping the V2 prompt adapter contract in `OpenClaw.Shared` and defaulting the null handler to a synchronous fail-closed `Deny` gives the coordinator a safe seam without pulling WinUI into shared code.

I validated this two ways because the branch was behind the newly merged chat-tool work:

- On the PR branch, the focused `ExecApprovalV2PromptAdapterTests` pass.
- On a local merged tree with current `origin/master`, full validation passed:
  - `./build.ps1`
  - `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore`: 1798 passed / 28 skipped
  - `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore`: 1122 passed

No blockers from my review. I?m going to squash-merge this now.

PR #454, opened 2026-05-18
fix(wizard): restore static 'Configuring Gateway' heading; step title back in card

indierawk2k2:
## Summary

Reverts the heading layout in the embedded Gateway Wizard. The earlier QoL pass (ed9be7e + 656d46f) over-corrected: it removed both static "Configuring Gateway" labels (the V2 host page's and the wizard's) and promoted the per-step title to the top, leaving the page without its recognizable identity.

This PR restores the wizard's static top heading to "Configuring Gateway" (`Onboarding_Wizard_Title`) and moves the per-step / lifecycle title back into the gateway config card as a bold sub-heading above the message. The V2 host page's own duplicate heading remains suppressed when the wizard is embedded — so we end up with exactly one "Configuring Gateway" label, not two and not zero.

## Changes

`src/OpenClaw.Tray.WinUI/Onboarding/GatewayWizard/GatewayWizardPage.cs`

- Top heading: static `Onboarding_Wizard_Title` (was: dynamic `displayTitle` with fallback)
- Card: re-added bold `displayTitle` TextBlock above the message
- Top heading no longer needs `MaxLines=2` + `CharacterEllipsis` (fixed short string)
- Card sub-heading inherits the same defensive `MaxLines=3` + `CharacterEllipsis` cap that 656d46f added, since the untrusted server-supplied `stepTitle` now lives there in the `active` state (defends against long titles pushing content past `MaxHeight(350)` and long unbroken tokens clipping horizontally inside the `HorizontalScrollMode=Disabled` ScrollView)

## Review

Ran a Hanselman dual-model review (Opus + Codex). HIGH-consensus finding: the in-card title needed the same long-server-title protection that the previous top heading had — addressed before merge. Other findings (top heading locale wrapping; phantom line on empty `displayTitle`; reduced visibility of lifecycle status in card sub-heading) were either pre-existing conditions or by-design per the requested layout.

## Validation

- ✅ `./build.ps1` — all 4 projects build clean
- ✅ `OpenClaw.Shared.Tests` — 1776 passed, 28 skipped
- ✅ `OpenClaw.Tray.Tests` — 1089 passed

### DIFF
--- a/src/OpenClaw.Tray.WinUI/Onboarding/GatewayWizard/GatewayWizardPage.cs
+++ b/src/OpenClaw.Tray.WinUI/Onboarding/GatewayWizard/GatewayWizardPage.cs
@@ -775,7 +775,14 @@ public sealed class GatewayWizardPage : Component<GatewayWizardState>
             TextBlock(LocalizationHelper.GetString("Onboarding_Wizard_Title"))
                 .FontSize(28)
                 .SemiBold()
-                .HAlign(HorizontalAlignment.Center),
+                .HAlign(HorizontalAlignment.Center)
+                .TextWrapping()
+                .Set(t =>
+                {
+                    t.MaxLines = 2;
+                    t.TextTrimming = Microsoft.UI.Xaml.TextTrimming.CharacterEllipsis;
+                    t.TextAlignment = Microsoft.UI.Xaml.TextAlignment.Center;
+                }),
 
             Border(
                 ScrollView(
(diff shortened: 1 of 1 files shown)


shanselman:
﻿Thanks, @indierawk2k2. This is the right small UX correction: keep one stable "Configuring Gateway" page identity while putting the step-specific title back inside the card where it belongs.

I ran a dual review and local validation. Both reviewers caught the same localization/layout edge case: the restored static heading still needs wrapping/trimming because localized strings can be longer than the English text. I pushed a small maintainer patch for that, and the PR is green now.

PR #455, opened 2026-05-18
Keep WSL gateway alive across tray restarts and Windows logon

indierawk2k2:
## What

Makes the OpenClawGateway WSL2 distro stay reachable across tray restarts and Windows logon. The keepalive now persists across the tray's lifetime so the user's gateway is reachable any time the tray icon is running (or about to run).

## Why

Even after a successful install of the easy-button local gateway, the gateway became unreachable in two real scenarios:

1. **Windows reboot.** Tray autostart at logon never re-armed the keepalive — only the onboarding flow did. ~20 s after the first `/health` probe, the WSL2 VM idled out (vmIdleTimeout), systemd died with it, and the gateway with that.
2. **Tray crash / restart.** `WslDistroKeepAlive` registered `AppDomain.CurrentDomain.ProcessExit += StopAll`, killing the keepalive whenever the tray exited.

Symptoms reported in the wild include `connecting…` that never resolves after rebooting Windows. `loginctl enable-linger openclaw` does not save us here because the WSL VM itself is what shuts down — linger only protects user services across logout *while the VM is running*.

## How

`WslDistroKeepAlive` becomes a real lifecycle primitive:

- **Marker file** at `%LOCALAPPDATA%/OpenClawTray/wsl-keepalive/<distro>.json` containing `distroName + PID + StartTimeUtc + processName`, with atomic `File.Move(..., overwrite: true)`.
- **Adoption** on every `EnsureStarted`: if the marker references a live process whose name + start time match (500 ms tolerance), adopt instead of spawning a duplicate.
- **Anti-PID-recycle**: `Stop` only kills when marker identity matches; otherwise just deletes the marker. Defensive `SafeStartTimeUtc` for fast-exit `InvalidOperationException`.
- **Detached spawn**: no stdio redirect, no `ProcessExit` cleanup. The keepalive outlives the tray.
- `App.OnLaunched` calls `TryEnsureLocalGatewayKeepAliveAsync` before `InitializeGatewayClient`. Bounded 5 s `wsl --list` probe; never blocks tray startup. If the user is no longer in local-gateway mode, the helper also calls `Stop` to release any stale keepalive (no zombie VM after mode switch).
- `LocalGatewayLifecycleManager.RepairAsync`: `Stop` before terminate, `EnsureStarted` after the gateway is healthy.
- `LocalGatewayLifecycleManager.RemoveAsync`: `Stop` before terminate so the marker never survives `--unregister`.
- Internal `IWslKeepAliveProcessHost` test seam plus `__TestMarkerDirectory` override to keep unit tests off real `wsl.exe`.

## Review

Ran the Hanselman dual-model adversarial review (Opus + Codex). High-consensus fixes already applied in this PR:

| Issue | Fix |
| --- | --- |
| `File.Move` without `overwrite: true` orphans keepalive in concurrent-launch races | Single atomic `File.Move(temp, dest, overwrite: true)` |
| No teardown on local→remote mode switch left zombie VM pinned | `TryEnsureLocalGatewayKeepAliveAsync` calls `Stop` when URL is non-local |
| `process.StartTime` can throw on fast-exit | `SafeStartTimeUtc` falls back to `UtcNow` |
| 50 ms identity tolerance too tight for cross-process roundtrip | Widened to 500 ms (still well below PID-recycle window) |

Skipped after analysis: cross-process mutex (tray already enforces single-instance via named `Mutex` in `OnLaunched`); `StopAll` lock scope (only runs from uninstall); filename-collision sanitizer (production distro name is hard-coded `OpenClawGateway`).

## Out-of-scope observation

Codex flagged a sibling concern: `SshTunnelService.EnsureStarted` is never called from `App.OnLaunched` either. Users on SSH-tunnel mode with autostart will see the tray come up but the tunnel will not arm until they touch the connection. This is a pre-existing gap not caused by this change — happy to file as a follow-up if useful.

## Validation (run in clean worktree based on origin/master)

- `./build.ps1` — all 4 projects built clean
- `dotnet test ./tests/OpenClaw.Shared.Tests/...` — **1776 passed / 28 skipped / 0 failed**
- `dotnet test ./tests/OpenClaw.Tray.Tests/...` — **1100 passed / 0 failed** (includes 11 new `WslDistroKeepAliveTests` covering adoption, recycled-PID, dead-PID, corrupt marker, Stop/StopAll, sanitization, null inputs)

### DIFF
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -2004,45 +2004,12 @@ public partial class App : Application, OpenClawTray.Services.IAppCommands
 
             if (string.IsNullOrWhiteSpace(distroName)) return;
 
-            // Backoff schedule for the cold-logon case. Each entry is the delay BEFORE
-            // the next attempt. After the last entry we fall through to a defensive
-            // spawn regardless of probe outcome — wsl.exe itself will exit quickly if
-            // the distro genuinely does not exist, which is preferable to silently
-            // never arming the keepalive.
-            TimeSpan[] retryDelays = [TimeSpan.Zero, TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(15), TimeSpan.FromSeconds(60)];
             var runner = new WslExeCommandRunner(new AppLogger(), defaultTimeout: TimeSpan.FromSeconds(4));
-
-            for (var attempt = 0; attempt < retryDelays.Length; attempt++)
-            {
-                if (retryDelays[attempt] > TimeSpan.Zero)
-                    await Task.Delay(retryDelays[attempt]);
-
-                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
-                try
-                {
-                    var distros = await runner.ListDistrosAsync(cts.Token);
-                    if (!distros.Any(d => d.Name.Equals(distroName, StringComparison.OrdinalIgnoreCase)))
-                    {
-                        Logger.Warn($"[WslKeepAlive] Configured local-gateway distro '{distroName}' not present; skipping keepalive.");
-                        return;
-                    }
-
-                    WslDistroKeepAlive.EnsureStarted(distroName, new AppLogger());
-                    return;
-                }
-                catch (OperationCanceledException)
-                {
-                    Logger.Warn($"[WslKeepAlive] wsl --list probe timed out (attempt {attempt + 1}/{retryDelays.Length}); will retry.");
-                }
-            }
-
-            // All probes timed out. WSL/LxssManager may be wedged, mid-update, or just
-            // very slow on this cold-logon. Attempt the spawn anyway — the wsl.exe
-            // child exits quickly if the distro truly does not exist, and adoption on
-            // the next launch will either recover (if the distro is fine) or skip
-            // (if it really is gone).
-            Logger.Warn("[WslKeepAlive] wsl --list never returned; spawning keepalive defensively.");
-            WslDistroKeepAlive.EnsureStarted(distroName, new AppLogger());
+            await WslKeepAliveStartupArmer.ArmAsync(
+                distroName,
+                cancellationToken => runner.RunAsync(["--list", "--verbose"], cancellationToken),
+                () => WslDistroKeepAlive.EnsureStarted(distroName, new AppLogger()),
+                Logger.Warn);
         }
         catch (Exception ex)
         {
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -767,6 +767,11 @@ public partial class App : Application, OpenClawTray.Services.IAppCommands
 
         // Initialize connections — always create operator client for UI data,
         // additionally create node service for gateway node mode or local MCP.
+        // Before any connection attempt, re-arm the WSL keepalive so the local
+        // gateway VM stays up across tray restarts and across the 20s WSL
+        // vmIdleTimeout window observed on some hosts. The keepalive runs detached
+        // from the tray's lifetime — see WslDistroKeepAlive in LocalGatewaySetup.cs.
+        await TryEnsureLocalGatewayKeepAliveAsync();
         InitializeGatewayClient();
 
         // Pre-warm chat window (WebView2 init takes 1-3s, do it now so left-click is instant)
@@ -1958,6 +1963,98 @@ public partial class App : Application, OpenClawTray.Services.IAppCommands
     {
         if (_suppressNodeDuringSetup) return false;
         return _settings?.EnableNodeMode == true || _settings?.EnableMcpServer == true;
+    }
+
+    /// <summary>
+    /// Re-arms the WSL keepalive process for the local OpenClawGateway distro on every
+    /// tray launch when the active gateway is local. The keepalive runs detached from
+    /// the tray (see <see cref="WslDistroKeepAlive"/>) so the WSL2 VM stays up even after
+    /// the tray exits — addressing the vmIdleTimeout window where Windows tears down the
+    /// distro (and therefore systemd + the gateway service) ~20s after the last
+    /// interactive wsl.exe session ends.
+    /// </summary>
+    /// <remarks>
+    /// Best-effort: this method swallows every failure. The tray must always reach its
+    /// chrome initialization regardless of WSL state.
+    /// </remarks>
+    private async Task TryEnsureLocalGatewayKeepAliveAsync()
+    {
+        try
+        {
+            if (_settings is null) return;
+
+            var gatewayUrl = _settings.GetEffectiveGatewayUrl();
+            var distroName = await ResolveLocalGatewayDistroNameAsync();
+
+            if (string.IsNullOrWhiteSpace(gatewayUrl) || !LocalGatewayUrlClassifier.IsLocalGatewayUrl(gatewayUrl))
+            {
+                // User is no longer using the local gateway (mode switch since the last
+                // launch). Tear down any keepalive marker that a prior session left
+                // behind so the WSL VM is allowed to idle out; we never reach
+                // EnsureStarted in this branch.
+                if (!string.IsNullOrWhiteSpace(distroName))
+                    WslDistroKeepAlive.Stop(distroName, new AppLogger());
+                return;
+            }
+
+            if (string.IsNullOrWhiteSpace(distroName)) return;
+
+            // Bounded probe: never let a hanging `wsl --list` block tray startup. If WSL
+            // is broken / mid-update, log + skip — the next launch will retry.
+            using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+            var runner = new WslExeCommandRunner(new AppLogger(), defaultTimeout: TimeSpan.FromSeconds(4));
+            try
+            {
+                var distros = await runner.ListDistrosAsync(cts.Token);
+                if (!distros.Any(d => d.Name.Equals(distroName, StringComparison.OrdinalIgnoreCase)))
+                {
+                    Logger.Warn($"[WslKeepAlive] Configured local-gateway distro '{distroName}' not present; skipping keepalive.");
+                    return;
+                }
+            }
+            catch (OperationCanceledException)
+            {
+                Logger.Warn("[WslKeepAlive] wsl --list timed out; skipping keepalive (will retry next launch).");
+                return;
+            }
+
+            WslDistroKeepAlive.EnsureStarted(distroName, new AppLogger());
+        }
+        catch (Exception ex)
+        {
+            Logger.Warn($"[WslKeepAlive] Startup keepalive failed (non-fatal): {ex.Message}");
+        }
+    }
+
+    /// <summary>
+    /// Resolves the WSL distro name to keep alive. Prefers the value persisted by
+    /// onboarding in <c>setup-state.json</c> so the keepalive always targets the distro
+    /// the user actually installed. In DEBUG / test builds, an
+    /// <c>OPENCLAW_WSL_DISTRO_NAME</c> environment override is honored to match
+    /// <see cref="LocalGatewaySetupRuntimeConfiguration.FromEnvironment"/>. Falls back
+    /// to the upstream default <c>OpenClawGateway</c>.
+    /// </summary>
+    private async Task<string?> ResolveLocalGatewayDistroNameAsync()
+    {
+        try
+        {
+            var store = new LocalGatewaySetupStateStore();
+            var state = await store.LoadAsync();
+            if (state is not null && !string.IsNullOrWhiteSpace(state.DistroName))
+                return state.DistroName;
+        }
+        catch (Exception ex)
+        {
+            Logger.Warn($"[WslKeepAlive] Failed to read setup-state.json: {ex.Message}");
+        }
+
+#if DEBUG || OPENCLAW_TRAY_TESTS
+        var envOverride = Environment.GetEnvironmentVariable(LocalGatewaySetupRuntimeConfiguration.DistroNameVariable);
+        if (!string.IsNullOrWhiteSpace(envOverride))
+            return envOverride;
+#endif
+
+        return "OpenClawGateway";
     }
 
     // The pre-unification ShouldInitializeNodeService(GatewayRecord, string) overload
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -767,11 +767,12 @@ public partial class App : Application, OpenClawTray.Services.IAppCommands
 
         // Initialize connections — always create operator client for UI data,
         // additionally create node service for gateway node mode or local MCP.
-        // Before any connection attempt, re-arm the WSL keepalive so the local
-        // gateway VM stays up across tray restarts and across the 20s WSL
-        // vmIdleTimeout window observed on some hosts. The keepalive runs detached
-        // from the tray's lifetime — see WslDistroKeepAlive in LocalGatewaySetup.cs.
-        await TryEnsureLocalGatewayKeepAliveAsync();
+        // Re-arm the WSL keepalive so the local gateway VM stays up across tray
+        // restarts and across the 20s WSL vmIdleTimeout window observed on some
+        // hosts. Fire-and-forget on a background task so a slow LxssManager at
+        // cold logon never delays InitializeGatewayClient. The keepalive itself
+        // runs detached from the tray — see WslDistroKeepAlive in LocalGatewaySetup.cs.
+        _ = Task.Run(TryEnsureLocalGatewayKeepAliveAsync);
         InitializeGatewayClient();
 
         // Pre-warm chat window (WebView2 init takes 1-3s, do it now so left-click is instant)
@@ -1974,8 +1975,12 @@ public partial class App : Application, OpenClawTray.Services.IAppCommands
     /// interactive wsl.exe session ends.
     /// </summary>
     /// <remarks>
-    /// Best-effort: this method swallows every failure. The tray must always reach its
-    /// chrome initialization regardless of WSL state.
+    /// Best-effort, fire-and-forget. Cold Windows logon is exactly when LxssManager is
+    /// slow to come up, so a hanging <c>wsl --list</c> probe must never block tray
+    /// startup. We retry with bounded backoff; if every probe times out, we still try
+    /// the spawn and let <c>wsl.exe</c> itself decide whether the distro exists —
+    /// missing a keepalive arm at cold logon is the exact bug this method exists to
+    /// prevent (see PR review feedback).
     /// </remarks>
     private async Task TryEnsureLocalGatewayKeepAliveAsync()
     {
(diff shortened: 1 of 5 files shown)


indierawk2k2:
Heads-up @shanselman — this is the WSL keepalive lifecycle fix we discussed for the `connecting` post-reboot issue your friend hit. Based on latest master (`cd4acf0`); full validation (build + 1776 Shared / 1100 Tray tests) green. Reviewer assignment via fork is blocked by repo permissions, hence the mention.

Emrullah007:
@indierawk2k2 Thank you for resolving the issue so quickly. I’ll test it again later after doing a fresh installation.

shanselman:
﻿�� **Bot review: PR #455 — WSL gateway keepalive**

I took a deeper look at this one because it changes the lifecycle contract for the local WSL gateway. Overall: **this is solving a real problem, and the proposed primitive is mostly the right shape, but I think this should get a small revision before merge.**

### Why this is a real problem

The core diagnosis checks out. WSL2 can shut down its VM after the idle timeout when there are no active Linux processes. When that happens, `systemd` and the local gateway service disappear with the VM. `loginctl enable-linger` is not enough to prevent this, because linger is a Linux/systemd concept inside the VM; it does not stop the Windows-side WSL host from deciding the VM is idle.

I also checked the `.wslconfig` alternative. Microsoft documents `vmIdleTimeout` as a global WSL2 VM setting under `%UserProfile%\.wslconfig`, applying across all WSL2 distributions. So changing `vmIdleTimeout` globally would be too broad for this app: it would affect OpenClaw, Ubuntu, Docker-related distros, and anything else the user runs under WSL2. A scoped keepalive for the OpenClaw gateway distro is the better product shape.

### What looks right

- A long-running in-distro process like `sleep` is a valid way to keep the WSL2 VM from becoming idle.
- Scoping it to the OpenClaw gateway distro is better than editing global WSL config.
- Re-arming on tray launch directly targets the Windows logon / tray restart symptom.
- The marker-file adoption design is directionally good: PID plus start-time matching is the right defense against PID reuse.
- The tests cover the main marker/adoption/recycle/corruption paths well.

### Local validation

I validated this PR in an isolated PR worktree:

- `./build.ps1` — passed
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj` — passed, 1804 total / 0 failed / 28 skipped
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj` — passed, 1100 total / 0 failed
- no-restore reruns for Shared and Tray — passed

### Requested changes before merge

#### 1. Don’t skip keepalive arming on a cold-logon `wsl --list` timeout

`TryEnsureLocalGatewayKeepAliveAsync` currently awaits `wsl --list` before `InitializeGatewayClient()`, and if the probe times out it logs and returns:

```csharp
catch (OperationCanceledException)
{
    Logger.Warn("[WslKeepAlive] wsl --list timed out; skipping keepalive (will retry next launch).");
    return;
}
```

That is risky because cold Windows logon is exactly when WSL/LxssManager startup can be slow. If `wsl --list` takes longer than the 5s bound, this PR can still miss the keepalive arm, and the local gateway can idle out shortly afterward — the same class of issue this PR is meant to fix.

Suggested fix:

- Do not block `InitializeGatewayClient()` on the WSL list probe.
- Arm/retry the keepalive in the background with bounded retry/backoff.
- If the distro existence probe times out, either retry later or attempt the keepalive spawn and let `wsl.exe -d <distro>` fail naturally if the distro truly does not exist.

This is the main merge-blocking concern for me because it undermines the target scenario.

#### 2. Prove or harden the “detached survives tray exit” behavior for packaged/MSIX runs

The implementation uses `Process.Start` with no redirected stdio and then disposes the process handle. That may be enough in many desktop contexts, but the load-bearing claim is stronger: the `wsl.exe` relay must survive when the packaged WinUI tray exits.

Please either:

- add/manual-record a packaged-app validation showing the keepalive PID remains alive after exiting the tray and after the WSL idle window elapses, or
- use a more explicit detach/breakaway mechanism if packaged process/job semantics kill the child on app exit.

The unit tests are good, but they fake the host process and cannot prove this part. This behavior is central to the PR.

#### 3. Make spawn + marker write transactional

Right now, if `host.Spawn(distroName)` succeeds but `WriteMarker(...)` fails, the app can leave behind an untracked keepalive process. Future launches will not adopt it because no marker exists, and future stop/remove flows cannot clean it up through the marker path.

Suggested fix: if marker write fails after spawn, kill the just-spawned PID before returning/logging. That keeps the durable marker and the actual process in sync.

### Non-blocking follow-ups worth filing

- Re-arm on Windows resume/modern standby, not only `OnLaunched`. Sleep/resume can tear down WSL without relaunching the tray.
- Stop the keepalive immediately when the user switches from local gateway mode to remote while the tray is still running. The current launch-time cleanup helps on next launch, but not necessarily live mode changes.
- Consider fallback cleanup if the marker is missing/corrupt. Current `Stop` is deliberately marker-dependent, which is safe for avoiding wrong-process kills, but it can leave the VM pinned if the marker is damaged.
- Document the expected power/resource tradeoff: keeping the local gateway available means intentionally keeping the WSL VM alive while local mode is enabled.
- Document uninstall behavior. A detached keepalive may outlive an MSIX uninstall until reboot/manual cleanup unless there is another cleanup path.

### Bottom line

I like the direction. This is not churn; it is a real reliability fix for the local gateway experience. But I would revise the cold-logon timeout behavior before landing, and I would explicitly verify the packaged-app child-process survival assumption because that is the core mechanism this PR depends on.

indierawk2k2:
Thanks for the deep review @shanselman — addressed all three blocking items in `3303a01`. Master is unchanged since the PR opened, so still based on the latest tip.

### 1. Cold-logon `wsl --list` timeout no longer skips arming
`OnLaunched` now calls `_ = Task.Run(TryEnsureLocalGatewayKeepAliveAsync)` so the keepalive arm never blocks `InitializeGatewayClient()`. The helper itself loops with backoff `[0s, 5s, 15s, 60s]`; if every probe times out we fall through and spawn the keepalive defensively — `wsl.exe` will exit quickly on its own if the distro genuinely doesn't exist, which is the correct degradation versus silently never arming.

### 2. MSIX/packaged "detached survives tray exit"
Replaced the bare `Process.Start` with a P/Invoke `CreateProcessW` path that requests `CREATE_BREAKAWAY_FROM_JOB | DETACHED_PROCESS | CREATE_NO_WINDOW` (`Win32DetachedSpawn` helper). The packaged tray's host job kills child processes on app exit unless the child breaks away from the job. Fallback chain:

1. `CreateProcess` with breakaway → preferred path for MSIX.
2. If breakaway returns `ERROR_ACCESS_DENIED` (host job forbids breakaway), retry without the flag — keepalive at least arms while the tray runs.
3. If `CreateProcess` fails outright, fall back to `Process.Start` so we don't regress unpackaged dev runs.

I'll do the manual `packaged-tray-exits, keepalive PID persists past vmIdleTimeout` validation once we have an MSIX build of this branch and post the result on the PR. Happy to gate merge on that if you'd prefer.

### 3. Transactional spawn + marker write
`WslDistroKeepAlive.EnsureStarted` now kills the just-spawned PID if `WriteMarker` throws, so we never leak an untracked orphan that adoption/Stop cannot reach. New test `EnsureStarted_MarkerWriteFails_KillsOrphanSpawn` covers it.

### Test infrastructure clean-up (caught by your changes)
The Repair/Remove additions now invoke `WslDistroKeepAlive` internally. The existing `LocalGatewayLifecycleManager_*` tests were therefore about to spawn real `wsl.exe` and pollute LocalAppData. Fixed by:

- Promoting `FakeKeepAliveHost` to internal and adding `KeepAliveIsolation` (an `IDisposable` that swaps in a no-op host + temp marker dir).
- Wrapping the two lifecycle tests with `using var _ = new KeepAliveIsolation();`.
- Adding a shared `WslDistroKeepAliveSerial` xunit collection with `DisableParallelization = true` so the global `__TestHost` / `__TestMarkerDirectory` seams can't interleave across test classes.

### Non-blocking follow-ups
I'll file separate issues for:
- Re-arm on Windows resume / modern standby (`SystemEvents.PowerModeChanged`).
- Live mode-switch teardown while the tray is still running (today we only catch it at next launch).
- Fallback cleanup when the marker is missing/corrupt — needs a careful design that doesn't risk killing unverified PIDs.
- Documentation for the power/resource tradeoff and the MSIX-uninstall behavior.

### Re-validation in the clean worktree
- `./build.ps1` — clean
- Shared.Tests `--no-restore` — 1776 passed / 28 skipped / 0 failed
- Tray.Tests `--no-restore` — 1101 passed / 0 failed (now includes `EnsureStarted_MarkerWriteFails_KillsOrphanSpawn`)

shanselman:
﻿🤖 **Follow-up with a concrete fix path**

Thanks for the fast update — this is very close. The new commit addressed the right three areas in spirit:

- `OnLaunched` no longer blocks `InitializeGatewayClient()`.
- There is bounded retry/backoff for cold logon.
- `CREATE_BREAKAWAY_FROM_JOB` is now attempted for packaged/MSIX survivability.
- marker write failure now kills the just-spawned keepalive, which fixes the untracked orphan issue.

There is one important implementation trap left: `ListDistrosAsync()` is the wrong API for the retry loop because it collapses all `wsl.exe` failures into an empty list.

Current behavior:

```csharp
public async Task<IReadOnlyList<WslDistroInfo>> ListDistrosAsync(CancellationToken cancellationToken = default)
{
    var result = await RunAsync(["--list", "--verbose"], cancellationToken);
    return result.Success ? ParseDistroList(result.StandardOutput) : [];
}
```

`RunProcessAsync()` returns `new WslCommandResult(-1, ..., "wsl.exe timed out")` on timeout. `ListDistrosAsync()` then turns that into `[]`, so the new startup code sees an empty distro list and returns as if the distro is absent:

```csharp
if (!distros.Any(d => d.Name.Equals(distroName, StringComparison.OrdinalIgnoreCase)))
{
    Logger.Warn($"[WslKeepAlive] Configured local-gateway distro '{distroName}' not present; skipping keepalive.");
    return;
}
```

That means the retry/backoff may never run in the exact cold-logon timeout case this PR is meant to fix.

### Suggested minimal shape

In `TryEnsureLocalGatewayKeepAliveAsync()`, call `RunAsync()` directly for the probe so you can distinguish:

1. `wsl --list` succeeded and the distro is absent → skip.
2. `wsl --list` succeeded and the distro exists → ensure keepalive.
3. `wsl --list` timed out or failed → retry, and after retries defensively spawn.

Something like this:

```csharp
for (var attempt = 0; attempt < retryDelays.Length; attempt++)
{
    if (retryDelays[attempt] > TimeSpan.Zero)
        await Task.Delay(retryDelays[attempt]);

    using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(5));
    WslCommandResult result;
    try
    {
        result = await runner.RunAsync(["--list", "--verbose"], cts.Token);
    }
    catch (OperationCanceledException)
    {
        Logger.Warn($"[WslKeepAlive] wsl --list probe cancelled/timed out (attempt {attempt + 1}/{retryDelays.Length}); will retry.");
        continue;
    }

    if (!result.Success)
    {
        Logger.Warn($"[WslKeepAlive] wsl --list failed (exit {result.ExitCode}, attempt {attempt + 1}/{retryDelays.Length}): {result.StandardError}");
        continue;
    }

    var distros = WslExeCommandRunner.ParseDistroList(result.StandardOutput);
    if (!distros.Any(d => d.Name.Equals(distroName, StringComparison.OrdinalIgnoreCase)))
    {
        Logger.Warn($"[WslKeepAlive] Configured local-gateway distro '{distroName}' not present; skipping keepalive.");
        return;
    }

    WslDistroKeepAlive.EnsureStarted(distroName, new AppLogger());
    return;
}

Logger.Warn("[WslKeepAlive] wsl --list never succeeded; spawning keepalive defensively.");
WslDistroKeepAlive.EnsureStarted(distroName, new AppLogger());
```

The key point: only treat “distro not present” as authoritative when `wsl --list --verbose` itself succeeded. Any timeout/failure should stay retryable.

### Test I’d like to see

Please add coverage for this exact regression. The test does not need to spawn real WSL. Ideally extract the keepalive arming/probe loop behind a small seam, or add a focused helper that takes a probe function and an ensure-started callback. The test should prove:

- first probe returns `ExitCode = -1` / failure;
- second probe returns success with `OpenClawGateway` in stdout;
- `EnsureStarted` is called once;
- the code does **not** return early as “distro missing” on the failed first probe.

A second useful test: all probes fail → defensive spawn is attempted once.

### Two small hardening items while you are in there

#### 1. Avoid bare `wsl.exe` in `CreateProcessW`

The P/Invoke path currently uses `lpApplicationName: null` and a command line beginning with `wsl.exe`. With `CreateProcessW`, that can use Windows search-order resolution. Since this process is now intentionally detached/breakaway, please pass the explicit System32 path as `lpApplicationName`:

```csharp
var wslPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.System), "wsl.exe");
```

Then call `CreateProcessW(lpApplicationName: wslPath, ...)`.

You can still keep `wsl.exe` as argv[0] in the command line if you want, but the executable should be pinned.

#### 2. Escape or validate `distroName` for the manual command line

This line quotes but does not escape the value:

```csharp
return $"wsl.exe -d \"{distroName}\" -u openclaw -- sleep 2147483647";
```

Either reject unexpected distro names with a conservative allowlist such as `^[A-Za-z0-9._-]+$`, or add a proper Windows command-line quoting helper for the manual `CreateProcessW` path. Since OpenClaw controls the expected distro name, the allowlist is probably simplest and safest.

### Re-validation

After the patch, please rerun:

```powershell
./build.ps1
dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore
dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore
```

Once the retry path distinguishes failed probes from a genuinely missing distro, I think this PR is back on track.

shanselman:
﻿🤖 **Addendum: full issue checklist**

Adding the complete checklist so nothing is lost. I would split these into **must fix before merge**, **strong hardening**, and **follow-up acceptable**.

## Must fix before merge

### 1. Failed `wsl --list` probes are still treated as “distro missing”

This is the main blocker.

`ListDistrosAsync()` hides failure details:

```csharp
var result = await RunAsync(["--list", "--verbose"], cancellationToken);
return result.Success ? ParseDistroList(result.StandardOutput) : [];
```

So timeout/failure becomes `[]`. The startup keepalive code then treats `[]` as authoritative “distro not present” and returns without retrying or defensively spawning.

Fix shape:

- use `runner.RunAsync(["--list", "--verbose"], cts.Token)` directly in `TryEnsureLocalGatewayKeepAliveAsync()`;
- only parse distro output if `result.Success`;
- only return “distro missing” if the command succeeded and the distro is absent;
- on timeout/failure, retry;
- after all retries fail, call `WslDistroKeepAlive.EnsureStarted(...)` defensively.

Tests to add:

- first probe fails/times out, second succeeds with `OpenClawGateway` → `EnsureStarted` is called once;
- all probes fail → defensive `EnsureStarted` is called once;
- successful probe with no matching distro → skip without spawning.

This likely needs a small seam/helper so the retry loop can be unit tested without real WSL.

### 2. Pin the executable path for `CreateProcessW`

The new P/Invoke path currently starts from a command line beginning with bare `wsl.exe` and passes `lpApplicationName: null`. With `CreateProcessW`, that can invoke Windows search-order behavior. Since this child is intentionally detached/breakaway, do not rely on ambient search order.

Fix shape:

```csharp
var wslPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.System), "wsl.exe");
```

Pass that as `lpApplicationName` to `CreateProcessW`. I would also use the same explicit path for the `Process.Start` fallback for consistency.

### 3. Validate or correctly quote `distroName` for the manual command line

`BuildCommandLine()` currently does this:

```csharp
return $"wsl.exe -d \"{distroName}\" -u openclaw -- sleep 2147483647";
```

That quotes the value but does not escape embedded quotes/backslashes. The expected OpenClaw distro names are simple, so the safest fix is probably to validate with a conservative allowlist before building the manual command line, e.g. alphanumerics plus `.` `_` `-`.

If you want to support broader WSL distro names, add a real Windows command-line argument quoting helper and tests for spaces, quotes, and trailing backslashes. For this product path, I would prefer the allowlist and fail/log clearly.

## Strong hardening before merge if small

### 4. Delete malformed marker files during explicit stop/cleanup

`Stop()` currently returns if `TryReadMarker()` fails. That is safe from a wrong-process-kill perspective, but it leaves corrupt/stale marker files behind during mode switch, repair/remove, or uninstall cleanup.

Fix shape:

- if the marker file exists but cannot be parsed, log and delete the marker best-effort;
- still do **not** kill any PID unless identity matches.

### 5. Log breakaway fallback paths

If `CREATE_BREAKAWAY_FROM_JOB` fails with `ERROR_ACCESS_DENIED`, the code retries without breakaway. That may mean the keepalive is armed only while the tray stays alive, depending on the package/job behavior.

Please log:

- breakaway denied → retrying without breakaway;
- non-breakaway CreateProcess failed → falling back to Process.Start;
- final spawn failure.

That will make field diagnosis much easier if a packaged build still does not survive tray exit.

### 6. Add manual packaged/MSIX validation evidence

The code now attempts the right mechanism, but the PR should still record the manual validation because this is the load-bearing behavior.

Suggested evidence in the PR description or a comment:

```powershell
# packaged/MSIX install
# launch tray in local gateway mode
# capture marker PID or Get-Process wsl
# exit tray
# wait past vmIdleTimeout window
# verify same keepalive PID still exists and gateway remains reachable
```

Unit tests cannot prove this part.

## Follow-up acceptable, but should be tracked

### 7. Stop keepalive on live local→remote mode switch

The launch-time cleanup handles the next start, but if the user switches away from local gateway while the tray remains running, the keepalive can continue pinning WSL.

Follow-up shape:

- hook the existing settings-save/settings-changed path;
- detect transition from local gateway URL to non-local;
- call `WslDistroKeepAlive.Stop(distroName, logger)`.

### 8. Re-arm keepalive on Windows resume

`OnLaunched` covers logon/restart. Sleep/modern standby/resume can also tear down WSL without relaunching the tray.

Follow-up shape:

- subscribe to a power resume event in the tray lifecycle;
- fire-and-forget the same keepalive arming path on resume.

### 9. Cross-process spawn/write race can still orphan one keepalive

The new marker-write transaction fixes “spawn succeeded, marker write failed.” It does not fully protect against two independent processes both doing:

1. no valid marker;
2. spawn;
3. both marker writes succeed, with the second overwriting the first.

The loser process is then untracked. The single-instance app guard probably makes this unlikely, so I would not block on it. If you want defense-in-depth, use a named mutex per distro around adopt/spawn/write.

### 10. `CaptureIdentity()` best-effort fallback is acceptable but weak

If the process exits between `CreateProcessW` and `GetProcessById`, the marker uses empty name + `DateTime.UtcNow`. That should self-heal on the next adoption attempt, but it weakens the identity check for that marker.

Not blocking. If you want to tighten it, return start time from the original process handle via `GetProcessTimes` before closing handles.

## Re-validation checklist

After the next patch, please run:

```powershell
./build.ps1
dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore
dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore
```

The big thing is preserving the good direction of the PR while making the timeout/failure path actually distinguishable from a genuinely missing distro.

indierawk2k2:
Addressed the latest review in `8b63efd`.

### Must-fix items

1. **Failed `wsl --list` probes no longer look like “distro missing”.**
   - Replaced startup use of `ListDistrosAsync()` with a testable `WslKeepAliveStartupArmer.ArmAsync` helper.
   - It calls `RunAsync(["--list", "--verbose"])` directly and only treats a missing distro as authoritative when the command itself succeeds.
   - Failed/time-out probes retry; after all retries fail, it defensively calls `EnsureStarted`.

2. **Pinned `CreateProcessW` executable.**
   - `DefaultWslKeepAliveProcessHost` now resolves `%SystemRoot%\System32\wsl.exe` via `Environment.SpecialFolder.System` and passes it as `lpApplicationName`.
   - The `Process.Start` fallback uses the same explicit path.

3. **Conservative distro-name validation.**
   - The manual CreateProcess path now validates distro names with `^[A-Za-z0-9._-]+$` before building the command line.

### Hardening items

4. **Malformed marker cleanup.**
   - `Stop` and `StopAll` now delete malformed marker files but still never kill a PID unless marker identity matches.

5. **Breakaway fallback logging.**
   - Logs breakaway denied, non-breakaway CreateProcess failure, and fallback to `Process.Start`.

6. **Hanselman follow-up review.**
   - Ran the requested dual-model review again after the patch.
   - No CRITICAL/HIGH issues found.
   - Applied the one clean robustness improvement from Opus: `IWslKeepAliveProcessHost.Spawn` now returns `KeepAliveSpawnResult(Pid, Identity)` directly, removing the temporal coupling around `GetCurrentPidForLastSpawn`.
   - Triaged remaining non-blocking findings:
     - Allowlist may reject exotic future/custom distro names — intentional per your requested product-scoped shape; OpenClaw production names are simple.
     - Cross-process duplicate spawn race — already mitigated by tray single-instance mutex; left as follow-up/defense-in-depth.
     - Startup armer cancellation — follow-up together with resume/live-mode-switch handling.

### New/updated tests

Added tests for:
- first probe fails, second succeeds -> ensure once
- all probes fail -> defensive ensure once
- successful probe with missing distro -> skip
- malformed marker Stop cleanup

### Re-validation

- `./build.ps1` — clean
- Shared.Tests `--no-restore` — 1776 passed / 28 skipped / 0 failed
- Tray.Tests `--no-restore` — 1105 passed / 0 failed

indierawk2k2:
Addressed the validation ghost-window issue in commit 2b940fd. Tray tests now install a test-only cleanup hook that hides/closes blank explorer-owned Windows App SDK ApplicationFrameWindow shells after each test and at testhost shutdown.\n\nValidation on Windows:\n- .\\build.ps1 ✅\n- dotnet test .\\tests\\OpenClaw.Shared.Tests\\OpenClaw.Shared.Tests.csproj --no-restore ✅ 1776 passed / 28 skipped\n- dotnet test .\\tests\\OpenClaw.Tray.Tests\\OpenClaw.Tray.Tests.csproj --no-restore ✅ 1105 passed\n- post-Tray validation blank explorer ApplicationFrameWindow count: 0 ✅

indierawk2k2:
Follow-up for the workstation ghost-window report: I cleaned the visible blank App SDK shell frames on the machine, then hardened the Tray test cleanup in 436ec0f so it runs before/after each test, on a 1s watchdog timer while testhost is alive, and repeatedly during testhost shutdown to catch delayed explorer/ApplicationFrameHost frames.\n\nValidation after the hardening:\n- .\\build.ps1 ✅\n- dotnet test .\\tests\\OpenClaw.Shared.Tests\\OpenClaw.Shared.Tests.csproj --no-restore ✅ 1776 passed / 28 skipped\n- dotnet test .\\tests\\OpenClaw.Tray.Tests\\OpenClaw.Tray.Tests.csproj --no-restore ✅ 1105 passed\n- post-Tray validation blank ApplicationFrameWindow count after delay: 0 ✅

indierawk2k2:
Follow-up on the screenshot: the remaining field of windows was not the earlier blank Windows App SDK `ApplicationFrameWindow` issue. The visible frames were Windows Terminal top-level windows (`WindowsTerminal.exe` / `CASCADIA_HOSTING_WINDOW_CLASS`) titled `Terminal`.

I cleaned the workstation and added a test-only guard in `33292ca` that snapshots existing generic Terminal frames at testhost startup and closes only newly-created ones during/after Tray validation, alongside the existing blank App SDK frame cleanup.

Validation after rebasing on the latest PR branch:
- `./build.ps1` passed
- Shared tests: `1776 passed / 28 skipped`
- Tray tests: `1105 passed`
- Post-Tray ghost probe after delay: Terminal ghosts `0`, blank `ApplicationFrameWindow` ghosts `0`

shanselman:
﻿Thanks for sticking with this one, @indierawk2k2. I think this is the right app-scoped fix for a real WSL2 idle-shutdown problem: keep the OpenClaw gateway distro alive with a detached `wsl.exe -d OpenClawGateway -- sleep ...` instead of asking users to change global `.wslconfig` behavior.

What made me comfortable landing it:

- startup arming now treats a cold-logon `wsl --list --verbose` timeout/failure as retryable instead of assuming the distro is absent;
- process creation uses the explicit System32 `wsl.exe` application path;
- distro names are constrained before they reach process creation;
- stale/corrupt keepalive marker files are cleaned up;
- the tests cover marker adoption, PID reuse, corrupt markers, startup retry, and fallback paths;
- I added a final maintainer patch so the test ghost-window cleanup only targets newly-created blank WinAppSDK shell frames, not unrelated desktop/Terminal windows.

Non-blocking follow-ups I still want after this lands: re-arm after Windows resume, make live local/remote mode switching explicitly tear down any old keepalive, and keep documenting/proving packaged/MSIX behavior as we exercise this in release builds.

PR #461, opened 2026-05-19
Fix node capability binding diagnostics

ranjeshj:
## Summary
- Lazily creates NodeService during node ClientCreated so capabilities can bind before connect
- Logs and records diagnostics when settings, NodeService, or command registration are unavailable instead of silently no-oping
- Keeps the tray from crashing if capability binding fails

## Validation
- .\\build.ps1
- dotnet test .\\tests\\OpenClaw.Shared.Tests\\OpenClaw.Shared.Tests.csproj --no-restore (1782 passed, 28 skipped)
- dotnet test .\\tests\\OpenClaw.Tray.Tests\\OpenClaw.Tray.Tests.csproj --no-restore (1112 passed)

### DIFF
--- a/src/OpenClaw.Connection/NodeConnector.cs
+++ b/src/OpenClaw.Connection/NodeConnector.cs
@@ -79,6 +79,7 @@ public sealed class NodeConnector : INodeConnector
         catch (Exception ex)
         {
             _logger.Warn($"[NodeConnector] ClientCreated handler threw: {ex.Message}");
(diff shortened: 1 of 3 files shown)


shanselman:
﻿Thanks, @ranjeshj. This is the right fix order-wise: node capability binding needs to happen before the node connect handshake serializes caps/commands, and making the `ClientCreated` bridge construct `NodeService` lazily closes the setup-time gap.

I reviewed the lifecycle path, ran local validation, and added one small regression test so a throwing `ClientCreated` bridge records a visible node diagnostic instead of failing silently. The refreshed CI is green.

PR #466, opened 2026-05-19
fix: ID-based parallel tool call tracking with truthful Interrupted state

ranjeshj:
## Problem

Tool calls in the chat window get stuck in 'running' state due to:
1. Single \ActiveToolCallId\ slot can't track parallel tool calls
2. Turn end/error events don't finalize in-progress tools
3. Legacy fallback misroutes outputs to wrong tools when IDs are involved

## Solution

- **ID-based parallel tracking**: \ActiveToolCalls\ (ImmutableDictionary) maps \	oolCallId → entryId\ for parallel-safe correlation
- **Truthful Interrupted state**: Tools that never completed show as 'Interrupted' (grey dash) instead of fake 'Success'
- **Strict ID resolution**: If a \	oolCallId\ is provided but mapping is gone, returns null (no-op) — never misroutes to the wrong tool
- **Robust event ordering**: Output mapping preserved until turn end to handle \command_output\ + \item end\ arriving in any order
- **Legacy fallback preserved**: \ActiveToolCallId\ only set for events without correlation IDs (sequential legacy path)

## UI

- Interrupted tools render with grey dash glyph in \TextFillColorTertiaryBrush\
- Aggregate status priority: Error > InProgress > Interrupted > Success
- Localized to all 5 supported locales

## Testing

- 9 new regression tests covering parallel tools, ID correlation, interrupted state, legacy fallback
- All 2902 tests pass

### DIFF
--- a/src/OpenClaw.Chat/ChatModels.cs
+++ b/src/OpenClaw.Chat/ChatModels.cs
@@ -30,7 +30,8 @@ public enum ChatToolCallStatus
 {
     InProgress,
     Success,
-    Error
+    Error,
+    Interrupted
 }
 
 public enum ChatTone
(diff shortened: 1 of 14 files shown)


shanselman:
Thanks @ranjeshj ? this PR is solving the right problem. The single active tool slot was not enough for parallel tool calls, and the new ID-based `ActiveToolCalls` map plus `Interrupted` state makes the timeline much more truthful.

I pushed one maintainer patch on top of your branch after a dual review found a real edge case:

- For shell tools, the live event sequence can be `item:start` -> `command_output:end` with real stdout/stderr -> `item:end` with empty text.
- Because this PR correctly keeps the tool-call mapping alive until turn end, that later empty `item:end` event could resolve back to the same tool entry and overwrite the real output with `""`.
- The reducer now preserves existing `ToolOutput` when a subsequent output event for the same tool call has empty text, while still allowing empty `item:end` events to mark no-output tools as `Success`.
- I added reducer coverage and a provider-level regression test for the actual `item`/`command_output` sequence.

I also fixed a local validation flake I hit while running the full tray suite: `SettingsManagerIsolationTests` and `LocalGatewayUninstallTests` both mutate `OPENCLAW_TRAY_DATA_DIR`, so they now share a non-parallel xUnit collection.

Local validation on the patched branch:

- `./build.ps1` passed
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore` passed: 1782 passed / 28 skipped
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore` passed: 1122 passed

I?m waiting for CI on the pushed commit now. If it stays green, this looks merge-ready to me.

PR #474, opened 2026-05-20
Fix WSL gateway node allowlist setup

indierawk2k2:
## Summary
- configure managed WSL gateway setup to restore only the Windows desktop-host commands that upstream previously allowed by default but now filters at runtime: system.run, system.run.prepare, system.which, browser.proxy, and screen.snapshot
- remove camera.snap, camera.clip, and screen.record from the managed allowlist so privacy-sensitive capture remains explicit opt-in
- replace the Python-based config union with a single openclaw config set call for the managed local gateway

## Validation
- .\\build.ps1
- dotnet test .\\tests\\OpenClaw.Shared.Tests\\OpenClaw.Shared.Tests.csproj --no-restore
- dotnet test .\\tests\\OpenClaw.Tray.Tests\\OpenClaw.Tray.Tests.csproj --no-restore

PR #475, opened 2026-05-20
Fix Windows node gateway metadata

indierawk2k2:
Supersedes #474 with the less intrusive upstream-aligned fix.

## Summary
- send canonical Windows `deviceFamily` metadata in the node connect payload
- normalize v3 device-auth platform metadata the same way the gateway verifier does
- document that normal first-party Windows node flow should not require a global gateway command allowlist, while `camera.snap`, `camera.clip`, and `screen.record` remain explicit opt-ins

## Validation
- `./build.ps1`

### DIFF
--- a/docs/gateway-node-integration.md
+++ b/docs/gateway-node-integration.md
@@ -28,11 +28,14 @@ The gateway has hardcoded defaults per platform (from `PLATFORM_DEFAULTS`):
 | **macOS** | canvas.*, camera.list, location.get, device.info/status, contacts.search, calendar.events, reminders.list, photos.latest, motion.*, system.run/which/notify, screen.snapshot, browser.proxy |
 | **iOS** | canvas.*, camera.list, location.get, device.info/status, contacts.*, calendar.*, reminders.*, photos.latest, motion.*, system.notify |
 | **Android** | canvas.*, camera.list, location.get, notifications.*, device.*, contacts.*, calendar.*, callLog.search, reminders.*, photos.latest, motion.*, system.notify |
-| **Windows** | **system.run, system.run.prepare, system.which, system.notify, browser.proxy** |
+| **Windows** | camera.list, location.get, device.info, device.status, system.*, browser.proxy, screen.snapshot |
 | **Linux** | system.run, system.run.prepare, system.which, system.notify, browser.proxy |
 | **Unknown** | canvas.*, camera.list, location.get, system.notify |
 
-**Windows and Linux get almost nothing by default** — only system commands. No canvas, no camera, no screen, no location. This is because Windows/Linux were originally designed as headless "node host" platforms (exec-only), not full companion apps like macOS/iOS.
+Desktop host commands such as `system.run`, `browser.proxy`, and
+`screen.snapshot` are filtered unless the node reports canonical desktop
+metadata and the command is approved through pairing/live session state or
+explicit config.
 
 ### 1.2 "Dangerous" Commands (Always Need Explicit Opt-In)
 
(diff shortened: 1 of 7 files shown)


indierawk2k2:
@shanselman replacement for #474 using the upstream-aligned approach: the Windows node now sends canonical deviceFamily metadata instead of configuring a global gateway allowCommands list. Privacy-sensitive commands remain explicit opt-ins.

shanselman:
Bot review questions from the Windows node side before we merge this:

1. Can you link the matching gateway verifier / `node-command-policy` change that makes `deviceFamily: Windows` the expected canonical Windows companion metadata?
2. Does the gateway verifier normalize `platform` and `deviceFamily` the same way this PR now signs them? This PR sends JSON `deviceFamily: Windows`, but signs the v3 payload as lowercase `windows`; I want to confirm the verifier canonicalizes before comparing.
3. Is it intentional that this changes the previously signed Windows node metadata from `windows|desktop` to `windows|windows`? If so, is there a minimum gateway version this node change depends on?
4. Should `BuildConnectPayloadV3_NormalizesPlatformMetadataForGatewayAuth` be a normal `[Fact]` instead of `[IntegrationFact]` so the signature contract is covered by default test runs?
5. For existing paired nodes, do users need to reject/re-pair to pick up the updated command/default-policy behavior, or does the gateway refresh metadata on reconnect?

Local validation on the merged PR tree was clean: `build.ps1`, Shared tests, and Tray tests all passed. The only blocker from my review is confirming the gateway-side contract above.

indierawk2k2:
Thanks — I chased each gateway-side contract point and pushed the test coverage update in `20f30b5`.

1. **Gateway contract for `deviceFamily: Windows`:** the matching upstream change is openclaw/openclaw#81880 (`Require canonical node platform IDs`). The relevant policy is in [`src/gateway/node-command-policy.ts`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/gateway/node-command-policy.ts): it normalizes `platform`/`deviceFamily`, exact-matches canonical platform IDs, and for Windows requires the normalized family to be `windows`. The first-party upstream node-host now maps Node `win32` to `platform: "windows"` and `deviceFamily: "Windows"` in [`src/node-host/runner.ts`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/node-host/runner.ts), with test coverage in [`src/node-host/runner.test.ts`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/node-host/runner.test.ts).

2. **Verifier normalization vs signed payload:** yes, the gateway verifier canonicalizes before comparing. [`buildDeviceAuthPayloadV3`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/gateway/device-auth.ts) runs `platform` and `deviceFamily` through [`normalizeDeviceMetadataForAuth`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/gateway/device-metadata-normalization.ts), which trims and ASCII-lowercases. The handshake verifier rebuilds that v3 payload from `connectParams.client.platform` and `connectParams.client.deviceFamily` in [`handshake-auth-helpers.ts`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/gateway/server/ws-connection/handshake-auth-helpers.ts). So JSON `deviceFamily: "Windows"` is expected to verify against signed metadata `windows|windows`.

3. **`windows|desktop` -> `windows|windows`:** intentional. `desktop` is not the canonical Windows device-family token the new gateway policy checks; normalized `Windows` is. This behavior is only meaningful with gateways that include #81880 or newer. The wire addition remains compatible with the existing connect schema because `client.deviceFamily` is optional; older gateways should ignore it or continue through the existing v2-signature fallback if they cannot verify v3.

4. **Default test coverage:** agreed. I changed `BuildConnectPayloadV3_NormalizesPlatformMetadataForGatewayAuth` from `[IntegrationFact]` to `[Fact]` so the signature normalization contract runs in normal Shared test runs.

5. **Existing paired nodes:** reconnect should be enough when the prior pairing record did not pin a conflicting device family, because the gateway records live node session metadata from each connect and this PR does not change the declared command list. If an existing paired record is already pinned to a different family such as `desktop`, the gateway treats that as a metadata-upgrade path in [`message-handler.ts`](https://github.com/openclaw/openclaw/blob/9eee202a694b84f25dd818501255a680d2ddf34b/src/gateway/server/ws-connection/message-handler.ts), so the user may need to approve/re-pair once. That is the conservative security behavior I would document as the caveat.

Validation after the `[Fact]` change:
- `./build.ps1`

PR #479, opened 2026-05-20
feat: tool metadata cache, tool call toggle, stale-closure fix, and perf optimizations

christineyan4:
## Summary

Improves the chat experience with tool call caching, a toggle UI, and critical bug fixes.

### Tool Metadata Cache
- Persists tool call metadata (name, label, timestamp) to disk so flattened tool outputs from gateway history replay can be recovered with proper tool names/icons
- Sequential order matching between cache and history entries
- Session limits (20 sessions, 500 entries per session)
- Debounced saves (500ms quiescence timer) to prevent concurrent file writes
- Atomic file writes (write .tmp then File.Move) to prevent partial JSON on crash

### Tool Call Toggle
- Wrench icon in the toolbar to show/hide tool calls in chat
- 0.55 opacity when hidden, full opacity when shown
- Tooltips on all toolbar buttons for accessibility
- Collapses all expanded tool chips when hiding

### Stale-Closure Fix (Critical)
- Fixed a bug where UseState closures in ChatExplorationState.Changed handlers captured the initial value (0), causing Set(1) to be a no-op after the first toggle
- Affected all 4 components: Timeline, Composer, ChatRoot, ExplorationsPanel
- Fixed with UseRef-based counter pattern

### Performance Optimizations
- Pre-clear native StackPanel before render so SyncChildren only does inserts
- SyncChildren fast path for empty panels (skip RemoveFromParent scans)
- ChatRoot reads explorationRevRef.Current inside dispatcher lambda

### Tests
- 19 new tests (1161 total): ToolMetaCacheTests (sequential matching, exhaustion, guards), classifier edge cases
- All existing tests pass

### Other
- ChatPage NavigationCacheMode=Enabled, suppressAutoDispose on FunctionalHostControl
- Improved ClassifyFlattenedToolOutput heuristics (bash, view, grep, git, edit, glob detection)

### DIFF
--- a/src/OpenClaw.Tray.WinUI/Chat/Explorations/ChatExplorationState.cs
+++ b/src/OpenClaw.Tray.WinUI/Chat/Explorations/ChatExplorationState.cs
@@ -340,6 +340,20 @@ public static class ChatExplorationState
         set { if (_showToolCalls != value) { _showToolCalls = value; RaiseChanged(); } }
     }
 
+    /// <summary>
+    /// Monotonic counter incremented when all tool chip expanded states
+    /// should be reset. The timeline checks this and clears its local
+    /// <c>expandedToolChips</c> set when the value changes.
+    /// </summary>
+    public static int CollapseToolChipsVersion { get; internal set; }
+
+    /// <summary>Signal all tool chip expanded states to collapse.</summary>
+    public static void CollapseAllToolChips()
+    {
+        CollapseToolChipsVersion++;
+        RaiseChanged();
+    }
+
     public static double BubbleMaxWidth
     {
         get => _bubbleMaxWidth;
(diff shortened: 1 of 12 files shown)


shanselman:
@christineyan4 I found and pushed a fix for the tool metadata cache race I was worried about. The new commit serializes debounced saves, uses unique temp files for atomic writes, prevents stale timer callbacks from overwriting newer snapshots, flushes on provider disposal, and adds a concurrent-cache test that verifies complete valid JSON with no leftover temp files.\n\nLocal validation on the pushed branch: ./build.ps1, dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore, and dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore all passed.

PR #521, opened 2026-05-23
refactor(tray): native back navigation for cross-page back

bkudiess:
## Summary

Reworked per review feedback to **use the native `Frame` / `NavigationView` back navigation instead of custom back affordances.**

Cross-page back is now handled by a single **native back button in the Hub title bar**, driven by the `NavigationView` `ContentFrame`'s real back stack — no bespoke origin tracking or per-page back links.

> Supersedes this PR's original approach (custom "consistent back-links across cross-page navigations"). The earlier commits that added those links are now removed by the final commit in favor of native navigation.

## What changed

**Added**
- Native back button in the Hub title bar wired to `ContentFrame.GoBack()`.
  - Enabled/disabled by `CanGoBack` (mirrors `NavigationView`'s native greyed-at-root behavior).
  - **Alt+Left** keyboard accelerator.
  - Left rail re-syncs its selected item on Back/Forward.
- `NavBackButton` strings across all five locales (en, fr, nl, zh-cn, zh-tw).

**Removed**
- All custom origin-tracking infrastructure: `NavOriginLabels` helper, the 2-arg `IAppCommands.Navigate`/`NavigateTo` overloads, `LastNavigationOrigin` + pending-origin state on `HubWindow`, and the related `LocalizationHelper` additions.
- The per-page **"Back to {origin}"** links (Config / Connection / Debug / Voice) and **"Back to Connection"** links (Sessions / Instances / Permissions), with their handlers and visibility logic.
- Reverted all 2-arg navigation call sites back to single-arg `Navigate` across 11 pages.
- Obsolete resw keys for the removed links.

**Intentionally kept**
- In-page form/detail **close** buttons — ConnectionPage "Add a gateway" (also the form's Cancel) and DebugPage log-detail "Back to Diagnostics". These are view-state toggles **within** a page, not `Frame` navigations, so the native back button does not own them.

## Net effect

`32 files changed`, ~**307 lines net removed** (171 insertions / 478 deletions), including deletion of `Helpers/NavOriginLabels.cs`.

## Validation

- ✅ `build.ps1` — all projects (ARM64, .NET 10)
- ✅ Shared tests — **2045 passed**, 0 failed
- ✅ Tray tests — **934 passed**, 0 failed
- ✅ Manual smoke: rail navigation + native back button + Alt+Left re-highlight the correct rail item; in-page form/detail close buttons unchanged.

Closes the review feedback to rely on native back navigation.


ranjeshj:
We could probably use the NavigationView's built-in back stack instead of tracking where each page navigation came from and manually navigating back to that origin. That would keep the behavior tied to the actual navigation history rather than a separate origin tag.

We can also use the back button affordance provided by NavigationView instead of adding per-page back-link UI.

ranjeshj:
Closed as superseded by #780, which includes this change plus conflict-resolution, validation, localization, and review fixes.

PR #534, opened 2026-05-25
Add XAML localization resource regression coverage

shanselman:
## Summary
- Add static XAML localization validation for unsupported resource properties on x:Uid elements.
- Catch the AgentEventsPage TextBlock/Content crash class by failing if a TextBlock x:Uid is paired with a .Content resource.
- Include the one-line AgentEventsPage XAML fix needed for the new master-based regression test to pass.

## Validation
- `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --filter FullyQualifiedName~LocalizationValidationTests.XamlControlsWithXUid_DoNotUseUnsupportedLocalizationProperties`
- `.\build.ps1`

Base: `master`.


ranjeshj:
Thanks for putting this together and for adding coverage for the XAML localization crash class.

This is now already fixed on `main`: `AgentEventsPage.xaml` uses `x:Uid="AgentEventsPage_ClearButton"` with `AgentEventsPage_ClearButton.Text` resources, so the visible Clear label remains localized without hitting the stale `ClearButton.Content` path. This branch also conflicts with `main` and would remove that localization, so I’m closing it as superseded.

PR #562, opened 2026-05-27
feat: add sanitized diagnostics log bundle

christineyan4:
## Summary
- Add a sanitized diagnostics bundle export flow for the WinUI tray app, including generated debug summary, connection event timeline, tray log tails, structured diagnostics JSONL, crash log, and setup log tails.
- Shift the privacy boundary toward source-side sanitization before logs/diagnostics are written or recorded, with `DiagnosticsExportRedactor` kept as defense-in-depth for older/raw export text.
- Refactor export redaction away from a broad regex-only approach: parser-style handling now covers private keys, signed handshakes, DPAPI blobs, agent session keys, command-line secret options, sensitive key/value pairs, GUID-like IDs, and common URL/path/token cleanup.
- Improve diagnostics preview/copy readability by decoding common JSON escape sequences like `\u0022` and escaped CRLFs before final redaction, while keeping plain Windows paths stable and guarding control/backslash unicode escapes.
- Add fail-closed sanitizer-timeout handling for log-tail section sanitization and final full-bundle sanitization so a timeout produces `[REDACTED_SANITIZER_TIMEOUT]` instead of aborting export.
- Preserve the preview-before-copy/save full diagnostics flow and native Win32 Save dialog for the self-hosted WinUI app.
- Add regression coverage for redaction safety, readability, timeout behavior, split-line/nested secrets, missing files, and diagnostics page copy/preview contracts.

## Validation
- `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore --filter FullyQualifiedName~DiagnosticsExportRedactorTests`
- `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore --filter "FullyQualifiedName~DiagnosticsBundleBuilderTests|FullyQualifiedName~DiagnosticsPageContractTests"`
- `.\build.ps1`
- `dotnet test .\tests\OpenClaw.Connection.Tests\OpenClaw.Connection.Tests.csproj --no-restore`
- `git diff --check`
- Hanselman-style adversarial review completed with Opus + Codex; follow-up fixes added for nested/double-escaped JSON secrets, conservative escape decoding for Windows paths, unicode control/backslash guards, deterministic escaped-CRLF decoding, and additional timeout/readability regression tests.

### DIFF
--- a/src/OpenClaw.Connection/ConnectionDiagnostics.cs
+++ b/src/OpenClaw.Connection/ConnectionDiagnostics.cs
@@ -27,7 +27,11 @@ public sealed class ConnectionDiagnostics
 
     public void Record(string category, string message, string? detail = null)
     {
-        var evt = new ConnectionDiagnosticEvent(_clock.UtcNow, category, message, detail);
+        var evt = new ConnectionDiagnosticEvent(
+            _clock.UtcNow,
+            TokenSanitizer.SanitizeLogMessage(category),
+            TokenSanitizer.SanitizeLogMessage(message),
+            detail is null ? null : TokenSanitizer.SanitizeLogMessage(detail));
         lock (_lock)
         {
             _buffer[_head] = evt;
--- a/src/OpenClaw.SetupEngine/SetupLogger.cs
+++ b/src/OpenClaw.SetupEngine/SetupLogger.cs
@@ -1,6 +1,9 @@
 using System.Collections.Concurrent;
+using System.Text.Encodings.Web;
 using System.Text.Json;
+using System.Text.Json.Nodes;
 using System.Text.RegularExpressions;
+using OpenClaw.Shared;
 
 namespace OpenClaw.SetupEngine;
 
@@ -84,7 +87,7 @@ public sealed partial class SetupLogger : IDisposable
     {
         if (level < _minLevel) return;
 
-        var sanitizedMessage = Sanitize(message);
+        var sanitizedMessage = NormalizeLogString(Sanitize(message));
         var entry = new LogEntry(DateTimeOffset.UtcNow, _runId, level, sanitizedMessage, SanitizeData(data));
         _recentEntries.Enqueue(entry);
         while (_recentEntries.Count > MaxRecentEntries)
@@ -125,6 +128,10 @@ public sealed partial class SetupLogger : IDisposable
     internal static string Sanitize(string input)
     {
         if (string.IsNullOrEmpty(input)) return input;
+        input = TokenSanitizer.SanitizeLogMessage(input);
+        if (input == TokenSanitizer.SanitizerTimeoutSentinel)
+            return input;
+
         input = PrivateKeyPattern().Replace(input, "[REDACTED-PRIVATE-KEY]");
         input = BearerPattern().Replace(input, "$1[REDACTED]");
         input = JwtPattern().Replace(input, "[REDACTED-JWT]");
@@ -140,20 +147,64 @@ public sealed partial class SetupLogger : IDisposable
             return null;
 
         if (data is string value)
-            return Sanitize(value);
+            return NormalizeLogString(Sanitize(value));
 
         try
         {
             var json = JsonSerializer.Serialize(data, _jsonOptions);
             var sanitized = Sanitize(json);
-            using var doc = JsonDocument.Parse(sanitized);
-            return doc.RootElement.Clone();
+            var node = JsonNode.Parse(sanitized);
+            return NormalizeJsonNode(node);
         }
         catch (Exception ex) when (ex is JsonException or NotSupportedException or ArgumentException)
         {
-            return Sanitize(data.ToString() ?? string.Empty);
+            return NormalizeLogString(Sanitize(data.ToString() ?? string.Empty));
         }
     }
+
+    private static JsonNode? NormalizeJsonNode(JsonNode? node)
+    {
+        switch (node)
+        {
+            case JsonObject obj:
+                var normalizedObject = new JsonObject();
+                foreach (var property in obj)
+                    normalizedObject[property.Key] = NormalizeJsonNode(property.Value);
+                return normalizedObject;
+
+            case JsonArray array:
+                var normalizedArray = new JsonArray();
+                for (var i = 0; i < array.Count; i++)
+                    normalizedArray.Add(NormalizeJsonNode(array[i]));
+                return normalizedArray;
+
+            case JsonValue value when value.TryGetValue<string>(out var text):
+                var normalized = NormalizeLogString(text);
+                var trimmed = normalized.TrimStart();
+                if (trimmed.StartsWith('{') || trimmed.StartsWith('['))
+                {
+                    try
+                    {
+                        return NormalizeJsonNode(JsonNode.Parse(normalized));
+                    }
+                    catch (JsonException)
+                    {
+                        // Keep malformed JSON-shaped text as a flattened string.
+                    }
+                }
+
+                return JsonValue.Create(normalized);
+
+            default:
+                return node?.DeepClone();
+        }
+    }
+
+    private static string NormalizeLogString(string value) =>
+        value
+            .Replace("\r\n", " ", StringComparison.Ordinal)
+            .Replace('\r', ' ')
+            .Replace('\n', ' ');
 
     private static string Truncate(string input, int max = 4096)
         => input.Length <= max ? input : input[..max] + $"... [truncated {input.Length - max} chars]";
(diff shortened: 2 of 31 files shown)


christineyan4:
Real behavior proof for diagnostics bundle privacy boundary:
 
 - “Create diagnostics bundle” opens a preview before log-tail diagnostics can be copied or saved.
 - Preview shows sanitized diagnostics/log content with useful context preserved.
 - “Save to file” opens the native Windows Save dialog.
 - Direct “Copy summary debug bundle” is summary-only and explicitly excludes log tails; log-tail diagnostics 
require the preview flow.


01-preview-dialog
02-native-save-dialog
03-summary-copy-card

shanselman:
This direction makes sense for a shareable diagnostics bundle, but the redactor is carrying a lot of regex responsibility in one place. I would not block solely because it uses regex — conservative over-redaction is probably right here — but I think it needs a maintainability/perf pass before merge: group the rules by threat class, consider source-generated regexes or clearly named rule objects, and add a worst-case/perf regression test so a long log line or malformed URL/token cannot cause pathological backtracking. The current test coverage for common secret shapes is good; my concern is mostly long-term maintainability and worst-case behavior.

shanselman:
I took a closer pass on the redaction implementation. The feature is valuable and regex is not inherently the wrong tool for free-form diagnostics, but I think this needs a small reliability/perf cleanup before merge.

Concrete concerns:

1. `RegexMatchTimeoutException` can abort the whole export. `DiagnosticsLogTailReader.BuildSection()` calls `DiagnosticsExportRedactor.Sanitize(...)`, but its catch filter only handles IO/unauthorized/not-supported exceptions. If a regex hits the 100ms timeout on a long/malformed log line, the timeout escapes and the diagnostics export fails. Please catch `RegexMatchTimeoutException` there and around the final full-bundle `Sanitize(builder.ToString())` pass in `DiagnosticsBundleBuilder.Build()` so the bundle can include a sanitization-timeout sentinel instead of failing.

2. `SlackSigningSecretPattern` looks redundant/dead. It matches exactly 32 hex chars, but `HexTokenPattern` runs earlier and matches 32+ hex chars, so it should already replace every value that SlackSigningSecretPattern could see. I’d remove the Slack-specific regex and document/test that `HexTokenPattern` covers that shape.

3. `KeyValueSecretPattern` has unbounded key-prefix scans before a large alternation. With `MaxLineChars = 8000`, a long non-matching key can burn work until the timeout. Consider bounding the key prefix/suffix length, or using a small rule pipeline that parses the key and then checks sensitive-key words outside regex.

4. Please add tests for the scary cases: timeout does not abort bundle export, redaction is idempotent, JSON non-string secret values are handled or explicitly documented as a known gap, version strings/IP false positives are documented, and one max-size/perf smoke proves the redactor budget stays reasonable.

Longer-term, because this is `net10.0`, source-generated regexes (`[GeneratedRegex]`) or named redaction-rule objects would make this much easier to audit than a flat list of compiled regex fields. I would not block on “regex exists”, but I would block on timeout handling and at least one worst-case/perf regression guard.

christineyan4:
Screenshot 2026-06-11 131134
Screenshot 2026-06-11 131119

christineyan4:
@clawsweeper re-review

kmahone:
@christineyan4, take a look at Scott's feedback so that the sanitization is happening at log time instead of cleaning the log at export time.

shanselman:
Christine, I think the diagnostics bundle feature is valuable, but I do not think this PR is safe to merge in its current shape. I would recommend redesigning it around a smaller, side-effect-free export pipeline rather than continuing to patch this version.

The main concern is the privacy/performance boundary. `DiagnosticsBundleBuilder.Build()` currently calls `NormalizePersistedDiagnosticsLogs()`, which can rewrite active tray logs, diagnostics JSONL, crash logs, raw chat logs, and every setup `*.jsonl` file just because the user previews or exports a bundle. That makes a read/export action mutate diagnostic evidence, can race with the async log writers, and can become slow on machines with large or many logs.

I think the safer design is:

1. Never rewrite source logs during export. Export should tail-read bounded sections and sanitize into the bundle only. Any legacy log migration should be a separate maintenance path with explicit locking and limits.
2. Bound the bundle aggressively. Set max files, max lines, max bytes per line, max bytes per section, and max total bundle size. Tail first, then redact; do not read/sanitize entire historical logs.
3. Use one redaction path for every copy/save surface. The existing "Summary debug bundle" path still builds `Recent Tray Log` from raw lines via `CommandCenterTextHelper.BuildRecentTrayLogTail()`, bypassing the new bundle builder. All diagnostics copy/export paths should go through the same side-effect-free sanitizer.
4. Fail closed. If redaction times out or produces invalid JSONL, emit a safe sentinel record like `{"event":"redacted_unsafe_log_line"}`. Never fall back to the original line.
5. Prove reliability and speed with tests. Add regression tests for long/adversarial log lines, sanitizer timeout, large log files, many setup logs, concurrent writer/export behavior, legacy unsanitized lines, and all UI copy/export paths.

A smaller salvageable v1 would be: remove persisted-log normalization, keep only bounded tail export, sanitize every emitted line/string at export time, route all copy/save actions through that path, and include perf/timeout tests. Once that is solid, source-side sanitization can still be improved separately, but export must remain a defense-in-depth boundary rather than assuming logs are already safe.

christineyan4:
https://github.com/user-attachments/assets/6307e714-6b66-4fea-a13e-a8c10b88e0e7

Screenshot 2026-06-26 111154

christineyan4:
@clawsweeper re-review

christineyan4:
diagnostics bundle preview, with PII sanitized:
https://github.com/user-attachments/assets/a9115b93-fd25-4043-8d2b-28a9a36e9ad7

full diagnostics bundle, copied and pasted:
[OpenClaw Windows Tray Diagnostics copy-paste.txt](https://github.com/user-attachments/files/29388457/OpenClaw.Windows.Tray.Diagnostics.copy-paste.txt)

full diagnostics bundle .txt, after clicking "save to file"
[openclaw-diagnostics-20260626-120629.txt](https://github.com/user-attachments/files/29388370/openclaw-diagnostics-20260626-120629.txt)

summary bundle: 
https://github.com/user-attachments/assets/31cc0884-54ee-4f68-9924-fc90e3afcb48

christineyan4:
@clawsweeper re-review

christineyan4:
@clawsweeper re-review

christineyan4:
latest proof:
diagnostics bundle preview:
https://github.com/user-attachments/assets/7e912e67-8faa-4d69-ae6e-4200d15fe8d2

full bundle, copy/paste:
[OpenClaw Windows Tray Diagnostics pasted.txt](https://github.com/user-attachments/files/29391269/OpenClaw.Windows.Tray.Diagnostics.pasted.txt)

full diagnostics bundle .txt, after clicking "save to file":
[openclaw-diagnostics-20260626-134405.txt](https://github.com/user-attachments/files/29391380/openclaw-diagnostics-20260626-134405.txt)

summary bundle:
[OpenClaw Windows Tray Debug Bundle summary.txt](https://github.com/user-attachments/files/29391398/OpenClaw.Windows.Tray.Debug.Bundle.summary.txt)

for local validation:  build + shared tests + tray tests passed on the latest commit.

christineyan4:
Fresh current-head proof for diagnostics bundle flow at `802e44a6`:

- Preview dialog opens from **Create diagnostics bundle**.
- Full **Copy to clipboard** output was captured.
- **Save to file** output was captured.
- Copied full bundle and saved bundle are byte-identical.
- Full bundle includes bounded log-tail sections and sanitized markers.
- Checks found no raw `%USERPROFILE%` path and no raw `/home/<user>` path in copied/saved full bundle text.
- Summary-only debug bundle proof was captured separately and excludes log-tail sections.

Local validation on this head already passed: `./build.ps1`, shared tests, tray tests, and `git diff --check`.

Artifacts prepared for attachment:

1. `01-diagnostics-preview-current-head.png`
2. `02-full-bundle-copy-current-head.txt`
3. `03-saved-bundle-current-head.txt`
4. `04-summary-copy-menu-current-head.png`
5. `05-summary-copy-current-head.txt`

christineyan4:
Text artifacts for the fresh current-head proof are also available here:

https://gist.github.com/christineyan4/93b473224ce4349cbbb11087b97cc74a

Included:
- `02-full-bundle-copy-current-head.txt`
- `03-saved-bundle-current-head.txt`
- `05-summary-copy-current-head.txt`
- proof summary / validation note

The local screenshot artifacts prepared for upload are:
- `01-diagnostics-preview-current-head.png`
- `04-summary-copy-menu-current-head.png`

The text artifacts show current-head full copy/save output and summary-only content; full copy and save outputs are byte-identical, include sanitized log tails, and the summary artifact excludes log-tail sections.

christineyan4:
@clawsweeper re-review

christineyan4:
Updated proof gist now contains all five current-head artifacts:

https://gist.github.com/christineyan4/93b473224ce4349cbbb11087b97cc74a

Files:
- `01-diagnostics-preview-current-head.png.base64.txt` (preview screenshot, base64-encoded PNG)
- `02-full-bundle-copy-current-head.txt`
- `03-saved-bundle-current-head.txt`
- `04-summary-copy-menu-current-head.png.base64.txt` (summary-copy menu screenshot, base64-encoded PNG)
- `05-summary-copy-current-head.txt`

PR #575, opened 2026-05-28
Fix local dev setup: Node.js prereq, log link, WSL diagnostics

kmahone:
Fix local dev setup: Node.js prereq, log link, WSL diagnostics

Three small fixes layered on PR #573 to make local dev/run more
reliable when the WinUI tray hits setup failures:

1. build.ps1: detect missing Node.js / npm before invoking the
   tray build, and emit an actionable hint pointing at the
   nodejs.org installer. Avoids confusing C# build errors when
   the prereq is missing.

2. LogFileLauncher (SetupEngine.UI): wire up the "open log"
   hyperlink on CompletePage / ProgressPage so it actually opens
   the failure log in Explorer (via /select), and translate the
   MSIX virtualized path back to a real %LOCALAPPDATA% path so
   the file is reachable outside the packaged process.

3. SetupSteps.TryGetEnvironmentIssue (SetupEngine): match WSL
   environment diagnostics against output wsl.exe actually emits.
   The previous matcher keyed on a fabricated "Virtual Machine
   Platform" + "optional component" phrase that modern wsl.exe
   does not produce, so the VM Platform / Hyper-V branch never
   fired in practice. Replace with two checks grounded in
   observed wsl.exe output:
   - "virtualization is not enabled" (older firmware-virt-off
      message)
   - "0x80370102" (HCS_E_SERVICE_NOT_AVAILABLE -- the canonical,
      locale-independent HRESULT for a missing Virtual Machine
      Platform / Hyper-V feature)
   Tests updated to use real wsl.exe output samples.

Also: run-app-local.ps1 passes --executable "OpenClaw.Tray.WinUI.exe"
in the -UseWinApp path so the wrapper launches the WinUI tray when
running unpackaged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,7 @@ jobs:
         dotnet-version: 10.0.x
 
     - name: Cache NuGet packages
+      continue-on-error: true
       uses: actions/cache@v5
       with:
         path: ~/.nuget/packages
@@ -290,6 +291,7 @@ jobs:
         dotnet-version: 10.0.x
 
     - name: Cache NuGet packages
+      continue-on-error: true
       uses: actions/cache@v5
       with:
         path: ~/.nuget/packages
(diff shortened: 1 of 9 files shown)


kmahone:
This includes fixes cherry-picked from #573.
Once #573 is merged, I will rebase this branch so they are not included in this PR.

Issue #554, opened 2026-05-27
P1: App.xaml.cs (4204 lines) and ConnectionPage.xaml.cs (2840 lines) are unmaintainable god files


QQSHI13:
## Problem
Multiple files have grown far beyond the threshold where they can be safely reviewed or unit-tested in isolation.

- `src/OpenClaw.Tray.WinUI/App.xaml.cs` — **4,204 lines**
- `src/OpenClaw.Tray.WinUI/Pages/ConnectionPage.xaml.cs` — **2,840 lines**

`App.xaml.cs` currently contains:
- Single-instance mutex logic
- Protocol/deep-link activation
- CLI uninstall handling
- Update check
- Onboarding flow orchestration
- Global hotkey setup
- Tray menu construction
- Gateway event wiring (auth failure, notifications, session commands)
- Voice service lifecycle
- Channel toggle logic

This violates the Single Responsibility Principle and makes it nearly impossible to write unit tests for any individual concern.

## Impact
- A bug fix in the tray menu can accidentally break onboarding because they share scope and fields.
- New developers cannot understand the app startup flow without reading 4,000+ lines.
- Code review diffs are too large to reason about.

## Suggested fix
Extract cohesive modules into dedicated services or partial classes:
- `AppLifecycleService` — mutex, single-instance, protocol activation, uninstall
- `AppUpdateService` — update check, skip logic
- `TrayMenuBuilder` — menu construction and snapshot handling
- `AppGatewayEventRouter` — connection status, auth failure, notifications
- `AppVoiceLifecycle` — voice service, TTS, STT setup
- `ConnectionPageViewModel` / `ConnectionPagePlan` — move remaining row-state and plan logic out of the code-behind

## Files
- `src/OpenClaw.Tray.WinUI/App.xaml.cs`
- `src/OpenClaw.Tray.WinUI/Pages/ConnectionPage.xaml.cs`

shanselman:
PR #564 landed the safe first stage for this issue: App refactor contract tests plus a small tested toast activation extraction. Keeping this open because the larger App/ConnectionPage decomposition remains follow-up work; in particular, ConnectionPage row/add/scan flows share mutable state and should get dedicated view-model or builder extraction rather than a partial-only split.

AlexAlves87:
A few more focused `App.xaml.cs` extractions are up for this — each a self-contained change with build + runtime proof, all independent of each other (separate branches off `master`, mergeable in any order):

[...]

PR #607, opened 2026-05-30
refactor: extract AppCrashLogger from App.xaml.cs

AlexAlves87:
## Summary
Extract the crash-logging sink out of `App.xaml.cs` into a small dedicated
`AppCrashLogger`, mirroring the existing `AppRunMarker`. Pure refactor, no
behavior change.

## What changed
- New `Services/AppCrashLogger.cs`: takes the crash-log path in its constructor
  and exposes `Log(string source, Exception? ex)` — the exact body of the former
  static `LogCrash` (append to `crash.log` + `Logger.Error`).
- `App.xaml.cs`: removed the static `CrashLogPath` field and the static
  `LogCrash` method; added a `_crashLogger` field. The three unhandled-exception
  handlers and the `ShowTrayMenuPopup` catch now delegate to it.

## Testing
- `dotnet build` clean (0 warnings, 0 errors).
- Tray test suite green (872 tests).

## Proof
Forced a crash at runtime (a temporary throw on a background thread, **not part
of this change**) to exercise the extracted logger through
`OnDomainUnhandledException`. `crash.log` received the expected entry:

Captura de pantalla 2026-05-31 011319

```
[2026-05-31 01:12:38.464] DomainUnhandledException
System.InvalidOperationException: ProofCrash #554
```
(screenshot below)

## Notes
Behavior is byte-for-byte identical: same timestamp format, same append mode,
same two-stage exception swallowing, same `Logger.Error` calls. No locking added
(the original had none). `OnProcessExit` / `s_runMarker` are untouched.

Addresses part of #554.

### DIFF
--- a/src/OpenClaw.Tray.WinUI/App.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/App.xaml.cs
@@ -197,7 +197,7 @@ public partial class App : Application, OpenClawTray.Services.IAppCommands
             Environment.GetEnvironmentVariable("OPENCLAW_TRAY_APPDATA_DIR")
                 ?? Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData),
             "OpenClawTray");
-    private static readonly string CrashLogPath = Path.Combine(DataPath, "crash.log");
+    private readonly AppCrashLogger _crashLogger = new(Path.Combine(DataPath, "crash.log"));
     private static readonly AppRunMarker s_runMarker = new(Path.Combine(DataPath, "run.marker"));
     private const string DisableMouseInPointerEnv = "OPENCLAW_DISABLE_MOUSE_IN_POINTER";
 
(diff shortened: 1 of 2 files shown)


AlexAlves87:
Thanks @clawsweeper for the thorough sweep — the security pass and the runtime-proof acknowledgement are much appreciated! 🙏

One small thing on the "superseded by #493" conclusion: I think there may be a mix-up there. #493 extracted `AppRunMarker` + `CliUninstallHandler` and removed the dead recording handler — it never touched the crash logger. `CrashLogPath` and the static `LogCrash` are still inline in `App.xaml.cs` on current `master` (around lines 200 and 317), which is exactly what this PR moves into `AppCrashLogger`.

So this one isn't a duplicate of #493 — they share the `App.xaml.cs` extraction *pattern*, but the code actually moved here isn't on `master` yet. The two PRs are independent slices.

@clawsweeper could you take another look when you get a chance? Happy to rebase or adjust if I'm missing something. 🙂

PR #616, opened 2026-06-01
fix(pairing): clarify reconnect and startup state

vincentkoc:
## Summary
- register setup-engine startup opt-in through a logon-triggered scheduled task, with the existing Run-key path as fallback
- carry operator/node credential source through connection snapshots so Connection UI can say when reconnect is using a paired device token
- suppress placeholder `1.0.0` app-version display for offline orphan paired-node rows, while preserving matched/online versions

## Validation
- `git diff --check`
- local `.NET` validation blocked on this mac shell: `dotnet` and `pwsh` are not installed/resolving, and `./build.ps1` is not directly executable here

## Notes
- This follows up PR #613. The merged PR used GitHub's rebase merge path because repository settings reject merge commits; it was not squashed.

### DIFF
--- a/src/OpenClaw.Connection/ConnectionStateMachine.cs
+++ b/src/OpenClaw.Connection/ConnectionStateMachine.cs
@@ -13,6 +13,8 @@ internal sealed class ConnectionStateMachine
     private RoleConnectionState _nodeState = RoleConnectionState.Idle;
     private string? _operatorError;
     private string? _nodeError;
+    private string? _operatorCredentialSource;
+    private string? _nodeCredentialSource;
     private bool _nodeEnabled;
 
     /// <summary>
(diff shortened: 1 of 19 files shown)


shanselman:
Thanks for the cleanup work here. I don't think this is safe to merge yet because the startup/scheduled-task behavior still needs author clarification and hardening before we can be confident it won't misreport or change autostart state incorrectly.

Could you please address the scheduled-task scope/enabled-state concerns directly — especially whether the task is per-user vs all-user, how the UI determines the true enabled state, and how reconnect/startup status behaves after upgrades or missing tasks? Once that's explicit in code/tests/proof, this can get another focused review.

shanselman:
Maintainer note: thanks for this — the direction is useful, but the branch is currently conflicting in `CompletePage.xaml.cs`, `ConnectionPage.xaml.cs`, and `ConnectionPageApproveCommandTests.cs`. Because this touches connection/startup state and autostart behavior, I don't want to do a quick maintainer-side conflict resolution and accidentally change behavior. Please rebase on current `main` and keep the reconnect-state changes tightly scoped; happy to re-review after that.

PR #659, opened 2026-06-03
feat(setup): inject Windows-node guidance into AGENTS.md after pairing

paulcam206:
## Summary

Adds a setup-owned Windows-node context block to the effective main/default agent workspace after onboarding. This teaches the gateway agent to route Windows work through the paired tray node instead of attempting Windows-only operations inside WSL.

The maintainer pass integrates the feature into the native onboarding wizard and hardens it for upgrades, retries, custom agent workspaces, and deterministic uninstall.

## Behavior

- Runs Windows context finalization after onboarding, including the wizard-only path.
- Resolves the effective main/default agent workspace from `openclaw agents list --json`; per-agent overrides do not rewrite global defaults.
- Feature-detects `openclaw setup --baseline` while retaining compatibility with the pinned 2026.6.11 gateway CLI.
- Injects one managed block into `AGENTS.md`, preserving unrelated content, CRLF, and existing file mode.
- Uses WSL stdin execution so `wsl.exe` cannot mutate shell variables in the script argument.
- Retries only the Windows integration step after onboarding succeeds; no repeated provider/device setup.
- Records the exact distro, user, and workspace target so uninstall removes the block from the location that was actually modified.
- Retains cleanup state on failure and rejects unsafe symlink mutation.

## Implementation notes

The file mutation remains POSIX-shell based and does not assume Node.js exists in a fresh WSL distro. Managed markers make repeated application idempotent and rollback scoped. A focused WSL argv reference documents why scripts containing shell variables must be sent through stdin.

## Validation

- AutoReview full-branch review: clean; no accepted/actionable findings.
- Windows 11 ARM64 VM: `OpenClaw.SetupEngine.Tests` 418/418 passed.
- Windows 11 ARM64 VM, fresh Git clone: Shared 2697 passed / 31 skipped; Tray 1578/1578 passed; full ARM64 build passed.
- Windows 11 ARM64 VM: local dev installer built, installed side-by-side, launched the native wizard, and reached automatic WSL gateway creation.
- The Apple-silicon Parallels VM cannot provide nested virtualization; the wizard surfaced `HCS_E_HYPERV_NOT_INSTALLED` with actionable recovery text at the expected boundary.
- [Exact-head CI](https://github.com/openclaw/openclaw-windows-node/actions/runs/28779911851): unit/integration suite, real WSL setup/connect, recovery shards, and x64/ARM64 builds all passed.

Co-authored with and based on the original contribution from @paulcam206.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,6 +77,7 @@ Start with these docs before changing connection, pairing, node, MCP, or tray UX
 - `docs/MCP_MODE.md` - local MCP server mode and the `EnableNodeMode` / `EnableMcpServer` matrix.
 - `docs/WINDOWS_NODE_TESTING.md` - Windows node capabilities, manual smokes, and gateway-dependent behavior.
 - `docs/ONBOARDING_WIZARD.md` - first-run setup flow, setup-code/bootstrap pairing, and test isolation.
+- `docs/WSL_EXE_ARGV_PITFALL.md` - wsl.exe argv variable-expansion pitfall; required reading before adding any multi-line WSL script through `RunInWslAsync`.
 
 ## Architecture Guardrails for Large Refactors
 
--- a/docs/ONBOARDING_WIZARD.md
+++ b/docs/ONBOARDING_WIZARD.md
@@ -34,7 +34,7 @@ The Capabilities page applies the selected profile to both setup config and runt
 
 ### OpenClaw onboard
 
-After node pairing, local WSL setup runs the pinned gateway CLI's non-interactive baseline initializer to seed the runtime workspace, then writes fixed Windows-node guidance into a setup-owned managed section of that workspace's `AGENTS.md`. The section is replaced idempotently between markers, preserves user-authored `AGENTS.md` content and file permissions outside those markers, and does not modify OpenClaw source files. This helps the initial companion-app OpenClaw session know to use the Windows node / `nodes` tool for Windows desktop, files, screenshots, camera, notifications, browser proxy, and Windows command tasks.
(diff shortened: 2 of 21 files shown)


paulcam206:
just as a heads up -- this change is a little wonky in terms of what we're doing... it _works_, but I wouldn't say that it's elegant.

anyways, here's a completely clean install being asked to pop a notification: 
image

shanselman:
@paulcam206 does it only do it if the message is coming in via the Windows app?

shanselman:
Automated Copilot review note from the wide PR triage pass:

I would not treat this as safe to merge at 90% confidence yet, even aside from the current conflict state. The core concern is scope: this PR injects Windows-node guidance into the gateway workspace's global AGENTS.md, so it can affect agent behavior for requests that did **not** originate from the local Windows tray chat — for example Telegram-originated work, remote/gateway sessions, or future multi-node scenarios.

Two things I think should be tightened before merge:

1. Scope the guidance to cases where the task explicitly needs this paired Windows machine. The current wording can over-bias the agent toward the local Windows node for broadly described Windows/files/browser/notification work, even when the request came from another channel or another node context.
2. Avoid putting permission-relaxing config advice in global agent guidance. The suggested openclaw config set tools.exec.ask off and openclaw config set tools.exec.security full commands are especially risky when Telegram/remote messages may influence the agent. If those defaults are needed for setup, they should be handled by explicit setup/config UX with user consent, not broad AGENTS.md instructions.

Separately, because this is post-pairing enrichment, I would prefer injection failures/timeouts/malformed existing markers to be non-terminal or self-healing where possible. Optional guidance should not block onboarding after pairing succeeds.

shanselman:
Thanks for the proof and the implementation here. I don't think this is safe to merge as-is yet because the generated `AGENTS.md` guidance is workspace-global, so it can influence sessions regardless of where the request originated — local tray, remote gateway, Telegram, or another paired node. It also currently suggests broad `tools.exec` permission changes, which is risky guidance to inject automatically.

Could you please revise this so the injected guidance is clearly scoped to tasks that intentionally need this paired Windows node, remove or soften the permissive `tools.exec` config recommendations, and ensure AGENTS enrichment is non-blocking for pairing/onboarding if it times out or fails? After that, this should get another focused review.

PR #676, opened 2026-06-03
Native chat: cross-turn overwrite + reasoning bubbles + exec.approval delivery + multi-segment dedup

RBrid:
Five related fixes for chat timeline rendering and exec-approval delivery, plus a follow-up fix for the rejected `exec.approval.resolve` response path, all validated by adversarial dual-model code review (Opus 1M + GPT-5.3-Codex) until both converged on no remaining actionable issues.

1. Cross-turn assistant overwrite (the original repro)

In a user → reply1 → tool-call → tool-output → reply2 flow, reply2 was silently overwriting reply1's text in place instead of appending a new bubble. Cause: the chat.message final for reply2 carried reconcilePrevious=true, and UpsertAssistant's fast-path merged into the most recent assistant entry unconditionally — including a previously finalised one.

Fix in ChatTimelineReducer.UpsertAssistant: narrow reconcile to only merge into entries that are still IsStreaming=true. A finalised assistant entry (IsStreaming=false) belongs to a completed turn and must not be overwritten. The byte-equal duplicate safety net remains unconditional so genuine duplicate emissions still collapse.

Defense-in-depth: extract ClearStreamingAtTurnBoundary helper and call from both AddLocalUser (production typed-input path) and ApplyUserMessage (gateway-injected path), hoisted above the nonce early-return. This handles the dropped-ChatTurnEndEvent scenario where a stale ActiveAssistantId could otherwise leak across turns.

2. Concatenated thinking prose

The gateway emits a stream:"item", kind:"reasoning", phase:"end" bracket marker between distinct thinking passes within a single turn. Map it to a new ChatReasoningEndEvent that nulls ActiveReasoningId so the next reasoning chunk starts a fresh bubble instead of appending to / replacing the previous one. Also tag delta-state ChatMessageEvents with IsStreaming=true so the reducer's new reconcile guard distinguishes them from finals.

3. Exec-approval delivery in native chat

Native WinUI chat was silently dropping exec.approval.requested / exec.approval.resolved because the gateway client only forwarded the agent stream, and the tray responder sent /approve <id> as a slash command — which deadlocks while the agent is blocked waiting for that very approval. Net effect: the Allow/Deny banner never appeared, or appeared and did nothing.

 - Gateway client: HandleEvent now translates top-level exec.approval.requested / exec.approval.resolved events into synthetic AgentEventInfo (Stream="approval", phase derived locally from the event name — payload.phase is intentionally ignored). Multi-subscriber invoke isolates per-subscriber exceptions so one throwing handler can't suppress others.
 - New RPC path: IChatGatewayBridge.ResolveExecApprovalAsync calls the exec.approval.resolve RPC. Decision is allowlisted at the send site (allow-once | allow-always | deny).
 - Tray responder: RespondToPermissionAsync swaps the slash command for the RPC. On RPC failure the pending-permission banner is preserved so the user can retry instead of silently losing the prompt.

4. Trace logging plumbing

Add IOpenClawLogger.Trace verbose channel with a default no-op so existing implementers (tests, console logger, setup logger) don't need updating. Logger.Trace in the tray is gated by OPENCLAW_TRAY_TRACE=1|true. Forward Trace through DiagnosticTeeLogger and AppLogger; SetupOpenClawLogger inherits the default no-op (no opt-in gate available in setup engine). OpenClawGatewayClient now surfaces item-event kind+phase metadata (no payload content) at trace level for shape diagnostics.

5. Multi-segment chat-bubble duplication (last bubble repeats prior segments)

When a turn produced 3+ assistant segments split by tool calls (A1 → Tool → A2 → Tool → A3), the gateway emits cumulative-for-turn chat.message frames containing every prior segment. The fast-path strip already handled this while ActiveAssistantId was live, but late frames arriving AFTER ApplyTurnEnd cleared ActiveAssistantId came through the scan-back reconcile branch, which wrote the raw cumulative text into the last bubble verbatim. Users observed A3 render correctly first, then mutate to repeat A1+A2's content.

 - ChatTimelineReducer scan-back reconcile merge now calls StripCumulativeTurnPrefix (with excludeIndex=li) so the merge target retains only its own segment tail.
 - Symmetric empty-result guard mirroring the new-bubble path's duplicate-resend skip: if a stale / duplicate cumulative frame is fully consumed by other priors, leave the streaming candidate intact instead of blanking it.
 - StripCumulativeTurnPrefix excludeIndex caller contract documented in xml-doc (self-peel hazard).

6. exec.approval.resolve response-await + rejected-resolve banner-preserve (ClawSweeper P1 follow-up, commit 77a707f7)

Round 5 of the ClawSweeper review flagged that ResolveExecApprovalAsync used SendTrackedRequestAsync, which returns once the WS frame leaves the socket. An ok:false rejection from the gateway was logged and dropped while the chat banner silently cleared, leaving the agent blocked with no retry path.

 - Register a per-requestId TaskCompletionSource in _pendingApprovalResolves and route ok:true / ok:false in HandleResponse so the caller awaits the actual gateway verdict.
 - Disconnect cleanup faults pending TCSes with OperationCanceledException (matches the _pendingWizardResponses cleanup taxonomy). Explicit contract comment notes callers must catch System.Exception (not narrow OCE) to preserve banner-preserve UX, because RunFireAndForget swallows OCE silently.
 - 15s bounded timeout (approval-resolve traverses gateway → operator → optional UI confirm → ack; longer budget than the 5s chat.send sibling). Timeout / send-fail branches call RemovePendingRequest so _pendingRequestMethods does not leak.
 - Task.WhenAny followed by an IsCompleted check rather than a returned-task identity compare, to avoid a same-tick race that would surface a real ok:true / ok:false response as a spurious TimeoutException.
 - Post-send !IsConnected re-check using a TryRemove-first idiom: if we win the race and pull the TCS out of the dict, throw InvalidOperationException immediately (no exception is ever set on the discarded TCS, so no UnobservedTaskException risk); if the cleanup loop already claimed the entry and faulted it with OCE, fall through to the WhenAny path so the cleanup-set exception is observed by the caller. Closes a narrow window where a disconnect between the up-front IsConnected guard and TCS registration would otherwise have stranded the caller for the full 15s timeout.

Coverage matrix for the rejected-resolve / banner-preserve contract

| Rejected-resolve path                                | Pinned by test                                                         | Test file                                                 |
|------------------------------------------------------|------------------------------------------------------------------------|-----------------------------------------------------------|
| Gateway returns ok:false with error.message          | ResolveExecApproval_OkFalseResponse_SurfacesAsException                | tests/OpenClaw.Shared.Tests/OpenClawGatewayClientApprovalTranslationTests.cs |
| Gateway returns ok:false with no error.message       | ResolveExecApproval_OkFalseWithoutErrorMessage_UsesFallback            | tests/OpenClaw.Shared.Tests/OpenClawGatewayClientApprovalTranslationTests.cs |
| Gateway returns ok:true (happy path completes await) | ResolveExecApproval_OkTrueResponse_CompletesCaller                     | tests/OpenClaw.Shared.Tests/OpenClawGatewayClientApprovalTranslationTests.cs |
| Disconnected at call time                            | ResolveExecApprovalAsync_ThrowsWhenDisconnected                        | tests/OpenClaw.Shared.Tests/OpenClawGatewayClientApprovalTranslationTests.cs |
| Unknown decision value                               | ResolveExecApprovalAsync_RejectsUnknownDecision                        | tests/OpenClaw.Shared.Tests/OpenClawGatewayClientApprovalTranslationTests.cs |
| Tray UX: RPC throws → banner stays visible           | RespondToPermissionAsync_RpcThrows_BannerPreservedForRetry             | tests/OpenClaw.Tray.Tests/OpenClawChatDataProviderTests.cs |
| Tray UX: allow routes RPC + clears banner            | RespondToPermissionAsync_AllowRoutesAllowOnceThroughRpcAndClearsBanner | tests/OpenClaw.Tray.Tests/OpenClawChatDataProviderTests.cs |
| Tray UX: deny routes RPC + clears banner             | RespondToPermissionAsync_DenyRoutesDenyThroughRpcAndClearsBanner       | tests/OpenClaw.Tray.Tests/OpenClawChatDataProviderTests.cs |

Test coverage added (cumulative across the PR)

ChatTimelineReducerTests (Shared)

 - StaleStreamingPreview_DoesNotMergeAcrossUserBoundary
 - UserMessage_AsTurnBoundary_PreventsCrossTurnOverwrite
 - AddLocalUser_AsTurnBoundary_PreventsCrossTurnOverwrite
 - AddLocalUser_ClearsStaleStreamingPreviewAcrossTurns
 - IntraTurnLateReconcile_AfterTurnEnd_DoesNotPrependPriorSegments (id-equality assertion locks the test to the in-place merge path)
 - IntraTurnLateReconcile_RegressionCumulative_DoesNotBlankCandidate (defensive blank-guard)

OpenClawGatewayClientApprovalTranslationTests (Shared, 12 tests)

 - Phase mapping for exec.approval.requested / resolved
 - Decision allowlist (IsValidApprovalDecision) — rejects unknown decisions
 - IsConnected precheck — throws when disconnected before send
 - Slug / UUID alt-id wiring for terminal-event matching
 - Multi-subscriber isolation: first subscriber throws, second still invoked
 - Response routing: ok:true completes, ok:false surfaces InvalidOperationException, ok:false without error.message uses fallback

OpenClawChatDataProviderTests (Tray, 3 tests)

 - RespondToPermissionAsync_AllowRoutesAllowOnceThroughRpcAndClearsBanner
 - RespondToPermissionAsync_DenyRoutesDenyThroughRpcAndClearsBanner
 - RespondToPermissionAsync_RpcThrows_BannerPreservedForRetry

Validation (against commit 77a707f7)

 - ./build.ps1: succeeds
 - OpenClaw.Shared.Tests: 2142 passed / 29 skipped
 - OpenClaw.Tray.Tests: 1004 passed

Review

Five rounds of adversarial dual-model review (Claude Opus 4.7 1M + GPT-5.3-Codex). Rounds 1–3 covered fixes 1–4. Round 4 on the multi-segment dedup surfaced 3 applied fixes (empty-result guard on the new strip site, id-equality assertion in the regression test, excludeIndex caller-contract xml-doc). Round 5 covered fix 6's ClawSweeper P1 follow-up and surfaced 5 applied fixes (Fix A leak in timeout branch, Fix B OCE taxonomy, Fix D 15s timeout, Fix E IsCompleted check, Fix G→H post-send re-check rework using TryRemove-first idiom) plus 1 documentation fix (Fix I contract comment on the cleanup-loop OCE). Each round converged on both models reporting "ready to commit."

Proof status (per ClawSweeper)

Code review labels: 🐚 platinum hermit (patch quality), 🦪 silver shellfish (overall, capped by proof). Test-only coverage for the rejected-resolve banner-preserve path is pinned by the coverage matrix above; live native-chat proof of a denied / rejected exec.approval.resolve keeping the banner visible is pending Mantis or manual QA. Maintainers may either request live proof, ask Mantis to capture it, or accept test-only coverage and re-review.

Deferred follow-ups (intentional, tracked outside this PR)

 - True bubble-splitting in native chat (rendering N visually separate bubbles per multi-segment turn) — this PR only ensures the merged final bubble contains the correct (last) segment's text without duplicating prior segments.
 - Plugin approvals (plugin.approval.*) — same shape, not exercised by current scenarios.
 - Full strict-send infrastructure replacing the silent-swallow behavior of SendRawAsync — acknowledged in code comments as best-effort.
 - Deeper integration tests for ResolveExecApprovalAsync exercising SerializeRequest, registration ordering, real timeout, and real disconnect paths (current tests use reflection seams against HandleResponse / _pendingApprovalResolves).
 - Sibling-wide hardening of the permissive ok:false detection pattern across _pendingChatSendRequests / _pendingWizardResponses — applying it only to exec.approval.resolve would create inconsistency; needs a focused PR.


RBrid:
Screenshots of multi-bubbles / multi-tool-calls:

image

image

RBrid:
@clawsweeper re-review

RBrid:
@clawsweeper re-review

RBrid:
Latest screenshot:
image

PR #689, opened 2026-06-04
fix(wizard): surface upstream console output, unwrap nested notes, auto-launch OAuth URLs

### LINKED ISSUE #640: Config Wizard: browser links for auth written to hidden terminal window
The WSL Claw install flow worked great. I got to the Configuration Wizard and ran into a few issues. 

The first is that I tried to configure Codex as my preferred model and that needs a browser login. The issue is that the browser link gets written to a hidden terminal window that `openclaw config` is being run in. It's impossible to know what link you're supposed to open and since there's no back button you can't change your mind about what model you want to use. You end up having to plug in a bogus value for the auth response to get the start over button.

I get that you're driving a terminal session so you can't really backup. But you should probably surface a "Start Over" button on the wizard as a safety valve.

If you do end up extracting the browser link from the text you should also make sure you look for any codes that are required to plugin. In some flows you have to open a browser and then typing in a code they showed you.

bkudiess:
Fixes #640.

## Problem

Upstream openclaw plugins emit user-critical content (OAuth URLs, install fallbacks, error reasons) via `console.log` to gateway stdout rather than as `wizard.payload` WS frames. The tray UI only renders `wizard.payload`, so during several model-provider onboarding paths the user saw a blank/stuck step and the wizard either silently looped or bounced back to the provider list.

All wizard copy and step content is owned by upstream openclaw; this PR does not modify any of that text. It adds tray-side backstops so the missing content reaches the user.

## Changes

Three layered, narrowly-scoped fixes plus one already-committed gateway fix on this branch (`32a2c7a` — enables the bundled `device-pair` plugin so OAuth providers can actually pair after install).

**1. `WizardConsoleTail`** — `src/OpenClaw.SetupEngine.UI/WizardConsoleTail.cs`
Spawns `wsl.exe -d OpenClawGateway -- bash -c "tail -F /tmp/openclaw/openclaw-*.log"` and surfaces root-logger `console.log` lines into a new `ConsoleBanner` panel on the active wizard step. Filters out structured subsystem noise. Silently no-ops when wsl/distro is unavailable (remote/Tailscale case).

**2. `WizardPayloadHelpers.ExtractStepMessage`** — `src/OpenClaw.SetupEngine.UI/WizardPayloadHelpers.cs`
Reads `step.message` and tolerates the case where a plugin nests a `{type:note, title, message}` object instead of sending a string. Returns the string as-is, `"Title\n\nMessage"` for a nested note, or empty (never raw JSON). Shape-based, not provider-keyed.

**3. `WizardUrlLauncher`** — same file
Auto-opens OAuth/device-code URLs via `Windows.System.Launcher.LaunchUriAsync`. Conservative host allowlist (`auth.openai.com`, `auth.x.ai`, `accounts.google.com`, `login.anthropic.com`, `console.anthropic.com`, `claude.ai`, `github.com`, `auth.openclaw.ai`) plus pattern match on `/oauth`, `/authorize`, `response_type=code`, `device_code`. Per-session dedupe via `HashSet`, cleared in `StartWizardAsync` so **Start over** resets cleanly. Docs/security links are skipped.

**Plumbing** — `Pages/WizardPage.xaml{,.cs}`
New `ConsoleBanner` `StackPanel` (collapsed by default). `_consoleTail` + `_launchedUrls` fields reset on `StartWizardAsync`. `ApplyPayloadAsync` uses the new helper. `AppendLineTo` extracted as a shared renderer for body + banner; URL branch triggers auto-launch.

Helpers are XAML-free and linked into the tray test project via `<Compile Include>`.

## Tests

- **39 new xUnit tests** (15 tail filter, 24 helpers).
- Build: 5/5 projects green, win-arm64.
- Shared tests: 2045 pass, 0 fail.
- Tray tests: 968 pass, 0 fail.
- Manual validation across 6 onboarding paths: Codex browser, Codex npm-download, Claude error, xAI OAuth, Gemini OAuth, Start over.


src/OpenClaw.SetupEngine.UI/Pages/WizardPage.xaml.cs
```diff
@@ -564,30 +576,107 @@ private void RenderMessage(string message)
...
+        {
+            target.Children.Add(BuildCodeRow(codeMatch.Groups[1].Value, codeMatch.Groups[2].Value));
+            return;
+        }
+
+        var urlMatch = Regex.Match(trimmed, @"https?://[^\s\)\""]+", RegexOptions.IgnoreCase);
+        if (urlMatch.Success && Uri.TryCreate(urlMatch.Value.TrimEnd('.', ','), UriKind.Absolute, out var uri))
+        {
+            target.Children.Add(BuildLinkLine(trimmed, urlMatch.Value, uri));
+            TryAutoLaunchUrl(uri);
```
kmahone: I think this might be confusing. If the launched browser covers the setup window, they won't see the code that they need to enter. They would need to know to switch back to copy the code. 

Would it be better to leave it to the user to manually click the link?

kmahone (changes requested):

bkudiess:
Good catch — pushed 02a81d9 dropping auto-launch entirely.

You're right that for device-code flows (xAI, Gemini, GitHub device login) the user needs to read a short code from the wizard first, then paste it in the browser. Popping the browser on top of the wizard breaks that sequence. The console tail still renders the URL as a clickable HyperlinkButton, so the user clicks when they're ready.

This also moots the @clawsweeper concern about the predicate being too broad — `WizardUrlLauncher` is gone along with its 16 tests.

PR #751, opened 2026-06-12
Add inline exec approval flow

RBrid:
## Summary

This PR routes local `system.run` exec-approval prompts into the native chat timeline when an interactive native chat surface is available, while preserving the native approval dialog as the fallback when chat is not visible/available.

Key behavior:

- Shows inline command approval cards in chat with **Allow once**, **Always allow**, and **Deny once** actions.
- **Allow once** approves only the current request.
- **Always allow** approves the current request and persists an exact allow rule to `exec-policy.json` when the command can be safely represented exactly.
- **Deny once** denies only the current request and does not persist a deny rule.
- Inline approval cards expire after the gateway approval window, collapse to an expired state, and no longer leave enabled approval buttons behind.
- Pending inline approvals also create a persistent app notification breadcrumb with an **Open chat** action that opens/selects the relevant chat session.
- The notification uses localized display labels such as "Main chat" instead of raw session keys like `agent:main:main`; raw keys are kept only for routing.
- If no native chat surface is active, or inline routing fails, the original native approval dialog remains the fallback path.

## Policy and compatibility fixes

- Thread session keys through `node.invoke` and `system.run` approval paths so local prompts can route to the correct chat session.
- Keep legacy exec-policy compatibility by accepting canonical `prompt`, legacy `ask`, and numeric enum action values.
- Preserve stable `ChatPermissionDecision` enum values.
- Fix the Permissions page to read/write the same `exec-policy.json` path that `NodeService` enforces, including isolated `OPENCLAW_TRAY_DATA_DIR` runs.
- Remove the load-time policy rewrite that could project rules through the reduced UI model.
- Preserve policy fields (`shells`, `description`, `enabled`) on explicit save.
- Omit default `enabled: true` on save while preserving `enabled: false`.
- Keep explicit deny rules as hard denies; only `prompt` rules/defaults show approval UI.

## Review-driven fixes addressed

- ClawSweeper: stopped load-time policy rewrites and preserved full exec-policy rule fields.
- Hanselman/Codex: fixed tray chat active-state cleanup, notification chat-action dedupe, terminal approval phase mapping, pending notification lifecycle, stale deny notification clearing after manual approval, and user-facing session labels.
- Localized new/changed user-facing strings across supported locales and updated localization validation coverage.

## Proof screenshots

Type of (cut-off) UI that moved inline into chat history:

image

Existing inline approval style for comparison:

image

Resulting inline approval UI:

image

After clicking **Deny once**:

image

After clicking **Allow once**:

image

After clicking **Always allow**:

image

Then subsequent matching requests do not require re-approval:

image

Permissions rule action dropdown fix:

image

## Testing and review

Latest validation:

- Build: `./build.ps1` passed.
- Shared tests: `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore` passed (`2374 passed, 29 skipped`).
- Tray tests: `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore` passed (`1149 passed`).
- Localization validation: `LocalizationValidationTests` passed; new resource keys are present across supported locales with placeholder parity.

Manual testing performed with an isolated data directory:

- Verified Permissions edits persist to the same isolated `exec-policy.json` enforced by the node.
- Verified default `prompt` policy triggers inline approval for `tasklist`.
- Verified **Deny once**, **Allow once**, and **Always allow** flows.
- Verified **Always allow** persists an exact allow rule and survives app restart.
- Verified repeated matching request runs without re-approval after restart.
- Verified inline approval timeout collapses the approval card and disables action buttons.
- Verified pending approval notification opens/selects the relevant chat session.

Review notes:

- Bundled autoreview was run with the available Copilot engine because the default `codex` CLI is not installed in this environment; the run reported no accepted/actionable findings.
- Multiple Hanselman dual-model reviews were run during iteration. Accepted/actionable findings were addressed and followed by the validation above.


RBrid:
@clawsweeper re-review

shanselman:
Maintainer note: this is a substantial exec-approval UI flow and it is currently conflicting. Given the size and the security boundary involved, I don't think this is safe to resolve opportunistically during triage. Please rebase on current `main` (especially after the exec approvals persistence work that just landed), then add a short proof section showing the allow/deny/remember flows and the tests that cover them.

PR #741, opened 2026-06-10
Harden gateway connection and pairing flows

ranjeshj:
## Summary

Harden Windows companion gateway connection, setup-code pairing, node pairing, token recovery, and browser-control auth flows.

Key changes:
- keep operator and node WebSocket lifecycles separate, with setup-code operator connects using `role=operator`
- persist all returned role handoff tokens from `hello-ok.auth.deviceTokens[]`
- clear bootstrap credentials only after required role tokens are durably readable
- forward node device-token receipt from `WindowsNodeClient` through `NodeConnector` so bootstrap cleanup can complete after node reconnect
- request and handle pairing scopes correctly, including response-aware node/device approval and admin-gated fallback
- prevent stale node clients from mutating current state and abort node handshakes when capability registration fails
- preserve shared-token HTTP/dashboard/browser-control semantics, including browser proxy registration only when a shared gateway token is available
- add local MCP-only connection controls without exposing them through the gateway node transport
- add WSL setup retry for transient post-terminate `/etc/wsl.conf` read timing while keeping config validation strict
- add protocol research documentation and expanded connection/E2E coverage

## Validation

- `./build.ps1` passed
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore` passed
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore` passed
- `dotnet test ./tests/OpenClaw.Connection.Tests/OpenClaw.Connection.Tests.csproj --no-restore` passed
- `dotnet test ./tests/OpenClaw.SetupEngine.Tests/OpenClaw.SetupEngine.Tests.csproj --no-restore` passed
- `OPENCLAW_RUN_E2E=1 dotnet test ./tests/OpenClaw.E2ETests/OpenClaw.E2ETests.csproj -r win-arm64 --no-restore` passed: 17/17
- Copilot autoreview clean: no accepted/actionable findings
- Dual-model protocol audit completed; accepted multi-role handoff finding fixed and revalidated

## Notes

Draft PR for review. The changes intentionally avoid broad WebSocket lifecycle rewrites or gateway-side assumptions; fixes are scoped to client-side behavior verified against `docs/CONNECTION_PROTOCOL_RESEARCH.md` and `docs/CONNECTION_ARCHITECTURE.md`.

PR #760, opened 2026-06-14
fix(browser.proxy): only advertise when shared gateway token is available

QQSHI13:
## Problem

The Windows node advertises the `browser.proxy` capability whenever a gateway connection exists. However, the browser-control host running on `gatewayPort + 2` authenticates using the **shared gateway token**, not the node's device/bearer token.

As a result, a node that is paired via QR/bootstrap code (device token only) will:
1. Register `browser.proxy`.
2. Receive `browser.request` invocations from the gateway.
3. Proxy them to `http://<gateway>:<gatewayPort+2>` using `_token` (the node bearer/device token).
4. Be rejected with an authentication error by the browser-control host.

Upstream PR #741 attempted to address this, but it did not change the token passed to `BrowserProxyCapability` or gate registration on availability of the shared token. In practice the capability still advertised itself and still authenticated with the wrong token.

## Fix

This PR makes `NodeService` only register `browser.proxy` when the active `GatewayRecord` has a `SharedGatewayToken`. When present, that token is passed to `BrowserProxyCapability` so the browser-control host accepts the proxy requests.

Key changes:
- `App.EnsureNodeService` now supplies a `sharedGatewayTokenResolver` that reads `_gatewayRegistry?.GetActive()?.SharedGatewayToken`.
- `NodeService.RegisterCapabilities` resolves the shared token before constructing `BrowserProxyCapability`. If absent, registration is skipped and a diagnostic log line explains why.
- The token passed to `BrowserProxyCapability` is changed from `_token` (node bearer token) to the resolved shared gateway token.
- Updated the Command Center diagnostic message to accurately describe the requirement.

## Verification

- Built and ran the patched tray app from a Windows temp copy.
- Confirmed `browser.proxy` is advertised when a shared gateway token is configured.
- Confirmed `browser.proxy` is **not** advertised when only a QR/bootstrap device token is available.
- Confirmed browser proxy requests authenticate successfully against the browser-control host with the shared token.

## Scope

This is a minimal, targeted fix. It does not change gateway or CLI behavior; those surfaces still need to ensure the shared token is available to the node. It also does not alter the capability gating rules in `NodeCapabilityGating`.

### DIFF
--- a/src/OpenClaw.Shared/Models.cs
+++ b/src/OpenClaw.Shared/Models.cs
@@ -92,6 +92,13 @@ public class OpenClawNotification
     public string? Agent { get; set; }     // agent name/identifier
     public string? Intent { get; set; }    // normalized intent (reminder, build, alert)
     public string[]? Tags { get; set; }    // free-form routing tags
+
+    /// <summary>
+    /// The session key associated with this notification (e.g. the chat session
+    /// that produced an assistant response). Used by toast activation to open
+    /// the specific session instead of the default/main session.
+    /// </summary>
+    public string? SessionKey { get; set; }
 }
 
 /// <summary>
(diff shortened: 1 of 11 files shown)


QQSHI13:
Opened a clean, focused follow-up as #761 with additional evidence showing the issue is still present on current main. The previous PR mixed in unrelated chat-session routing commits; the new branch is based directly on upstream/master and only changes the browser.proxy auth path.

PR #777, opened 2026-06-17
Add Run gateway doctor action to the Diagnostics page

bkudiess:
## Summary

Adds a **Run gateway doctor** action to the Diagnostics page (`DebugPage`). The card opens a terminal in the app-managed WSL gateway, runs `openclaw doctor`, and leaves an interactive shell open so the operator can read the diagnostic report.

## Behavior

- Shown **only** when the active gateway is an app-managed WSL distro (`GatewayHostAccessPlan.CanControlWslGateway`). Hidden for SSH/remote/none.
- Reuses the same launcher path as the Connection page's **Open terminal** (`GatewayTerminalLauncher` / `GatewayHostAccessClassifier`): opens a themed Windows Terminal tab, with a direct `wsl.exe` fallback.
- The card lives in a conditional **Gateway** section above *Share diagnostics with support*; visibility is recomputed in `Initialize()` and on `AppState`/`Settings` changes.

## Why the keep-open avoids `;`

The keep-open command is `<PATH prefix> && openclaw doctor || true && exec bash` — deliberately **no `;`**. Windows Terminal treats `;` in its command line as a tab/pane delimiter and splits on it *even inside a quoted argument*, which caused `error 0x80070002` when launching. `&&`/`||` are not wt delimiters, so the command passes through intact (verified by reproducing the app's `ProcessStartInfo.ArgumentList` quoting). `|| true` absorbs doctor's frequent advisory non-zero exit so `exec bash` always runs.

`Start()` now uses `UseShellExecute = !command.UsesWindowsTerminal` so the direct `wsl.exe`/`ssh.exe` paths get a visible console window.

## UX rationale

`openclaw doctor` is a rich terminal TUI (banner, box-drawing, step spinners, ANSI color) that exits non-zero for advisory findings. Capturing it headless produces mojibake and an unreliable pass/fail signal, so the action is terminal-first (the operator reads the report directly).

## Changes

- `GatewayTerminalLauncher.cs` — `BuildGatewayDoctor` + `OpenGatewayDoctor`; `Start()` shell-execute tweak.
- `FluentIconCatalog.cs` — new `Doctor` glyph (U+E95E).
- `DebugPage.xaml` / `.cs` — Gateway section, card, gating + handler.
- `Resources.resw` (en-us + 4 locales) — 3 new keys.
- Tests — command-builder tests + Diagnostics page contract test.

## Validation

- `./build.ps1` ✅
- `dotnet test OpenClaw.Tray.Tests` — **962 passed** ✅
- `dotnet test OpenClaw.Shared.Tests` — **2049 passed** ✅
- Manual smoke on a real `OpenClawGateway` WSL distro: card appears, opens the themed terminal tab, runs doctor, drops to an interactive shell.
- Code review: no actionable findings.

### DIFF
--- a/src/OpenClaw.Tray.WinUI/Helpers/FluentIconCatalog.cs
+++ b/src/OpenClaw.Tray.WinUI/Helpers/FluentIconCatalog.cs
@@ -93,6 +93,7 @@ public static class FluentIconCatalog
     public const string Clear = "\uE74D";          // Delete — clear/reset a buffer
     public const string Develop = "\uE943";        // Code — engineering / explorations action
     public const string AgentEvents = "\uE81C";    // History — agent events feed
+    public const string Doctor = "\uE95E";         // Health — "Run gateway doctor" health-check action
 
     // ── Agents / Workspace surface ─────────────────────────────────
     // Workspace concept (per-agent file viewer). Reuses the Folder
--- a/src/OpenClaw.Tray.WinUI/Pages/DebugPage.xaml
+++ b/src/OpenClaw.Tray.WinUI/Pages/DebugPage.xaml
@@ -85,6 +85,32 @@
                         </controls:InfoBar.Content>
                     </controls:InfoBar>
 
+                    <!-- Gateway — shown only when the active gateway is an
+                         app-managed WSL distro we can run commands in
+                         (CanControlWslGateway). Hidden for SSH/remote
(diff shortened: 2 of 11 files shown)


ranjeshj:
Closed as superseded by #780, which includes this change plus conflict-resolution, validation, localization, and review fixes.

PR #793, opened 2026-06-20
browser.proxy: allow overriding the browser-control port for split/remote gateways

anagnorisis2peripeteia:
`BrowserProxyCapability` hard-derives the browser-control host as `gatewayPort + 2` on `127.0.0.1`. That is correct for a co-located gateway, but breaks for a **split/remote gateway** — e.g. the gateway running in WSL2 with the Companion on the Windows host, where the browser-control host is reached on a different local port (an SSH-forwarded port). Every `browser.proxy` request then fails as unreachable at the derived port.

This adds one optional override — `BrowserControlPort` — so the Companion can target the actual forwarded browser-control port. **Default is unchanged**: when unset, the port is derived exactly as today (`gatewayPort + 2`).

The browser-control **auth token is unchanged** — it continues to come from the per-gateway shared token (`record.SharedGatewayToken`) via the existing credential path, so this PR adds **no new credential storage** and stays within the `GatewayRegistry` credential model.

## Changes
- `OpenClaw.Shared/SettingsData.cs` — add `BrowserControlPort` (int?). Null (default) = derive as today.
- `OpenClaw.Tray.WinUI/Services/SettingsManager.cs` — forwarder.
- `OpenClaw.Shared/Capabilities/BrowserProxyCapability.cs` — ctor gains `controlPortOverride`; `TryResolveControlEndpoint` honours an explicit port (TCP-range validated) else `gatewayPort + 2`.
- `OpenClaw.Tray.WinUI/Services/NodeService.cs` — pass the setting into the capability ctor.
- `tests/OpenClaw.Shared.Tests/CapabilityTests.cs` — +3 `[Fact]`: port override targets the configured port; out-of-range port errors cleanly; and a real-HTTP round-trip proving the override directs genuine traffic.

## Scope boundary
- Port-only: a port is not a credential, so this stays clear of the `GatewayRegistry` credential boundary.
- Settings-editable; no Settings UI added (advanced/optional override).
- No gateway-side change; Companion-only.

## Tested
- `dotnet test tests/OpenClaw.Shared.Tests` → **2268 passed, 0 failed, 29 skipped**.
- **Real behaviour proof (not just unit logic):** one of the new facts, `BrowserProxy_ControlPortOverride_RealHttpRoundTripHitsOverridePort`, drives the actual capability with a **real `HttpClient`** (no mock handler) against a **real loopback HTTP server**. The gateway URL's port (18790) would derive control port 18792 — the wrong, unreachable port in a port-remapping tunnel — and the test asserts the request actually arrives over a real socket at the override port, carrying the per-gateway shared token, and returns 200. So the override is shown to direct genuine TCP/HTTP traffic, not just build a URL string.
- `dotnet build src/OpenClaw.Tray.WinUI -c Debug -r win-x64` → all assemblies compile.

## Not tested
- No Settings UI, so no UI test.
- Not captured against a production remote gateway in CI — the real-HTTP integration test exercises the actual socket path on loopback; the motivating real-world repro was a WSL2 ↔ Windows packet capture showing the request failing to reach the host at the derived port.

## Risk / mitigation
- *Misconfigured port* → out-of-range is validated and returns a clear error; unset falls back to today's `gatewayPort + 2`.
- *No credential change* → the auth-token path is untouched (per-gateway shared token), so there is no new secret storage or gateway-switch credential concern.

---
*Supersedes the initial revision, which also added a global `BrowserControlToken` setting. Dropped after review: current source already sends the per-gateway shared token (`record.SharedGatewayToken`) to the browser-control host (`NodeService` comment + ctor), so the token override was redundant — the only genuine gap was the hard-coded control port.*


anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
## Live behavior proof + full validation — `browser.proxy` split/remote (final head `4d7bda9`)

### Live 4-scenario proof (real `BrowserProxyCapability` + real `HttpClient` over real TCP, no mocks)
```
A  no override             -> control 18791  Ok=False   (split/remote breakage)
B  override 19000          -> Ok=True                   (#793 working: forwarded control host reached)
C  tunnel-local 18998 +2   -> Ok=True                   (browser-control scoped to the active tunnel)
D  tunnel->direct SWITCH   -> Ok=False, token-misrouted-to-stale-host=False
                                                        (active GatewayRecord wins; no stale-tunnel token leak)
PROOF_RESULT=PASS
```
Scenario B/C log `Authorization header present: True` on the real control host; Scenario D proves the post-switch request never reaches the stale 19000 listener. (The capability dial path is unchanged since `de0ae66`; `4d7bda9` only changes the diagnostics/topology layer below.)

### Full AGENTS-required validation set (final head `4d7bda9`)
| Command | Result |
|---|---|
| WinUI build (`dotnet build OpenClaw.Tray.WinUI -p:Platform=x64`) | **Compiles + links** — `OpenClaw.Tray.WinUI.dll` emitted. Only the post-Build `wxc-exec.exe` copy fails (MSB3030: npm `@microsoft/mxc-sdk` binary absent locally; present in CI). Zero `error CS`. |
| `dotnet test OpenClaw.Shared.Tests` | **2286 passed / 0 failed** (29 skipped) |
| `dotnet test OpenClaw.Tray.Tests` | **1098 passed / 0 failed** |
| `dotnet test OpenClaw.Connection.Tests` | **333 passed / 0 failed** |

### Patch-quality findings closed
- **NodeService stale global tunnel** → `BrowserProxyTunnelState.Resolve`: active `GatewayRecord` authoritative; null active tunnel = direct gateway, no global inheritance. (Scenario D, live.)
- **Command Center probe mismatch** → `PortDiagnosticsService.TryGetBrowserProxyPort` resolves through the same `BrowserControlEndpoint` the capability dials.
- **Command Center *topology* used stale settings tunnel** (`4d7bda9`) → `GatewayTopologyClassifier.Classify` was fed tunnel params from global `SettingsManager`, so after a tunnel→direct gateway switch the diagnostics probed `tunnel-local+2` and warned about a tunnel the active gateway no longer used. Now derived via `CommandCenterTopologyTunnelResolver` from the active `GatewayRecord` (mirrors `BrowserProxyTunnelState`): null `SshTunnel` = direct, no settings inheritance; falls back to settings only on the legacy no-registry path. This is the diagnostics-layer twin of the dial-layer fix.
- **Saved-gateway edit drops the override** → `GatewayRecordEditing.PreserveAdvancedFields(existing)`, called by both `ConnectionPage` save and `ConnectionStatusWindow` connect, carries `BrowserControlPort` across name/token/URL/SSH edits.

### Regression tests added
- `BrowserProxyEndpointScopingTests` (10) — gateway-switch tunnel scoping + override-diagnostic consistency (co-located / tunnel / remote-kind), plus 3 new `CommandCenterTopology_*` cases covering direct-gateway stale-settings bypass, active-tunnel record ports, and the legacy settings fallback.
- `GatewayRegistryTests.PreserveAdvancedFields_*` (2) — override survives a saved-gateway edit; form value wins; null-existing is a no-op.

The one remaining item is the maintainer design call: confirming the per-gateway config surface that selects the loopback endpoint receiving the shared browser-control token (the deliberate contract decision, not a defect).

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

anagnorisis2peripeteia:
@clawsweeper re-review

shanselman (approved):
Validated locally with build, shared/tray/connection tests, autoreview, and security review after maintainer hardening for browser.proxy endpoint scoping.

PR #799, opened 2026-06-21
Execute approved exec-approval commands as a resolved, shell-free argv

AlexAlves87:
## Summary

The exec-approvals V2 path could decide to allow a command but had no way to run
it: the allow result carried no payload, and the local runner could only execute
through a shell that re-tokenizes the command. This change makes an approved
command executable exactly as it was evaluated.

- The allow result now carries the approved execution payload: the validated
  argument vector with its executable resolved to an absolute path, the sanitized
  environment, the working directory, and the timeout. argv[0] is the resolved
  path, never the raw command token, so the executable cannot be re-resolved
  against PATH or the working directory at execution time. If the executable
  cannot be resolved, the allow fails closed.
- LocalCommandRunner gains a direct-argv mode: when a request carries an argv,
  the process is launched with ProcessStartInfo.ArgumentList and
  UseShellExecute=false, so the approved arguments reach the process verbatim
  with no shell re-parsing them. The legacy shell-wrapped path is unchanged when
  no argv is supplied.
- When the approved command uses a transparent `env` wrapper (e.g. `["env", "git",
  "status"]`), the payload is built from the effective argv the resolver evaluated,
  not the original request tail, so the approved and executed command identity
  match exactly.

## Safety

The payload builder fails closed before any approval state is written, so a
command that cannot be represented faithfully never leaves an allowlist entry
or usage record behind. It rejects:

- An unresolved executable (no absolute path to pin the command identity to).
- A batch script (.bat/.cmd), which needs cmd.exe and would re-parse the arguments.
- An `env` wrapper carrying modifiers (VAR=val assignments or flags such as `-i`
  or `--unset`), whose environment semantics a direct-argv payload cannot carry,
  including nested forms such as `env env FOO=bar node`, whose modifier the resolver
  would otherwise strip. Transparent `env` wrappers (including nested) are unwrapped safely.

The direct-argv runner enforces the same guards independently:

- An empty argv, a non-absolute executable (which Windows would otherwise guess
  from PATH/cwd), or a batch script are rejected rather than degraded to a shell.
- A non-null but empty argv is treated as invalid, never as a request to fall
  back to the shell.
- Argument values are not written to logs in direct mode; only the executable
  name and the argument count are logged.
- The payload takes defensive copies of argv and env so it cannot be mutated
  between approval and execution.

The sandboxed runner has no argv transport yet, so it fails closed rather than
degrade silently: when sandboxing is available and enabled and a request carries
a direct argv, it is blocked instead of being serialized as the legacy command
fields. When sandboxing is unavailable or disabled, the request routes to the
host runner, which does honor argv.

## Scope

This adds the capability to execute an approved command faithfully. It does not
enable the new path in production: wiring the coordinator behind its feature flag,
and a faithful argv transport for the sandboxed runner, are separate changes. The
new path stays unreachable until those land, and the sandbox fails closed in the
interim.

## Proof

A standalone program approves a command carrying deliberately hostile arguments
(spaces, an empty string, an embedded quote, a trailing backslash, shell
metacharacters, a tab, and Unicode with an emoji), runs it through the real
coordinator and the real runner, and has the launched child process dump the argv
it actually received. Each argument is shown with its character length and its
exact UTF-8 bytes in hex, so a space, a tab, and an empty string are
distinguishable rather than ambiguous on screen. No pass/fail verdict is printed
by the program: the three blocks are compared by reading the bytes.

- APPROVED — what was handed to the coordinator.
- PAYLOAD — what the coordinator emitted (argv[0] resolved to an absolute path,
  arguments unchanged).
- RECEIVED — what the child process got, dumped by the child itself.

APPROVED and RECEIVED match byte for byte; PAYLOAD is the same arguments with the
resolved executable as argv[0].

_Proof output below:_

Captura de pantalla 2026-06-21 145455

## Testing

Local validation after latest changes:

```
./build.ps1 (full solution build) — succeeded, 0 errors, 0 warnings
dotnet test OpenClaw.Shared.Tests  — 2333 passing, 0 failed
dotnet test OpenClaw.Tray.Tests    — 1088 passing, 0 failed
```

Coverage includes: launch-planning decision (direct-argv vs legacy, verbatim
argument preservation, fail-closed guards), payload builder (resolved path becomes
argv[0]; unresolved executable, batch script, and modified env wrapper each produce
no payload), empty-argv guard, sanitized environment in payload, end-to-end
coordinator→runner handoff without shell, regression for transparent env wrapper,
no allowlist persistence when the payload fails closed, and the sandboxed runner
failing closed on a direct-argv request while the host fallback still honors argv.

### DIFF
--- a/src/OpenClaw.Shared/ExecApprovals/ExecApprovalV2Result.cs
+++ b/src/OpenClaw.Shared/ExecApprovals/ExecApprovalV2Result.cs
@@ -1,6 +1,8 @@
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
+using OpenClaw.Shared;
 
 namespace OpenClaw.Shared.ExecApprovals;
 
@@ -14,6 +16,8 @@ namespace OpenClaw.Shared.ExecApprovals;
 /// </summary>
 public sealed record ExecApprovedExecution
 {
+    public const int MaxTimeoutMs = 600_000;
+
     public IReadOnlyList<string> Argv { get; }
     public string? Cwd { get; }
     public int TimeoutMs { get; }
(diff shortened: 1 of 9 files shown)


AlexAlves87:
@clawsweeper re-review

AlexAlves87:
@clawsweeper re-review

AlexAlves87:
@clawsweeper re-review

AlexAlves87:
@clawsweeper re-review

AlexAlves87:
@clawsweeper re-review

AlexAlves87:
@clawsweeper re-review

PR #813, opened 2026-06-24
Add gateway protocol drift guard (sessions/files/commands/compaction)

bkudiess:
## Summary

Adds a **gateway protocol drift guard** so the Windows companion stops silently drifting from the upstream OpenClaw gateway protocol. It pins a hand-maintained, **upstream-verified** mirror of the `sessions` / `files` / `commands` / compaction surface and statically cross-checks it against the typed client, so future drift fails CI instead of shipping. Because the gateway tolerates unknown/extra fields, this kind of drift otherwise ships without any test failing.

## What it adds

- **`tests/OpenClaw.Shared.Tests/Protocol/gateway-protocol-snapshot.json`** — canonical pin of the protocol surface: methods, request wire-keys, response envelopes, `sessions.list` response shape, and the `sessions.patch` tri-state nullable clear contract.
- **`tests/OpenClaw.Shared.Tests/Protocol/GatewayProtocolDriftTests.cs`** — six guards:
  - method surface (dispatched requests vs handled notifications)
  - `sessions.list` response field set
  - request wire-keys (actual constructed keys, not token presence)
  - response envelope read, method-scoped
  - snapshot integrity + required-method coverage
  - tri-state clear contract (`SessionPatch.Clear` -> explicit JSON null; value -> sent; null reference -> omit; blank -> omit)
- **`docs/gateway-protocol-drift-guard.md`** — how it works, enforced-vs-doc-only table, refresh steps.
- `ProductionSourceFiles.FindRepoRoot` made `internal` for reuse.

Deterministic and **repo-contained**: no upstream clone, no network. It complements (does not duplicate) the behavioural `GatewayProtocolModelsTests` and `GatewayProtocolLiveRoundTripTests` — those assert runtime JSON emission / wire frames; this guards the static method/field/envelope surface.

## Upstream verification

Verified 2026-06-23 against `openclaw/openclaw` **main** (`packages/gateway-protocol/src/schema/sessions.ts` blob `d938f80f`, `commands.ts` blob `a2497216`). All enforced pins — request fields, response envelopes, and the `sessions.patch` `Union([value, Null])` tri-state contract — **match upstream exactly**. The snapshot is a deliberate subset: some upstream `sessions.patch` fields (`label`, `agentId`, spawn/subagent metadata) and compaction-checkpoint `preCompaction`/`postCompaction` references are not implemented by the client yet — recorded in the snapshot's `upstreamVerified` note as not-yet-implemented, not drift.

## Validation

- `./build.ps1` — all projects pass
- `OpenClaw.Shared.Tests` pass (incl. the 6 drift guards)
- `OpenClaw.Tray.Tests` pass

Builds on the typed gateway protocol client already merged to `main`.

### DIFF
--- a/docs/gateway-protocol-drift-guard.md
+++ b/docs/gateway-protocol-drift-guard.md
@@ -1,0 +1,233 @@
+# Gateway Protocol Drift Guard
+
+> Source of truth: [`openclaw/openclaw` — `packages/gateway-protocol/src/schema/{sessions,commands}.ts`](https://github.com/openclaw/openclaw/tree/main/packages/gateway-protocol/src/schema)
+
+## Why this exists
+
+The Windows companion can silently drift from the upstream OpenClaw gateway
+protocol — when methods or session fields change upstream but the Windows client
+(`OpenClawGatewayClient`) keeps sending the old shapes. Because the gateway
+tolerates unknown/extra fields, such drift ships without any test failing.
+
+The **drift guard** pins a hand-maintained mirror of the upstream gateway protocol
+for the `sessions` / `files` / `commands` / compaction surface and fails the test
+suite whenever the typed client and the pinned schema diverge.
+
+The typed client implements the richer protocol surface —
+`commands.list`, the extended `sessions.patch` field set, `sessions.files.list`/
+`get`, and `sessions.compaction.list`/`get`/`branch`/`restore` — across
+`OpenClawGatewayClient.Protocol.cs` and the DTO/payload builders in
+`GatewayProtocolModels.cs`. This guard pins that whole surface.
+
+It is intentionally **deterministic and repo-contained**: it requires neither a
+clone of the upstream `openclaw/openclaw` repo nor any network access, so it runs
+in the normal `OpenClaw.Shared.Tests` suite. It **complements** the behavioural
+parser tests in `GatewayProtocolModelsTests` (which parse sample JSON and assert
+DTO mapping) by statically guarding the method / request-field / response-envelope
+*surface*, so an upstream rename or a dropped API is caught even when the parser
+unit tests still pass on their fixed sample payloads.
+
+## Files
+
+| File | Role |
+| --- | --- |
+| `tests/OpenClaw.Shared.Tests/Protocol/gateway-protocol-snapshot.json` | The pinned canonical schema mirror. **Edit this when upstream changes.** |
+| `tests/OpenClaw.Shared.Tests/Protocol/GatewayProtocolDriftTests.cs` | The guard. Cross-checks the snapshot against `OpenClawGatewayClient*.cs` + `GatewayProtocolModels.cs`. |
+
+## What the guard checks
+
+1. **Method surface** (`ScopedMethodSurface_matches_snapshot`)
+   - Every in-scope method the client actually *wires* — request methods it
+     **dispatches** (a method literal passed as a call argument to any dispatcher,
+     e.g. `TrySendTrackedRequestAsync("sessions.patch", …)`,
+     `TryRequestPayloadAsync("commands.list", …)`, or the private
+     `MutateCompactionAsync("sessions.compaction.branch", …)` forwarder), plus
+     notifications it **handles** via `case "…":` — must be pinned in the snapshot.
+   - Every method pinned `windowsUsage: "used"` must still be wired in the way its
+     `kind` implies: a **request** must still be *dispatched* (a leftover `case`
+     label, a `method is "…"` predicate, or a log sentence is not enough), a
+     **notification** must still be *handled*. Dropping one is the classic
+     drift regression.
+   - A `planned` method must **not** already be wired — if the client starts using
+     it, the guard forces you to flip it to `used` and verify its response shape.
+     (None currently — the typed client wires the whole surface.)
+2. **`sessions.list` response shape** (`SessionsList_responseShape_matches_snapshot`)
+   - The session wire-fields the client parses (scoped to the two session-parsing
+     methods) must exactly equal the snapshot's `responseFields` (both directions).
+3. **Request parameters** (`RequestParameterNames_arePresentInConstructionRegion`)
+   - For each `used` request method, the set of wire keys the client actually
+     constructs in the request-construction region must **exactly equal** the pinned
+     `requestFields` (modulo a per-method `allowedExtraRequestFields` allowlist).
+     The comparison is **bidirectional**:
+     - **missing** (pinned − constructed): the client no longer sends a field the
+       schema expects — e.g. renaming a wire key while the old name survives as a
+       local/parameter (`new { sessionKey = key }` when `key` is still pinned) is
+       caught, because the check compares *constructed keys*, not token presence.
+     - **unexpected** (constructed − pinned − allowed): the client still constructs
+       a stale/extra key that is no longer in the schema. Strict gateways
+       (`additionalProperties:false`) reject the whole request on an unknown field,
+       so this is real drift. Set `allowedExtraRequestFields` on a method to declare
+       an intentional client-only extra the gateway tolerates.
+   - A constructed wire key is an anonymous-object member (`new { sessionKey = key,
+     path }`) or a dictionary string key (`payload["model"] = …`) in the region.
+   - The region is the enclosing block of an actual dispatch call, or — when the
+     payload is built by a helper rather than inline — the explicit
+     `requestFieldsSource` builder body (the extended `sessions.patch` fields are
+     checked against `SessionPatch.ToPayload`; the compaction `branch`/`restore`
+     params against `MutateCompactionAsync`). `requestFieldsSource` must resolve
+     **uniquely** to a body that actually constructs keys, and a dispatch from an
+     expression-bodied member (where the region would degrade to the whole type)
+     **fails closed** asking for a `requestFieldsSource`.
+4. **Response envelope** (`ResponseEnvelopes_areReadByClient`)
+   - For each `used` method that pins a `responseEnvelope`, the client must read
+     that property — `TryGetProperty("…")` **or** `TryGetArray(payload, "…")` —
+     **inside that method's own parser body** (pinned via `responseEnvelopeSource`),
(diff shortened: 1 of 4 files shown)


bkudiess:
## Proof of effectiveness — drift guard catches real drift

Adding concrete evidence that the six guards are **non-vacuous** (they actually fail when the protocol drifts), that they verify against **real upstream**, and that the full suite is green.

### 1. Mutation matrix — every guard fails on a realistic drift

For each guard I injected a realistic protocol drift, ran that guard, then restored the file from `HEAD` and confirmed clean. Two flavors: **SOURCE** mutations simulate the Windows client drifting; **SNAPSHOT** mutations simulate an upstream schema change left unreconciled.

| Guard | Injected drift | Caught | Failure message produced |
|---|---|:--:|---|
| A method-surface | SOURCE: client renamed `case "sessions.changed"` handler | ✅ | `client wires 'sessions.changedXYZ' but it is NOT pinned in the snapshot` |
| B sessions.list-shape | SNAPSHOT: upstream added `newUpstreamFieldXYZ` | ✅ | `sessions.list response shape drift detected` |
| C request-wire-keys | SOURCE: renamed wire key `sessionKey`→`sessionKeyV2` (local `key` still present) | ✅ | `sessions.files.get.sessionKey: not constructed as a request wire key … Constructed keys: [path, sessionKeyV2]` |
| D response-envelope | SNAPSHOT: renamed `commands.list` envelope `commands`→`commandsV2` | ✅ | `commands.list: response envelope property "commandsV2" is never read … by its parser` |
| E snapshot-integrity | SNAPSHOT: `provisionalResponseFields=true` on a `used` method | ✅ | `provisionalResponseFields=true is only allowed while windowsUsage="planned"` |
| F tri-state-contract | SOURCE: broke **only** `AddEncoded`'s clear→null (left `AddString` intact) | ✅ | `tri-state contract marker 'AddEncoded emits explicit null on clear' … is missing from 'void AddEncoded'` |

**6/6 guards caught their drift.** Note guard C catches a wire-key rename even though the old name survives as a local variable (it compares against *constructed* keys, not token presence), and guard F's per-helper split catches an `AddEncoded`-only regression that a single broad check would miss. Working tree restored clean afterward (`git status` empty, `HEAD` unchanged); guards green again post-restore.

### 2. Upstream verification still current

Re-fetched the upstream schemas from `openclaw/openclaw` **main** today; both blob SHAs are unchanged from the snapshot's recorded `upstreamVerified` block:
- `packages/gateway-protocol/src/schema/sessions.ts` → `d938f80f68cb285719e20801e83f74893aa04332` ✅
- `packages/gateway-protocol/src/schema/commands.ts` → `a24972162dfbe4f95e05bf19211097707d30f9bb` ✅

All enforced pins (request fields, response envelopes, the `sessions.patch` `Union([value, Null])` tri-state contract) match upstream exactly.

### 3. Full suite green

- `./build.ps1` — all 5 projects ✅
- `OpenClaw.Shared.Tests` — 2401 passed / 0 failed (includes the 6 guards)
- `OpenClaw.Tray.Tests` — 1123 passed / 0 failed

The mutation matrix is reproducible — each step injects one drift, asserts the specific guard fails, and restores via `git checkout`.

shanselman:
Thanks for the solid protocol-drift guard work. I ran a dual-model adversarial review as part of today's triage. The branch is close, but I don't think we should merge it until one small coverage gap is handled.

The request-field guard currently verifies that pinned `requestFields` are present in the construction region, but it does not appear to fail when the client constructs extra wire keys that are no longer in the snapshot. That means if upstream removes/renames a request field, the snapshot is updated, and the client still sends the stale extra key, this class of drift can slip through.

Could you either:

1. add the reverse check for constructed request keys not present in the snapshot, with an explicit allowlist for intentional extras, or
2. document this as an intentional presence-only limitation in the test comments and `docs/gateway-protocol-drift-guard.md`?

Preferred fix shape: compare both directions for used request methods: `missing = snapshotFields - constructedKeys` and `unexpected = constructedKeys - snapshotFields - allowedExtras`, then fail on either non-empty set with an actionable message.

bkudiess:
@shanselman thanks — good catch, that was a real gap. I implemented your preferred fix (option 1): the request-field guard is now **bidirectional**.

For each `used` request method it now compares the pinned `requestFields` against the wire keys the client actually constructs in the region, both directions:
- `missing  = pinned − constructed` (the original check)
- `unexpected = constructed − pinned − allowedExtras`

and fails on either non-empty set with an actionable message. I added a per-method `allowedExtraRequestFields` allowlist in the snapshot for intentional client-only extras (currently empty — no method needed one; the existing construction regions are exactly the pinned sets).

Proof it works both ways (mutations on a real checkout, restored after each):
- **unexpected** — inject a stale extra key `new { sessionKey = key, path, staleExtraKey = key }` → fails:
  `sessions.files.get.staleExtraKey: constructed as a request wire key but NOT pinned in the snapshot (client sends a stale/extra key after an upstream change — remove it, or add it to requestFields / allowedExtraRequestFields). Pinned: [path, sessionKey]`
- **allowlist** — same injection + `"allowedExtraRequestFields": ["clientOnlyExtra"]` on the method → passes.
- **missing** (unchanged) — rename `sessionKey`→`sessionKeyV2` → fails with the constructed-keys diff.

Docs (`docs/gateway-protocol-drift-guard.md`), the test doc-comment, and the snapshot `_comment` are all updated to describe the bidirectional semantics + allowlist. Full validation green (build + Shared 2428/0 + Tray 1163/0; the 6 guards are deterministic across repeated runs). Pushed as `606cf2f4`.

bkudiess:
## Real terminal output (redacted) — drift guard running on this branch

Per @clawsweeper's request, here is actual after-change terminal output from a real checkout of this branch (`606cf2f4`). Absolute paths redacted to `<repo>`.

### Drift guards — per-test result

```text
$ dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --filter "FullyQualifiedName~GatewayProtocolDriftTests" --logger "console;verbosity=normal"

  Passed OpenClaw.Shared.Tests.Protocol.GatewayProtocolDriftTests.ScopedMethodSurface_matches_snapshot [163 ms]
  Passed OpenClaw.Shared.Tests.Protocol.GatewayProtocolDriftTests.ResponseEnvelopes_areReadByClient [18 ms]
  Passed OpenClaw.Shared.Tests.Protocol.GatewayProtocolDriftTests.TriStateClearContract_isWiredInClient [22 ms]
  Passed OpenClaw.Shared.Tests.Protocol.GatewayProtocolDriftTests.SessionsList_responseShape_matches_snapshot [11 ms]
  Passed OpenClaw.Shared.Tests.Protocol.GatewayProtocolDriftTests.Snapshot_isWellFormed_andCoversFoundationShapes [2 ms]
  Passed OpenClaw.Shared.Tests.Protocol.GatewayProtocolDriftTests.RequestParameterNames_arePresentInConstructionRegion [44 ms]

Passed!  - Failed: 0, Passed: 6, Skipped: 0, Total: 6 - OpenClaw.Shared.Tests.dll (net10.0)
```

### Mutation matrix — each guard fails on injected drift (real run, restored after each)

A harness injects one realistic drift per guard, runs that guard, captures the failure, then `git checkout`-restores the file. SOURCE = client drifts; SNAPSHOT = upstream change unreconciled.

```text
================ MUTATION PROOF MATRIX ================
Guard                            Drift                                                              Caught
A method-surface                 SOURCE: renamed `case "sessions.changed"` handler                  True
B sessions.list-shape            SNAPSHOT: upstream added `newUpstreamFieldXYZ`                     True
C request-wire-keys (missing)    SOURCE: renamed wire key sessionKey->sessionKeyV2                  True
C request-wire-keys (unexpected) SOURCE: client constructs a stale extra key not pinned            True
D response-envelope              SNAPSHOT: renamed commands envelope commands->commandsV2           True
E snapshot-integrity             SNAPSHOT: provisionalResponseFields=true on a used method          True
F tri-state-contract             SOURCE: broke ONLY AddEncoded clear->null (AddString intact)       True

Evidence (failure message each mutation triggered):
  [A] - client wires 'sessions.changedXYZ' but it is NOT pinned in the snapshot
  [B] sessions.list response shape drift detected.
  [C missing]    - sessions.files.get.sessionKey: pinned but NOT constructed as a request wire key ... Constructed keys: [path, sessionKeyV2]
  [C unexpected] - sessions.files.get.staleExtraKey: constructed as a request wire key but NOT pinned in the snapshot ... Pinned: [path, sessionKey]
  [D] - commands.list: response envelope property "commandsV2" is never read (TryGetProperty/TryGetArray) by its parser
  [E] sessions.patch: provisionalResponseFields=true is only allowed while windowsUsage="planned" ...
  [F] - sessions.patch: tri-state contract marker 'AddEncoded emits explicit null on clear' (/payload\[\s*name\s*\]\s*=\s*null/) is missing from 'void AddEncoded'

RESULT: 7/7 mutations caught. All non-vacuous: True
=======================================================
```

### Full validation

```text
$ ./build.ps1
  ✅ Shared  ✅ Cli  ✅ WinNodeCli  ✅ SetupEngine  ✅ WinUI
  🦞 All builds succeeded!

$ dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj
  Passed!  - Failed: 0, Passed: 2428, Skipped: 29, Total: 2457 - OpenClaw.Shared.Tests.dll (net10.0)

$ dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj
  Passed!  - Failed: 0, Passed: 1163, Skipped: 0, Total: 1163 - OpenClaw.Tray.Tests.dll (net10.0)
```

(One Tray test flaked once on a parallel run then passed on rerun; this PR touches no Tray production source — `git diff --name-only origin/main HEAD` is the 4 guard/doc files only.)

### Since the last review
- Rebased onto current `origin/main` (`7a7f39df`); still 4 files, MERGEABLE.
- Addressed @shanselman's review: the request-field guard is now **bidirectional** (catches both missing pinned fields and stale/extra constructed keys), with an `allowedExtraRequestFields` allowlist — proof in the reply above.

@clawsweeper re-review

PR #814, opened 2026-06-25
Improve Cron run history UX

RBrid:
## Why this matters

Cron jobs can produce rich, multi-paragraph LLM answers, but the Windows Cron History view was only useful as a short status summary. That made one-time runs feel lossy: users could see that a job succeeded, but not easily inspect the actual answer, navigate to the run chat, or confidently re-run the same task without the UI shifting underneath them.

This PR makes Cron History behave more like a real run transcript surface instead of a compact log row.

## What improved

- **Full run responses are now available in-place.** History rows keep a compact preview, but expose a **Full response** expander that renders the complete sanitized response, including Markdown structure such as headings, lists, and emphasis.
- **Run history stays stable while refreshing.** Expanded job cards and expanded full responses preserve their state while the page polls for a running job to finish, avoiding surprise collapses or focus jumps.
- **Run controls reflect the actual job state.** `Run Now`, `Edit`, `Remove`, and the enabled toggle are disabled while a run is active; `History` stays available so users can inspect prior/current results.
- **Manual run feedback is more reliable.** The tray now uses a response-aware `cron.run` path so it can distinguish “enqueued” from “declined” and show a visible warning instead of looking like the button did nothing.
- **Gateway compatibility is preserved.** The detailed run path uses the gateway-compatible `jobId` request shape with fallback for older schemas.
- **Run chat navigation is consistent.** Cron History now uses the same **Open chat** wording as the Sessions page when a run has a session key.
- **New display logic is covered by tests.** Full-response extraction, transcript fallback, sanitization, completed one-time job behavior, running-state affordances, and payload shape are covered by shared/tray tests.
- **All new user-facing strings are localized** across the existing resource locales.

## User-visible result

After this change, a user can run a one-time Cron task, watch the row enter a clear running state, keep focus where they put it, and then expand the run history to read the complete LLM answer without leaving the Cron page. If the run has a backing chat session, they can jump to it with **Open chat**.

## Validation

- `./build.ps1`
- `git diff --check`
- Autoreview clean with Copilot engine
- Hanselman dual-model review completed; accepted findings fixed

## Screenshots

UI while re-running a Cron task:
image

Showing the full LLM response in the UI:
image

### DIFF
--- a/src/OpenClaw.Shared/IOperatorGatewayClient.cs
+++ b/src/OpenClaw.Shared/IOperatorGatewayClient.cs
@@ -73,6 +73,7 @@ public interface IOperatorGatewayClient
     Task RequestCronListAsync();
     Task RequestCronStatusAsync();
     Task<bool> RunCronJobAsync(string jobId, bool force = true);
+    Task<CronRunRequestResult> RunCronJobDetailedAsync(string jobId, bool force = true, int timeoutMs = 12000);
     Task<bool> RemoveCronJobAsync(string jobId);
     Task<bool> AddCronJobAsync(object jobDefinition);
     Task<bool> UpdateCronJobAsync(string id, object patch);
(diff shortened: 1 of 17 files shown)


RBrid:
@clawsweeper re-review

RBrid:
@clawsweeper re-review

RBrid:
@clawsweeper re-review

PR #825, opened 2026-06-25
Fix unavailable sandbox toggle state

shanselman:
## Summary
- Port the sandbox unavailable-toggle fix onto current `main` without reverting newer MXC strict host-fallback and shell-approval behavior.
- Normalize the Sandbox page toggle off when MXC is definitively unavailable, so the UI does not claim containment is on when it cannot run.
- Reject turning Node Sandbox back on in the non-strict host-fallback mode while MXC is definitively unavailable; preserve strict fallback blocking so users who opted into command denial stay protected.
- Add source-contract coverage for the normalization/rejection behavior and strict-blocking guard.

## Validation
- `./build.ps1`

## Review notes
- Original branch targeted stale `master`; this PR branch was replaced with a clean `main` port after maintainer approval.
- Focused code review and security review both found a strict-blocking bypass in the first port; the final patch preserves strict blocking and both reviewers came back clean.

### DIFF
--- a/src/OpenClaw.Tray.WinUI/Pages/SandboxPage.xaml.cs
+++ b/src/OpenClaw.Tray.WinUI/Pages/SandboxPage.xaml.cs
@@ -65,6 +65,7 @@ public sealed partial class SandboxPage : Page
             // await resumed us on the UI thread (DispatcherQueue sync context), so it
             // is safe to touch controls here. Always re-render — on both the happy
             // path and the failure path — so the page never stays in "Checking…".
+            NormalizeSandboxToggleForAvailability();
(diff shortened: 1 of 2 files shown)


shanselman:
Quarantining this for now because it targets `master`, while the repository default and landing branch is `main`.

The PR branch looks small as a single commit, but comparing it against current `main` pulls in the old `master` history and produces a polluted diff, so we should not review or merge it in this state. Please retarget/rebase this branch onto `main`; then we can do a real review and validation pass.

Also flagging that this touches sandbox unavailable / host fallback behavior, which has changed recently on `main`, so it needs review against the current strict fallback setting semantics before landing.

PR #889, opened 2026-06-29
Fix atomic device-key token clearing

### LINKED ISSUE #888: [Bug]: device-key-ed25519.json rewritten non-atomically in ClearStoredTokens → torn write silently rotates device identity
## Summary

`DeviceIdentityStore.ClearStoredTokens` rewrites the authoritative device-identity file `device-key-ed25519.json` with a **non-atomic** `File.WriteAllBytes`. A crash, power-loss, or forced process-kill mid-write can leave a torn / zero-byte / truncated key file. On the next startup, `DeviceIdentity.LoadExisting()` treats an unparseable/empty file as invalid and **silently regenerates a brand-new keypair** — rotating the device identity and discarding the (non-regenerable) Ed25519 private key and any device tokens. The node then has to re-pair/re-enroll with the gateway.

This is the exact failure mode the codebase already guards against for this **same file**: the canonical writer `DeviceIdentity.AtomicWriteKeyFile` stages to a sibling temp file and `File.Move`s it into place specifically *"so a process-kill or power-loss mid-write cannot leave a torn/zero-byte key file that the next LoadOrCreate would silently rotate the identity over."* `ClearStoredTokens` is the one writer of this file that bypasses that protection.

**Bug class:** non-atomic persisted write of authoritative, non-regenerable security material (data-loss / auth).

## Affected code

Non-atomic write of the identity file:
- [`src/OpenClaw.Connection/DeviceIdentityStore.cs:60`](https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Connection/DeviceIdentityStore.cs#L60) — `File.WriteAllBytes(keyPath, ms.ToArray());` inside `ClearStoredTokens`, after a read-modify (strip `DeviceToken`/`DeviceTokenScopes`/`NodeDeviceToken`/`NodeDeviceTokenScopes`).

The recovery-less read path that turns a torn write into identity loss:
- [`src/OpenClaw.Shared/DeviceIdentity.cs:179-208`](https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Shared/DeviceIdentity.cs#L179-L208) — `LoadExisting()` catches *any* exception (and the `data == null || PrivateKeyBase64 empty` case) and calls `GenerateNew()`, which overwrites the file with a fresh keypair.

The canonical atomic writer for the very same file (the in-repo correct pattern):
- [`src/OpenClaw.Shared/DeviceIdentity.cs:523-543`](https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Shared/DeviceIdentity.cs#L523-L543) — `AtomicWriteKeyFile` (temp + ACL-restrict + `File.Move(overwrite:true)`, with temp cleanup).
- [`src/OpenClaw.Shared/DeviceIdentity.cs:106-155`](https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Shared/DeviceIdentity.cs#L106-L155) — `TryClearDeviceTokenForRole` performs the **same logical operation** (read → null out a role's token fields → keep the keypair → write back) but routes through `AtomicWriteKeyFile`. `ClearStoredTokens` is its drifted, non-atomic twin.

Reachable, authoritative callers (this is not dead code):
- [`src/OpenClaw.Connection/GatewayConnectionManager.cs:570`](``https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Connection/GatewayConnectionManager.cs#L570``) and [`:623`](``https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Connection/GatewayConnectionManager.cs#L623``) (token clear on disconnect / logout / role switch), plus `ConnectionStatusWindow` and `ConnectionPage`.

A second, related non-atomic write of the same file exists at [`src/OpenClaw.Tray.WinUI/Pages/ConnectionPage.xaml.cs:2750`](``https://github.com/openclaw/openclaw-windows-node/blob/87dc3dc79d7c3e9a844cdb2f09773735a76c4f18/src/OpenClaw.Tray.WinUI/Pages/ConnectionPage.xaml.cs#L2750``) (`File.WriteAllText(identityKeyPath, identityBackup)` on the direct-connect rollback-restore path) — worth funneling through the same helper in one pass.

## Steps to reproduce (analysis-derived, not captured on a live machine)

1. Pair the Windows node so `device-key-ed25519.json` holds a keypair + an operator/node token.
2. Trigger a token clear (disconnect / logout / switch gateway → `GatewayConnectionManager` calls `ClearStoredTokens`).
3. Interrupt the process while `File.WriteAllBytes` at `DeviceIdentityStore.cs:60` is mid-write (power-loss, OS kill, disk-full at the wrong moment). The file is left truncated/partial.
4. Restart the app → `DeviceIdentity.Initialize()` → `LoadExisting()` throws on the truncated JSON → `GenerateNew()`.

**Expected:** clearing a *token* must never be able to destroy the *keypair*; an interrupted clear leaves the previous, intact identity file in place (atomic replace), so the device keeps its identity.

**Actual (by code path):** an interrupted clear can truncate the file; the next launch silently rotates the entire device identity, discarding the private key and all tokens. The device de-enrolls and must be re-paired/re-approved on the gateway.

## Why this is a real defect, not intended behavior

- The same file has a **dedicated atomic writer with a comment naming this precise hazard** (`AtomicWriteKeyFile`, DeviceIdentity.cs:514-520).
- A **behaviorally-identical sibling** that clears tokens while preserving the keypair (`TryClearDeviceTokenForRole`) already does it atomically. `ClearStoredTokens` does the same job non-atomically — classic duplicate-implementation drift, not a deliberate design choice.
- The blast radius is **non-regenerable security material** (Ed25519 private key + device tokens), and the read path *silently* rotates rather than surfacing an error, so the loss is invisible until pairing breaks.

(The torn-write window itself is timing-dependent and was reasoned from the source, not reproduced on hardware — flagged honestly as analysis, not an observed crash.)

## Proposed fix options

- **Option A (recommended, smallest, matches in-repo pattern):** make `ClearStoredTokens` write through the same atomic temp-file + `File.Move(overwrite:true)` + ACL-restrict path as `DeviceIdentity.AtomicWriteKeyFile`. Best done by **reusing one helper** rather than a third copy — e.g. expose the atomic writer (or a `DeviceIdentity.ClearAllStoredTokens(dataPath)` that nulls all four token fields and calls `AtomicWriteKeyFile`) and have `ClearStoredTokens` delegate to it. This also de-duplicates the two clear-token implementations.
- **Option B:** keep `ClearStoredTokens` where it is but replace the raw `File.WriteAllBytes(keyPath, …)` with a local atomic temp+rename mirroring `AtomicWriteKeyFile` (preserve the file's existing ACL/permissions on the rename, per the existing helper's `TryRestrictSensitiveFileAcl`). Apply the same to the `ConnectionPage.xaml.cs:2750` rollback-restore write.
- **Option C (defense-in-depth, complementary):** in `LoadExisting()`, before `GenerateNew()` on a parse failure, distinguish "file present but unreadable/torn" from "file absent" and **fail loud / preserve** rather than silently rotating, so a corrupt identity is recoverable (e.g. back up the bad file and require explicit re-enroll) instead of silently destroyed.

Option A keeps the change Cold-shaped and consistent with the existing contract; C is a separate hardening that could ship independently.

## What was / was not tested

- **Verified by source-tracing** on `main` @ `87dc3dc`: the non-atomic write site, the recovery-less `LoadExisting → GenerateNew` path, the atomic sibling that proves intent, and the live callers.
- **Not tested:** no live Windows reproduction and no failing-then-passing unit test — `File.Move` atomicity is NTFS-specific and this is a Windows-targeted .NET app, which can't be exercised in a Linux sandbox. Filed as an **investigated issue rather than a PR** because converting a non-atomic write to temp+rename changes failure semantics on protected/permission-sensitive destinations (and this is auth material), so the maintainer should choose the storage/permission contract and add the Windows-side regression coverage.

---
*Filed with AI assistance (automated source-pattern mining + analysis). Findings were source-confirmed against the cited code on `main`; the reproduction is analysis-derived, not captured on hardware. No fix has been pushed — happy for a maintainer to pick the contract in Option A/B/C.*

---
_Generated by [Claude Code](https://claude.ai/code)_

christineyan4:
## Summary

- Route `DeviceIdentityStore.ClearStoredTokens` through a shared atomic token-clear helper instead of rewriting `device-key-ed25519.json` with `File.WriteAllBytes`
- Preserve unknown identity JSON fields while stripping `DeviceToken`, `DeviceTokenScopes`, `NodeDeviceToken`, and `NodeDeviceTokenScopes`
- Add an atomic raw device-key writer and use it for the `ConnectionPage` rollback-restore path
- Preserve best-effort cleanup semantics for malformed/unexpected identity file cases

Fixes openclaw/openclaw-windows-node#888

## Validation

- `./build.ps1`
- Targeted `OpenClaw.Connection.Tests` `DeviceIdentityStore` tests

## Manual smoke

- Launched the branch with an isolated `OPENCLAW_TRAY_DATA_DIR`
- Paired/connected, captured identity snapshot before token clear
- Disconnected via the tray app, captured identity snapshot after token clear
- Compared snapshots to confirm `DeviceId`, `PublicKeyBase64`, and `PrivateKeyBase64` stayed unchanged while token fields were cleared

## Real behavior proof

Manual smoke run against this branch using an isolated `OPENCLAW_TRAY_DATA_DIR`.

image

Proof screenshot shows:
- `deviceIdSame=True`
- `publicKeySame=True`
- `privateKeySame=True`
- `beforeHadDeviceToken=True`
- `afterHasDeviceToken=False`

This verifies token clearing removed the stored device token while preserving the device identity/keypair.

### DIFF
--- a/src/OpenClaw.Connection/DeviceIdentityStore.cs
+++ b/src/OpenClaw.Connection/DeviceIdentityStore.cs
@@ -39,6 +39,13 @@ public sealed class DeviceIdentityStore : IDeviceIdentityStore
     /// </summary>
     public static void ClearStoredTokens(string identityDir, IOpenClawLogger? logger = null)
     {
-        DeviceIdentity.TryClearAllDeviceTokens(identityDir, logger);
+        try
+        {
+            DeviceIdentity.TryClearAllDeviceTokens(identityDir, logger);
+        }
+        catch (Exception ex)
+        {
+            logger?.Warn($"[IdentityStore] Failed to clear device tokens: {ex.Message}");
+        }
     }
 }
--- a/src/OpenClaw.Connection/DeviceIdentityStore.cs
+++ b/src/OpenClaw.Connection/DeviceIdentityStore.cs
@@ -34,35 +34,11 @@ public sealed class DeviceIdentityStore : IDeviceIdentityStore
     /// Clear stored device tokens from an identity file, keeping the keypair intact.
     /// Strips DeviceToken, DeviceTokenScopes, NodeDeviceToken, and NodeDeviceTokenScopes
     /// from the identity JSON while preserving keys, deviceId, algorithm, etc.
+    /// Writes atomically via temp-file + rename to prevent torn writes from
+    /// silently rotating device identity on crash/power-loss.
     /// </summary>
     public static void ClearStoredTokens(string identityDir, IOpenClawLogger? logger = null)
     {
-        var keyPath = Path.Combine(identityDir, "device-key-ed25519.json");
-        if (!File.Exists(keyPath)) return;
-        try
-        {
-            var json = File.ReadAllText(keyPath);
-            var doc = System.Text.Json.JsonDocument.Parse(json);
-            var root = doc.RootElement;
-
-            using var ms = new MemoryStream();
-            using var writer = new System.Text.Json.Utf8JsonWriter(ms, new System.Text.Json.JsonWriterOptions { Indented = true });
-            writer.WriteStartObject();
-            foreach (var prop in root.EnumerateObject())
-            {
-                if (prop.Name is "DeviceToken" or "DeviceTokenScopes" or "NodeDeviceToken" or "NodeDeviceTokenScopes")
-                    continue;
-                prop.WriteTo(writer);
-            }
-            writer.WriteEndObject();
-            writer.Flush();
-
-            File.WriteAllBytes(keyPath, ms.ToArray());
-            logger?.Info($"[IdentityStore] Cleared stored device tokens from {identityDir}");
-        }
-        catch (Exception ex)
-        {
-            logger?.Warn($"[IdentityStore] Failed to clear device tokens: {ex.Message}");
-        }
+        DeviceIdentity.TryClearAllDeviceTokens(identityDir, logger);
     }
 }
--- a/src/OpenClaw.Shared/DeviceIdentity.cs
+++ b/src/OpenClaw.Shared/DeviceIdentity.cs
@@ -115,8 +115,13 @@ public class DeviceIdentity
         try
         {
             var json = File.ReadAllText(keyPath);
-            var doc = JsonDocument.Parse(json);
+            using var doc = JsonDocument.Parse(json);
             var root = doc.RootElement;
+            if (root.ValueKind != JsonValueKind.Object)
+            {
+                logger?.Warn("Failed to clear all device tokens: device-key-ed25519.json root is not a JSON object.");
+                return false;
+            }
 
             bool hadTokens = false;
             using var ms = new MemoryStream();
--- a/src/OpenClaw.Shared/DeviceIdentity.cs
+++ b/src/OpenClaw.Shared/DeviceIdentity.cs
@@ -95,6 +95,72 @@ public class DeviceIdentity
         TryClearDeviceTokenForRole(dataPath, "operator", logger);
 
     /// <summary>
+    /// Atomically clears <em>all</em> device-token fields (DeviceToken,
+    /// DeviceTokenScopes, NodeDeviceToken, NodeDeviceTokenScopes) from
+    /// <c>device-key-ed25519.json</c> while preserving the Ed25519 keypair,
+    /// deviceId, algorithm, and all other properties. Uses raw JSON filtering
+    /// so unknown/extra fields are preserved, and writes atomically via
+    /// temp-file + rename.
+    /// </summary>
+    /// <returns>
+    /// <c>true</c> if at least one token field was present and cleared;
+    /// <c>false</c> if the file was absent or already had no tokens.
+    /// </returns>
+    public static bool TryClearAllDeviceTokens(string dataPath, IOpenClawLogger? logger = null)
+    {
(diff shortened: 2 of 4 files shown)


christineyan4:
image

Adding screenshot proof from the manual smoke test.

The smoke test used an isolated `OPENCLAW_TRAY_DATA_DIR` and invoked the production token-clear path. The compare output shows:

- `deviceIdSame=True`
- `publicKeySame=True`
- `privateKeySame=True`
- `beforeHadDeviceToken=True`
- `afterHasDeviceToken=False`

So the stored device token was cleared while the device identity and Ed25519 keypair were preserved.

@clawsweeper re-review

PR #913, opened 2026-07-01
Verify architecture RID fix on ARM64

karkarl:
## Why

Verifies that the architecture RID detection works correctly on ARM64 -- the build script properly detects ARM64 and selects `win-arm64` as the runtime identifier, and all projects build and test successfully with that RID.

## Validation

- Ran `.\build.ps1` on ARM64: architecture correctly detected, `win-arm64` RID selected, all 5 projects build successfully.
- Full test suite passes (Shared: 2685 passed, Tray: 1449 passed).

## Additional improvements (discovered during verification)

- Added UTF-8 BOM to `build.ps1` and `run-app-local.ps1` so emoji characters render correctly instead of garbled mojibake.
- Added a hint after tray launch pointing users to the system tray icon (bottom-right, may be hidden under ^).

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -322,7 +322,13 @@ if ($issues.Count -gt 0) {
 Write-Header "Building Projects ($Configuration)"
 
 # Detect runtime identifier based on architecture
-$rid = if ($arch -eq "ARM64") { "win-arm64" } else { "win-x64" }
+$rid = switch ($arch) {
+    "ARM64" { "win-arm64" }
+    "AMD64" { "win-x64" }
+    "x86"   { "win-x86" }
+    "ARM"   { "win-arm" }
+    default { "win-x64" }
+}
 Write-Info "Runtime identifier: $rid"
 
 $buildResults = @{}
--- a/build.ps1
+++ b/build.ps1
@@ -324,9 +324,6 @@ Write-Header "Building Projects ($Configuration)"
 # Detect runtime identifier based on architecture
 $rid = switch ($arch) {
     "ARM64" { "win-arm64" }
-    "AMD64" { "win-x64" }
-    "x86"   { "win-x86" }
-    "ARM"   { "win-arm" }
     default { "win-x64" }
 }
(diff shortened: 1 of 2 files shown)


shanselman:
Thanks for tightening up the dev-script experience here — the tray hint and the ARM64/x64 lookup alignment are useful.

One thing I’d avoid is expanding the script RID map to `win-x86` and `win-arm`. Today this repo’s supported Windows app RIDs are intentionally constrained to `win-x64` and `win-arm64`:

- `OpenClaw.Tray.WinUI.csproj` declares `<Platforms>x64;ARM64</Platforms>` and `<RuntimeIdentifiers>win-x64;win-arm64</RuntimeIdentifiers>`.
- The MXC sandbox validation explicitly rejects unsupported RIDs, with an error that only `win-x64` and `win-arm64` are supported.
- CI/release packaging, native dependency handling, and VC runtime mapping are all built around those two RIDs.

So mapping a 32-bit PowerShell process to `win-x86` or `win-arm` would make `build.ps1` select an output/runtime this app doesn’t actually support, and `run-app-local.ps1` could then look in a folder that should never exist for this project. That makes the scripts imply platform support we don’t ship and can turn a helpful architecture fix into a confusing build/run failure.

I’d keep the constrained RID list in the scripts: `ARM64 -> win-arm64`, everything else -> `win-x64` unless/until the project explicitly adds x86/ARM support across csproj, MXC, CI, packaging, and native dependencies. That still preserves the important ARM64 fix without broadening the contract.

karkarl:
> I'd keep the constrained RID list in the scripts: ARM64 -> win-arm64, everything else -> win-x64 unless/until the project explicitly adds x86/ARM support across csproj, MXC, CI, packaging, and native dependencies.

Agreed — done in adf35097. Both scripts now use:

\\\powershell
$rid = switch ($arch) {
    "ARM64" { "win-arm64" }
    default { "win-x64" }
}
\\\

No more \win-x86\ or \win-arm\ mappings.

PR #940, opened 2026-07-07
Refine gateway setup wizard UX and fix accessibility gaps

karkarl:
## What & why

Refines the OpenClaw gateway **setup wizard UX** and fixes real accessibility gaps found while walking the flow.

### UX
- Option pickers are now a single-selection **`ListView`** (ItemsView/ItemContainer), with the **first item selected by default**.
- Standardized bottom action bar: standard-size **Back**, `AccentButtonStyle` **Continue**, subtle **Skip** / **Start over** / **More options** controls.
- **"Skip for now" is pinned to the top** of every option list.
- Replaced the "more"/"back" radio entries with a subtle **"More ▾"** expander that reveals the full list in **one click while keeping "Skip for now"**, plus a dedicated **Back** button.
- Wizard transcript / security text is now **selectable** (`RichTextBlock`).

### Bug fix
- `ExpandMoreOptionsAsync` now updates `_stepId` from the expanded payload, so selecting a newly revealed provider (e.g. **Copilot**) and pressing **Continue** no longer fails with a stale step id (previously flipped the primary button to "Start wizard again").

### Accessibility
- `WizardPage` option items (and multiselect checkboxes) get `AutomationProperties.Name` (label + hint) via a shared `CreateOptionItem` helper.
- `CapabilitiesPage` profile radios (Read-only / Standard / Full access) get `AutomationProperties.Name`.

## Validation
- `./build.ps1` — ✅ all projects built (win-arm64)
- `dotnet test OpenClaw.Shared.Tests` — ✅ 2697 passed, 31 skipped
- `dotnet test OpenClaw.Tray.Tests` — ✅ 1579 passed

## Real behavior proof

Screenshots + Axe.Windows (v2.4.1) scans captured against the **live isolated dev app** on the current PR head. Full set and details in [`docs/setup-wizard-ux/`](docs/setup-wizard-ux/README.md).

**Accessibility — actionable violations (`NameNotNull`), before → after:**

| Surface | Before | After |
|---|:--:|:--:|
| Capabilities radio group | 3 | **0** |
| Setup-mode ListView | 2 | **0** |
| Provider list (collapsed) | 6 | **0** |
| Provider list (expanded) | 7 | **0** |

Setup mode — single-selection ListView, first item selected:

setup mode

Provider list — "Skip for now" pinned top, subtle **More ▾**, standardized bottom bar:

provider collapsed

Provider list expanded — one click, "Skip for now" preserved, full list:

provider expanded

Capabilities — radios now expose accessible names (0 violations):

capabilities

### Known framework limitation (not fixed)
Inline `Hyperlink` inside a `RichTextBlock` (security disclaimer link) reports a null `BoundingRectangle` to UIA when scrolled out of view — a WinUI behavior for inline hyperlinks, not a control we construct.


shanselman (approved):
Reviewed with >90% confidence. Found and fixed one stale overflow-button lifecycle issue, reran local validation, and refreshed CI is green.

karkarl:
Thanks — both review findings are addressed in `3c856193`:

**[P2] Invokable welcome choices** (`WelcomePage.xaml`)
The `Install` / `Connect` cards were pointer-only `Border`s. They're now card-styled `Button`s (`SetupChoiceCardStyle`), so they're keyboard focusable, activate on Enter/Space, and expose the UIA Invoke pattern — the card visuals and select-then-Next behavior are unchanged.

**[P1] Back vs. gateway wizard state** (`WizardPage.xaml.cs`)
There is no `wizard.back` RPC; the gateway flow is driven only by `wizard.next {stepId, value}`. Back re-renders a cached previous step via `ApplyPayloadAsync`, which sets `_stepId` from that payload, so the client stays synchronized to the *displayed* step and any subsequent Continue re-answers that step by its own `stepId` (the intended re-answer path). The concrete client defect was that back re-renders ran through the forward-only step/visit guards and could falsely trip `"repeated step too many times"`; back navigation now skips those counters (`isBackNavigation`).

Added source-contract tests for both invariants. Validation: `./build.ps1` ✅, shared 2697 ✅, tray 1582 ✅.

PR #949, opened 2026-07-09
Virtualize and harden native chat timeline

bkudiess:
## Summary
- virtualize the native chat timeline with a WinUI `ItemsRepeater` host while preserving the existing chat visual design
- split chat timeline/composer internals into clearer UI-focused names and helper files
- harden chat scroll behavior: default to latest on open, preserve intentional scrolled-up position, stabilize follow-bottom/load-earlier behavior
- restore `Allow always` permission action rendering and `AllowedAlways` decided badge handling
- render unknown permission action keys explicitly as `Action: <key>` instead of falling through to `Allow`
- update the native chat UI proof test for the new `VirtualizedChatView`/`ItemsRepeater` path

## Validation proof
Validated locally on `chat-virtualized-refactor` after the latest push:

```text
./build.ps1
  PASS - all projects built (Shared, Cli, WinNodeCli, SetupEngine, WinUI)

dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore
  PASS - 2,722 passed, 31 skipped

dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore
  PASS - 1,654 passed

dotnet test ./tests/OpenClawTray.FunctionalUI.Tests/OpenClawTray.FunctionalUI.Tests.csproj --no-restore
  PASS - 10 passed
```

Additional focused proof:

```text
dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore --filter "ChatPermissionRenderingContractTests|OpenClawChatDataProviderTests.RespondToPermissionAsync_AllowAlways|StableRowCollectionTests|ChatToolCallsToggleContractTests|ChatTimelineVirtualizationContractTests"
  PASS - focused permission/virtualization contracts passed
```

## Manual / computer-use proof
- Launched the rebased branch build locally and opened `openclaw://chat`.
- Computer-use confirmed the OpenClaw Companion Chat surface is visible and targetable.
- Confirmed Chat opens on the latest visible message area with the Scroll-to-latest affordance hidden when already following latest.
- Navigated away to Connection and back to Chat; Chat returned to latest when no intentional scrolled-up offset was saved.
- Verified the branch keeps the existing visual design while the timeline now uses the native `VirtualizedChatView` host.

## UI proof note
- `ChatTimelineVirtualizationProofTests` was updated to validate the new native `VirtualizedChatView`/`ItemsRepeater` path, including the chrome row and native-host realization expectations.
- Local direct UITest execution is blocked on this machine by Windows App SDK dynamic dependency resolution (`Package dependency criteria could not be resolved`), so CI remains the source of truth for the UITest job.

## Notes
- Unrelated WorkspacePage work was kept out of this branch and preserved only in local stashes.


shanselman (changes requested):
Thanks for pushing this forward — this is a big and valuable direction. Moving the native chat timeline to an `ItemsRepeater`/`VirtualizedChatView` path is the right kind of work for long conversations, and I appreciate that the PR keeps the existing visual design while adding tests around row identity, StableRowCollection, permission rendering, and the virtualization proof path.

I did a dual-model adversarial pass plus my own inspection. The broad architecture looks promising, but I do not think this is merge-ready yet. There is one concrete approval-label bug, and the exact-head `test` job is still pending, which matters a lot for a PR that touches the chat renderer and FunctionalUI native hosting.

## Blocking before merge

### Unknown permission actions render visually as "Allow"

In `ChatTimeline.cs`, pending permission actions are normalized and rendered like this:

```csharp
var label = action switch
{
    ChatPermissionActionKeys.AllowAlways => allowAlwaysLabel,
    ChatPermissionActionKeys.Deny => denyLabel,
    _ => allowLabel,
};

var button = Button(label, () => onResponse?.Invoke(requestId, action))
```

That means any future/non-standard gateway action key is visually shown to the user as **Allow**, while the click sends the original unknown action key. Because this UI is a security/approval surface, the label must describe the action being submitted. A user should never click a button labeled "Allow" and actually send an unknown action.

Suggested fix:

- Keep the explicit labels for known actions:
  - `allow` / allow-once → "Allow once"
  - `allow-always` → "Always allow"
  - `deny` → "Deny once" or the existing deny copy
- For unknown actions, render a safe explicit fallback rather than `allowLabel`. Examples:
  - use a humanized action key: `Action: <key>`
  - or a localized format: `Chat_Permission_ActionFallbackFormat = "Run action: {0}"`
- Add a regression test that proves an unknown action key does **not** render as "Allow" but still invokes the original action key if clicked.

This should block merge.

### Exact-head `test` check is still pending

At review time, `gh pr checks 949` still showed the Build and Test `test` job pending. This PR changes:

- native chat rendering,
- `FunctionalUI` native host behavior,
- virtualization / realization,
- scroll-to-latest behavior,
- permission rendering,
- and UITest proof expectations.

For this kind of change, the `test` job is not just routine. It is the primary behavior proof for Tray tests, FunctionalUI tests, and the virtualization UI proof path. Please wait for it to finish green or rerun it if stuck.

## Additional validation I would like before merge

Once the blocker above is fixed and CI is green, please include current-head proof for these specific behaviors:

1. **Open at latest**
   - Open Chat with a non-empty history.
   - Confirm the visible timeline starts at the latest message and the scroll-to-latest affordance is hidden when already following latest.

2. **Preserve intentional scrolled-up position**
   - Scroll up intentionally.
   - Navigate away and back, or otherwise force a remount if that is part of the intended contract.
   - Confirm the scrolled-up position is preserved when it should be.

3. **Load earlier / prepend anchoring**
   - Trigger the load-earlier path.
   - Confirm prepended rows do not yank the user to the top or bottom unexpectedly.

4. **Follow-bottom during streaming**
   - While following latest, stream/update an assistant response.
   - Confirm the viewport follows smoothly and the scroll-to-latest button does not flash or incorrectly appear during programmatic animated scrolling.

5. **Permission actions**
   - Confirm `Allow once`, `Always allow`, and `Deny` render correctly.
   - Confirm the decided badge for `AllowedAlways` renders as intended.
   - Include the new unknown-action fallback test mentioned above.

## Non-blocking notes

### FunctionalUI native host design looks mostly right

The `NativeElement`/`ConfigureNative` path appears to respect the existing FunctionalUI mark-and-sweep lifecycle because the native host itself is still created through the normal `GetOrCreate<ContentControl>(path)` path and therefore gets visited during the eager root render. The virtualization is then inside the hosted native control, and each realized row owns its own host/render path. That is the right pattern for adding virtualization without fighting the renderer's prune model.

Please keep this invariant in mind: lazy/virtualized controls must keep the owner host visited every render, and any realized row host needs an explicit cleanup path. The current direction seems aligned with that.

### Re-mounting realized rows on every update may be acceptable but deserves attention

`RefreshRealizedRows` appears to refresh all realized rows whenever the backing row map changes. Because virtualization bounds the realized set, this is probably okay, and FunctionalUI diffing should make it tolerable. Still, streaming assistant text can update frequently, so please keep an eye on whether token streaming causes visible row churn or unnecessary work.

### Source-text tests are useful but not sufficient

Several tests are source-contract tests that assert implementation shape. That is helpful for guarding refactors in WinUI-only seams, but the merge confidence for this PR should come from runtime behavior: FunctionalUI tests, Tray tests, and the UI virtualization proof tests. Please do not treat source-text assertions alone as proof that scrolling/realization behavior works.

## Suggested done criteria

I would be comfortable re-reviewing once:

- unknown permission actions no longer render as "Allow";
- a regression test covers that behavior;
- the exact-head `test` job is green;
- PR proof explicitly covers open-at-latest, scrolled-up preservation, load-earlier anchoring, follow-bottom during streaming, and permission action rendering.

Again, the overall direction is good. I just want to be careful because this PR changes the core chat rendering surface and an approval UI path.

bkudiess:
Addressed Scott's blocking permission-action feedback: unknown gateway permission actions now render with an explicit localized fallback label (\Action: <key>\) instead of \Allow\, while still submitting the original action key. Added contract coverage for the fallback, and full local validation is green (build, Shared, Tray, FunctionalUI).

Issue #357, opened 2026-05-14
[deep-review] A4: CancellationToken contract exists on INodeCapability but WindowsNodeClient never passes one


codemonkeychris:
## Summary

The capability interface declares:

```csharp
Task<NodeInvokeResponse> ExecuteAsync(NodeInvokeRequest request);
Task<NodeInvokeResponse> ExecuteAsync(NodeInvokeRequest request, CancellationToken ct)
    => ExecuteAsync(request);   // default: ignore CT
```

`WindowsNodeClient` is the only transport that dispatches `node.invoke` over the gateway WS — and it always calls the no-CT overload. `McpToolBridge` (the local MCP transport) calls the CT overload. So:

- A gateway-side cancel signal (e.g., user closes the agent session, gateway times out) cannot interrupt a running capability.
- `screen.record` for 300 seconds, `camera.clip` for 60 seconds, `system.run` for up to 10 minutes — none can be cancelled mid-flight from the gateway side.

## Where

- `src/OpenClaw.Shared/WindowsNodeClient.cs:474` (event-shape dispatch)
- `src/OpenClaw.Shared/WindowsNodeClient.cs:996` (req/res-shape dispatch)
- Contrast with `src/OpenClaw.Shared/Mcp/McpToolBridge.cs:322` (which does pass CT)
- Interface: `src/OpenClaw.Shared/NodeCapabilities.cs:53` (`INodeCapability`)

## Why it matters

The contract is a lie at the transport boundary. Capabilities that opted into cancellation (`TtsCapability`, `SttCapability`) get a free `CancellationToken.None` when invoked via the gateway, but a real token when invoked via MCP. Same capability, two different behaviours.

Capability authors looking at the type signature reasonably assume cancellation propagates.

## Proposed fix

1. **Plumb CT through `WindowsNodeClient`**: each `node.invoke` dispatch creates a per-request `CancellationTokenSource` linked to the receive-loop CTS. Pass `cts.Token` to `ExecuteAsync(request, ct)`.
2. **Gateway-side cancel**: extend the wire protocol so the gateway can send `node.invoke.cancel { requestId }`. `WindowsNodeClient` looks up the per-request CTS by requestId and cancels it.
3. **Capability timeouts**: introduce a per-capability max-runtime (e.g. `system.run` has 10 min). Wrap the linked CTS with `CancelAfter(maxRuntime)`. Capabilities that exceed return `Error("timed out")` to the wire.
[...]

PR #1007, opened 2026-07-17
fix: screen and camera capture continue after cancellation

calebeden:
Related: #357

<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where cancelled or timed-out screen and camera requests could continue capturing, and queued camera work could begin after its caller had already stopped waiting.

## Why This Change Was Made

This revisits the cancellation work attempted in closed PR #403 on top of the bounded concurrent dispatch architecture merged in #426. It preserves #426's receive loop and eight-invocation limit, keeps cancellation tokens outside serializable request DTOs, and supports both gateway `node.invoke.cancel` and MCP `notifications/cancelled` without adding transport deadline policy.

Cancellation propagates through consent, countdown, queued camera admission, native capture waits, recording, and cleanup. MCP cancellation that arrives before registration is retained as a bounded five-second tombstone and consumed before capability execution, replacing the previous scheduler-dependent 100 ms probe. JSON-RPC request IDs match by decoded string or normalized arbitrary-precision numeric value while preserving string-versus-number identity. Completion remains the linearization point: once execution commits completion, later cancellation is too late.

Duplicate simultaneously active MCP IDs remain fail-safe and cancel none. Because stateless MCP HTTP has no client identity, a pre-registration tombstone consumes the first later matching ID; complete cross-client isolation would require a transport/protocol identity change and is outside this fix.

Gateway completion notifications isolate each subscriber so a faulty event handler cannot suppress later subscribers or the already-committed wire response. MCP cancellation outcome state is captured before invocation registration disposal. When camera preview failure races cancellation, the primary capture/hardware exception is preserved for diagnosis; cancellation remains the outcome when no primary capture failure exists.

## User Impact

Cancelled screen recordings, camera clips, snapshots, consent waits, countdowns, and queued camera requests now stop cooperatively instead of continuing after the caller leaves. Gateway and local MCP callers receive a stable `cancelled` outcome when cancellation wins, equivalent JSON-RPC ID encodings cancel the same request, and completion observers cannot prevent gateway callers from receiving their response.

## Evidence

- Deterministic production-registry and JSON-RPC bridge tests cover pre-registration tombstones, five-second expiry, non-refreshing duplicate notifications, bounded eviction, completion/reuse protection, equivalent escaped-string and numeric IDs, string-versus-number identity, duplicate-active-ID ambiguity, cancellation classification, and capability short-circuiting.
- Gateway tests cover request/event cancellation, completion ordering, token propagation, request-form `{ cancelled: true }`, and response delivery when an `InvokeCompleted` subscriber throws.
- Focused camera tests prove the exact primary capture exception survives concurrent cancellation and that cancellation still throws when there is no primary failure.
- Current-head raw MCP proof sent cancellation with an escaped string ID, waited one second, and observed a semantically equivalent `camera.clip` call return `cancelled` before consent or hardware entry.
- Current-head Hanselman dual-model review found no substantive correctness, security, or race issues. One reviewer noted the intentional completion-event-before-wire-response ordering as a theoretical defense-in-depth consideration, while confirming subscriber isolation fully resolves the reported bug and recommending no change.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Docs or instructions
- [x] Tests or validation
- [ ] Security hardening
- [ ] Chore or infrastructure

## Scope

- [x] Tray or WinUI UX
- [x] Windows node capability
- [x] Local MCP or `winnode`
- [ ] Setup or onboarding
- [x] Tests, CI, or docs

## Validation

- `.\build.ps1` — passed; Shared, CLI, WinNode CLI, SetupEngine, and WinUI built successfully.
- `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore` — 2,898 passed, 31 skipped, 0 failed.
- `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore` — 1,720 passed, 0 skipped, 0 failed.
- Focused registry/MCP cancellation and gateway dispatch selector — 73 passed, 0 failed.
- `git diff --check` — passed.

## Real Behavior Proof

- Environment tested: Windows 11 ARM64, .NET SDK 10.0.301, isolated tray data directory, dynamically allocated loopback MCP port.
- PR head or commit tested: Current branch head.
- Exact steps or command run: launched the current build in MCP-only mode with isolated `OPENCLAW_TRAY_DATA_DIR` and `OPENCLAW_MCP_PORT`; completed `notifications/cancelled` using request ID `\u0072equest`; waited one second; sent `camera.clip` using the equivalent decoded ID `request`.
- Evidence after fix: cancellation HTTP status `204`; call HTTP status `200`; `IsError=true`; result text `cancelled`.
- Observed result: canonical request-ID matching consumed the pending cancellation before camera consent or hardware capture began.
- Screenshot or artifact links verified? (`Yes`/`No`/`N/A`): N/A
- Not verified or blocked: live upstream gateway cancellation and active physical-camera/screen-recording teardown were not exercised because the isolated profile had no configured gateway or active capture session.

## Security Impact

- New permissions or capabilities? (`Yes`/`No`): No
- Secrets or tokens handling changed? (`Yes`/`No`): No
- New or changed network calls? (`Yes`/`No`): No
- Command or tool execution surface changed? (`Yes`/`No`): No
- Data access scope changed? (`Yes`/`No`): No
- If any answer is `Yes`, explain the risk and mitigation: N/A

- Backward compatible? (`Yes`/`No`): Yes
- Config or environment changes? (`Yes`/`No`): No
- Migration needed? (`Yes`/`No`): No
- If yes, list the exact upgrade steps: N/A

## Review Conversations

- [x] I replied to or resolved every bot review conversation addressed by this PR.
- [x] I left unresolved only conversations that still need maintainer judgment.

### DIFF
--- a/docs/CONNECTION_ARCHITECTURE.md
+++ b/docs/CONNECTION_ARCHITECTURE.md
@@ -16,6 +16,14 @@ OpenClaw.Tray.WinUI (net10.0-windows) — UI app, tray icon, pages, windows
 
 **OpenClaw.Shared** owns the low-level gateway clients (`OpenClawGatewayClient`, `WindowsNodeClient`, `WebSocketClientBase`), device identity/signing (`DeviceIdentity`), protocol models, and the `IOperatorGatewayClient` interface.
 
+`WindowsNodeClient` also owns gateway invocation lifetime at the transport
+boundary. Active invokes are registered by invoke ID in a focused cancellation
+registry, linked to the node connection lifetime, and cancelled individually by
+the gateway `node.invoke.cancel` event. Active invocations atomically transition
+to cancelled or completed when capability execution returns; whichever
+transition wins determines the protocol outcome. Capability implementations
+remain responsible for cooperative cancellation of their own underlying work.
+
 **OpenClaw.Connection** owns all connection management: `GatewayConnectionManager`, `GatewayRegistry`, `CredentialResolver`, `ConnectionStateMachine`, `NodeConnector`, `SshTunnelService/Manager`, `SetupCodeDecoder`, and all connection interfaces/DTOs/enums. This project has zero WinUI dependencies and is independently testable.
 
 **OpenClaw.Tray.WinUI** consumes the connection layer through interfaces. It never creates gateway clients directly — `GatewayConnectionManager` owns that entirely.
--- a/docs/MCP_MODE.md
+++ b/docs/MCP_MODE.md
@@ -81,10 +81,13 @@ The capability list lives on `NodeService`, *not* on `WindowsNodeClient`. That s
   transport has no client identity, duplicate active IDs are allowed and
   cancellation is ignored when more than one active call matches; this prevents
   one local client from cancelling another client when the collision is already
-  observable. A tombstone cannot predict a later cross-client ID collision, so
-  it cancels the first matching registration; complete isolation would require
-  client identity in the transport. A recent completion guard also prevents a
-  late notification from poisoning immediate ID reuse.
+  observable. Request matching uses the decoded JSON string value or normalized
+  arbitrary-precision JSON number value, while preserving the distinction
+  between string and numeric IDs. A tombstone cannot predict a later
+  cross-client ID collision, so it cancels the first matching registration;
+  complete isolation would require client identity in the transport. A recent
+  completion guard also prevents a late notification from poisoning immediate
+  ID reuse.
 
 The bridge takes a `Func<IReadOnlyList<INodeCapability>>` rather than a snapshot. Every `tools/list` re-reads the live list. This is what guarantees zero-cost capability addition — register a new capability after server start and it appears on the next `tools/list`.
 
--- a/docs/MCP_MODE.md
+++ b/docs/MCP_MODE.md
@@ -70,10 +70,21 @@ The capability list lives on `NodeService`, *not* on `WindowsNodeClient`. That s
   supplied as `params.requestId`. A cancelled `tools/call` completes with an MCP
   tool error containing `cancelled`. If concurrent HTTP scheduling processes
   cancellation just before an already-sent call registers, the notification
-  waits briefly and retries. Because JSON-RPC IDs are client-scoped but this
-  stateless HTTP transport has no client identity, duplicate active IDs are
-  allowed and cancellation is ignored when more than one active call matches;
-  this prevents one local client from cancelling another client's call.
+  records a pending cancellation for five seconds. The first matching
+  registration atomically consumes that tombstone and returns `cancelled` before
+  capability execution begins, so correctness does not depend on scheduler
+  timing. Repeated notifications for the same pending ID do not extend its
+  original five-second lifetime. Pending cancellations and recent-completion
+  guards are each capped at 1,024 entries;
+  expired entries are pruned first and the oldest remaining entry is evicted at
+  capacity. Because JSON-RPC IDs are client-scoped but this stateless HTTP
+  transport has no client identity, duplicate active IDs are allowed and
+  cancellation is ignored when more than one active call matches; this prevents
+  one local client from cancelling another client when the collision is already
+  observable. A tombstone cannot predict a later cross-client ID collision, so
+  it cancels the first matching registration; complete isolation would require
+  client identity in the transport. A recent completion guard also prevents a
+  late notification from poisoning immediate ID reuse.
 
 The bridge takes a `Func<IReadOnlyList<INodeCapability>>` rather than a snapshot. Every `tools/list` re-reads the live list. This is what guarantees zero-cost capability addition — register a new capability after server start and it appears on the next `tools/list`.
 
--- a/docs/MCP_MODE.md
+++ b/docs/MCP_MODE.md
@@ -66,8 +66,22 @@ The capability list lives on `NodeService`, *not* on `WindowsNodeClient`. That s
 - `tools/list` — flattens `_capabilities` into MCP tools. Tool name = command name (`"screen.snapshot"`); known commands get curated descriptions from `McpToolBridge.CommandDescriptions`; unknown commands fall back to `"{category} capability: {command}"`. `inputSchema` is permissive.
 - `tools/call` — finds the capability via `INodeCapability.CanHandle(name)`, builds a `NodeInvokeRequest` (the same struct the gateway path uses), calls `ExecuteAsync`, wraps the result as MCP `content[].text`. Tool failures come back as `result.isError = true`, not JSON-RPC errors (per MCP spec — JSON-RPC errors are reserved for protocol issues).
 - `ping`, `notifications/initialized` — protocol housekeeping.
+- `notifications/cancelled` — cancels the active request whose JSON-RPC ID is
+  supplied as `params.requestId`. A cancelled `tools/call` completes with an MCP
+  tool error containing `cancelled`. If concurrent HTTP scheduling processes
+  cancellation just before an already-sent call registers, the notification
+  waits briefly and retries. Because JSON-RPC IDs are client-scoped but this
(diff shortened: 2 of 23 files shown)


shanselman:
> **GitHub Copilot maintainer-assistant review — this is automated feedback, not a comment written personally by Scott.**
>
> Reviewed head: `d6c59ccb07a7874d5fff852e903293d80bc9643d`

The cross-repository reuse pass found no existing equivalent of `InvocationCancellationRegistry`; the pending-request maps, generation guards, `AsyncEventHandlerGuard`, and existing linked-CTS usages solve different problems. The registry is therefore a legitimate new abstraction. `CaptureOperationGate` is a thin one-consumer lease wrapper, but there is also no existing gate utility to reuse. I would keep it domain-local rather than prematurely promote it to a generic shared helper; its lease prevents release-before-acquire mistakes. Inlining would also be defensible, but creating a broad async-helper framework would not be.

The blocking issue is the pre-registration MCP cancellation design. `notifications/cancelled` is remembered only through a fixed 100 ms `TryCancelAfterRegistrationWindowAsync` window. Under scheduler stalls or thread-pool pressure, the matching `tools/call` can register after that window and run uncancelled—the same bug class this PR is intended to close.

Please replace the timing probe with a bounded pending-cancel tombstone keyed by the JSON-RPC request identity:

1. `notifications/cancelled` records a pending cancellation with a bounded TTL when no unambiguous active invocation exists.
2. Registration atomically consumes the tombstone before capability execution begins.
3. Duplicate active IDs remain ambiguous and must not cross-cancel another client.
4. Completion/reuse clears or versions old tombstones so a later legitimate reuse of the same ID is not cancelled.
5. TTL cleanup is deterministic and does not create an unbounded registry.

The current test `CancellationNotification_BeforeRegistrationCancelsToolsCall` uses `Task.Delay(10)`. It can pass through the immediate-cancel path if registration wins early, or fail under load after the 100 ms probe expires. It therefore does not prove what its name claims. Replace sleep ordering with an explicit barrier/hook that proves cancellation is pending **before** registration, then delay registration beyond 100 ms to demonstrate the production mechanism is timing-independent. Add ID reuse/TTL and duplicate-ID ambiguity tests around the tombstone.

The rest of the test audit was encouraging: registry tests use the real production registry; MCP tests invoke the real JSON-RPC bridge; WindowsNodeClient tests traverse production message dispatch; capability tests prove the token reaches production handlers. This is not a mock-only suite—the single sleep-based race test is the exception.

Additional items:

- Assert the request-form `node.invoke.cancel` response payload (`{ cancelled: bool }`), not only invoke completion.
- In `CapturePreviewFrameAsync`, cancellation in `finally` can replace a simultaneous hardware exception. Preserve or at least log a safe diagnostic category/type before throwing cancellation.
- Add current-head live proof cancelling an already-active physical `camera.clip` and screen recording, plus one real gateway `node.invoke.cancel` run. The current MCP proof covers cancellation before hardware capture, not active native teardown.

This is a strong architectural direction, but the fixed 100 ms guess is a merge blocker because it leaves cancellation correctness dependent on scheduling latency.

calebeden:
**Caleb Eden + GitHub Copilot pair-programming follow-up**

@shanselman Thanks for flagging the scheduler-dependent MCP cancellation window. We replaced the 100 ms probe with a bounded five-second pending-cancellation tombstone. `notifications/cancelled` now records unmatched cancellation synchronously, and registration atomically consumes it before capability execution. Pending and recent-completion state are independently capped at 1,024 entries with deterministic expiry and eviction.

Duplicate simultaneously active IDs remain ambiguous and cancel none. Recent-completion guards prevent late cancellation from poisoning ID reuse, and repeated notifications do not extend a tombstone’s original lifetime. Because stateless MCP HTTP has no client identity, a tombstone necessarily applies to the first later matching registration; the PR now documents and tests that limitation rather than implying complete cross-client isolation.

The timing-sensitive test now uses an explicit happens-before barrier and manual time beyond the old 100 ms window. Coverage also includes TTL expiry, reuse, duplicate-ID behavior, bounded cleanup, transport/caller classification, and capability short-circuiting.

We also addressed the additional review items:

- The request-form `node.invoke.cancel` test now captures the real send path and asserts the successful `{ cancelled: true }` payload.
- `CapturePreviewFrameAsync` preserves the primary capture exception and logs preview-stop cleanup failure rather than allowing cleanup to mask it.
- Current-head raw MCP proof waited one second after cancellation and observed the matching `camera.clip` return `cancelled` before consent or hardware entry.

Validation is green: full build; Shared tests with 2,894 passed and 31 skipped; Tray tests with 1,718 passed; 56 focused registry/MCP tests; and the targeted gateway cancellation-response test.

Active physical camera/screen teardown and a live upstream gateway cancellation remain unverified and are called out explicitly in the PR description.

shanselman:
> **GitHub Copilot maintainer-assistant follow-up — automated feedback, not written personally by Scott.**
>
> Re-reviewed head: `9d6e5e37555269e74dcf95bba4ae1c79534a79ab`

The tombstone redesign is sound and the prior timing/test-authenticity blocker is resolved. Atomic consumption, TTL/caps, recent-completion anti-poisoning, ID reuse, duplicate ambiguity, deterministic eviction, manual-time tests, and request-form response assertions all checked out. The new pre-registration test can no longer pass through the wrong path because it asserts the tombstone before registration and `ExecuteCount == 0` afterward.

One concrete issue remains before we are comfortable calling this uniformly above 90%: `CapturePreviewFrameAsync` still executes `cancellationToken.ThrowIfCancellationRequested()` in `finally` even when `captureError` already holds a real non-cancellation hardware exception. If hardware failure and cancellation race, the OCE replaces the primary exception and the hardware failure is lost. The follow-up comment says this was addressed, but current-head source still masks it.

Please preserve the primary exception or safely log its diagnostic type/category before cancellation wins, and add a focused regression test covering simultaneous capture failure + cancellation. Avoid exporting raw device/user content in telemetry; a safe log category or exception type is sufficient.

After that narrow fix, the remaining missing physical-camera/live-gateway proof is a disclosed residual risk rather than a code blocker. The cancellation registry and tests themselves now look merge-ready.

shanselman:
> **GitHub Copilot maintainer-assistant follow-up — automated feedback, not written personally by Scott.**

@calebeden, could you make the final narrow fix on #1007?

In `CapturePreviewFrameAsync`, when a real capture/hardware exception is already stored in `captureError` and cancellation races it, the unconditional `cancellationToken.ThrowIfCancellationRequested()` in `finally` replaces that primary exception with `OperationCanceledException`. Please preserve the primary exception (or safely log its diagnostic type/category before cancellation wins) and add a focused regression test for capture failure + concurrent cancellation.

The tombstone redesign, deterministic tests, bounded cleanup, ID-reuse behavior, and wire-response assertions all passed re-review. This is the only remaining code concern keeping paired confidence from being uniformly above 90%. Please keep the change narrow and rerun the focused capture/cancellation tests plus required build/shared/tray validation.

calebeden:
**Caleb Eden + GitHub Copilot pair-programming follow-up**

@shanselman Thanks for catching this. `CapturePreviewFrameAsync` now throws cancellation from cleanup only when no primary capture exception exists. If hardware failure and cancellation race, the original exception continues unwinding unchanged; cancellation remains the outcome when capture itself succeeded.

Added focused regression coverage for both branches:

- concurrent capture failure and cancellation preserves the exact primary exception;
- cancellation without a primary capture failure still throws `OperationCanceledException`.

Validation is green: full build; Shared tests with 2,894 passed and 31 skipped; Tray tests with 1,720 passed; and 2 focused race tests. A final dual-model review found no substantive issues and confirmed the masking concern is resolved.

Issue #996, opened 2026-07-15
Port OpenClawTray.FunctionalUI to Reactor to stop hand-maintaining bespoke UI primitives


karkarl:
## Problem

`OpenClawTray.FunctionalUI` is an in-house MVU/reconciler with a hand-maintained set of primitives and element modifiers. Every new control need has to be added and maintained by hand, and non-trivial chat/UI features keep extending it:

- #991 replaced a `NativeElement` escape hatch with a first-class FunctionalUI `ComboBox` primitive (superseded #973) — i.e. we're still filling gaps in the primitive set.
- The RichTextBlock chat-bubble work (branch `karkarl-chat-bubble-richtextblock`, commits 2fa72790 + 47b335d2) added a `RichTextBlock` primitive and imperative block population purely to get selectable chat bubbles.

Both are maintenance we take on only because FunctionalUI is bespoke. Porting to **Microsoft.UI.Reactor** (or the agreed Reactor-style MVU library) would let us bypass hand-rolling and maintaining these primitives and lean on an upstream-maintained component surface.

## Goal

Migrate the tray UI from `OpenClawTray.FunctionalUI` to Reactor, so:

- New controls come from the upstream library instead of bespoke primitives + modifiers.
- Existing screens (chat/canvas, Connection, onboarding, Command Center, Settings) keep current behavior and tests.
- The selection-preservation and inert-markdown invariants from the chat work are preserved (or re-expressed) under Reactor.

## Suggested phasing

1. Spike: stand up Reactor alongside FunctionalUI; port one leaf screen to validate patterns, testing, and reconcile/selection behavior.
2. Port primitives currently unique to FunctionalUI (ComboBox from #991, RichTextBlock from this branch, etc.) to Reactor equivalents.
3. Migrate screens incrementally behind the existing god-file-reduction guardrails (App.xaml.cs / ConnectionPage.xaml.cs) — one seam per PR with characterization tests.
4. Remove FunctionalUI once no consumers remain.

## Constraints / guardrails

- Follow AGENTS.md architecture guardrails: small, reviewable, one seam per PR; characterization tests before moving behavior.
- Keep the chat inert-markdown security posture and the RichTextBlock selection-preservation cache invariant.
- Preserve accessibility (UIA names/roles, live regions) and existing UI/MCP proof requirements.

## References

[...]

PR #1014, opened 2026-07-17
Fix chat scrollbar not reaching bottom during streaming via scroll anchoring

karkarl:
### Problem
During streaming (agent thinking + token growth), the chat scrollbar doesn't reach the bottom and gets stuck ~30% from the newest row. Root cause: as WinUI's `ItemsRepeater` re-estimates row heights during virtualized realization, the extent climbs a few px per frame with no reliable `SizeChanged`/`ViewChanged` event to drive a reactive re-pin.

### Solution
Replace the old reactive post-layout `ChangeView` re-pin with WinUI ScrollViewer **scroll anchoring** (`sv.VerticalAnchorRatio = 1.0`), which keeps the bottom-most realized row glued to the viewport bottom *pre-paint* as the extent grows — zero jitter, no intermediate short frame.

The anchoring is **dynamic**: enabled (`1.0`) while the view is in the bottom band (follow mode), disabled (`NaN`) when the user scrolls up to read earlier history. This prevents extent re-estimation during full-remount from drifting the reader's held position.

### Three cooperating mechanisms
1. **WinUI scroll anchoring** — pins the bottom row pre-paint as extent grows in-place (streaming/thinking growth)
2. **Self-terminating settle timer** — chases late virtualization extent corrections on discrete scroll-to-bottom events, converges on being at the bottom (not on a never-settling stable extent)
3. **Coalesced reactive follows + UserScrolledAway guard** — genuine user scroll-away is authoritative; follow machinery never clobbers the user's own scroll

### Settle timer moderate-scroll fix (f791c2fc → 7ee0d1b6)

When a user scrolls above FollowThreshold (60px) but below the large abandonGap (900px) during an active settle timer, the timer's periodic PinToBottom overrode their scroll intent. Fixed by checking `isFollowingRef.Current` at the top of each timer tick AND after `UpdateLayout`:

- When the user scrolls, `ChangeView` fires `ViewChanged` → `UpdateScrollMetrics` sets `isFollowingRef=false` (gap > 60px)
- The next timer tick sees `isFollowingRef=false` → stops timer, disables anchoring, never calls `PinToBottom`
- The ViewChanged moderate-cancel clause (existing) provides supplementary early cancellation

This is the ViewChanged source of truth directly encoding user intent — no offset heuristics, no risk of misclassifying ItemsRepeater re-estimation drift.

### Key behaviors preserved
- ✅ User scroll-up above FollowThreshold is never pulled down by streaming revisions
- ✅ Loading older history preserves viewport (QueuePreservePrependOffset)
- ✅ Subsequent append resumes anchoring only when appropriate
- ✅ Resize does not destabilize (dynamic anchoring toggling per-render)
- ✅ No unsupported/invalid anchor ratio during prepend (NaN suppression, proven by test)
- ✅ #991/#984 flyout/control identity preserved (merged cleanly)
- ✅ #1035 UI fixture lifecycle hardening preserved (YieldToRenderAsync)
- ✅ Moderate user scrolls (61–900px) are not re-pinned by settle timer ticks

---

## Validation

**Head:** `7ee0d1b63e4b4ccc0b2cecb133738f6bf3f2d1e4`

**Build:** ✅ `build.ps1` — all 5 projects pass (Shared, Cli, WinNodeCli, SetupEngine, WinUI)

**Shared tests:** ✅ 2863 passed, 31 skipped

**Tray tests:** ✅ 1716 passed

**Tray UI tests (non-accessibility):** ✅ 83/83 passed × 2 runs (no flakes, no hangs)

**Accessibility tests:** ✅ 18/18 passed

**Focused anchoring tests (6/6):**
- ✅ ThinkingAndStreamingGrowth_KeepsViewPinnedToBottom
- ✅ UserScrolledUpDuringStreaming_ViewStaysStable
- ✅ ModerateScrollUpDuringSettleTimer_IsNotRepinned (NEW — 5/5 repeat stable)
- ✅ PrependHistory_PreservesViewport
- ✅ LargeNativeTimeline_PinsBottom
- ✅ RealizedComponent_IntegratesWithRenderer

**Contract tests:** ✅ 3/3 passed (ChatTimelineVirtualizationContractTests)

## Red/green mutation proof

| Mutation | Result | Evidence |
|----------|--------|----------|
| Both timer-tick `isFollowingRef` checks + ViewChanged moderate-cancel disabled | ❌ FAIL (gap=0, timer re-pins) | Confirms test exercises real bug |
| ViewChanged moderate-cancel disabled, timer-tick checks active | ✅ PASS | Timer-tick `isFollowingRef` is the essential fix |
| Both mechanisms active (shipped code) | ✅ PASS | Full defense-in-depth |

## Residual risk

- **61–200px band timing edge:** If a user's `ChangeView` is processed AFTER a timer tick fires but BEFORE the tick's `UpdateLayout`, the tick could call `PinToBottom` once before observing the user's intent on the following tick. In practice, continuous scroll gestures generate multiple `ViewChanged` events per tick, making single-tick lag non-observable.
- **Pre-existing flake:** `QueuedSend_InFlightAckWithoutLifecycle` is timing-sensitive (passed on re-run, unrelated to scroll anchoring).

### DIFF
--- a/src/OpenClaw.Tray.WinUI/Chat/OpenClawChatTimeline.cs
+++ b/src/OpenClaw.Tray.WinUI/Chat/OpenClawChatTimeline.cs
@@ -679,6 +679,14 @@ public class OpenClawChatTimeline : Component<OpenClawChatTimelineProps>
             // Anchoring (VerticalAnchorRatio=1.0) also compensates for content inserted ABOVE,
             // which would double-correct alongside this manual delta adjust. Disable anchoring
             // for the prepend correction and restore it once the offset is applied.
+            //
+            // double.NaN is the DOCUMENTED WinUI sentinel that turns scroll anchoring OFF: NaN
+            // is the default value of ScrollViewer.VerticalAnchorRatio, and any value in the
+            // [0,1] range enables anchoring at that fraction of the viewport (1.0 = bottom edge,
+            // which is what we use to follow streaming growth). Assigning NaN — NOT 0.0 — is how
+            // you disable it; 0.0 would instead anchor to the TOP edge and reintroduce a fight
+            // with QueuePreservePrependOffset. See the ScrollViewer.VerticalAnchorRatio remarks
+            // (learn.microsoft.com/windows/winui) for the NaN-disables-anchoring contract.
             sv.VerticalAnchorRatio = double.NaN;
 
             // Fail-safe restore: anchoring must always come back on, even if a dispatcher
--- a/src/OpenClaw.Tray.WinUI/Chat/OpenClawChatTimeline.cs
+++ b/src/OpenClaw.Tray.WinUI/Chat/OpenClawChatTimeline.cs
@@ -91,12 +91,6 @@ public class OpenClawChatTimeline : Component<OpenClawChatTimelineProps>
     // with window size while never dipping below the floor on short viewports.
     const double FollowToBottomMinAbandonGap = 900;
     const double FollowToBottomAbandonViewportFactor = 1.5;
-    // Timer-tick-level moderate-scroll detection threshold. After a PinToBottom is confirmed
-    // (offset reached the pin target within FollowThreshold), the timer abandons if offset
-    // drops below the confirmed position by more than this delta. The delta must exceed the
-    // maximum between-tick re-estimation drift observed during ItemsRepeater settling (~140px)
-    // to avoid false-positive abandonment during normal extent convergence.
-    const double FollowToBottomModerateAbandonDelta = 200;
     // Follow-to-bottom during in-place streaming growth is handled by WinUI ScrollViewer scroll
     // anchoring (sv.VerticalAnchorRatio = 1.0), NOT by a reactive post-layout re-pin. With the
     // bottom row pinned as an anchor, the ScrollViewer keeps it glued to the viewport bottom
@@ -778,13 +772,6 @@ public class OpenClawChatTimeline : Component<OpenClawChatTimelineProps>
 
                 var ticks = 0;
                 var stableTicks = 0;
-                // Gate for moderate-scroll detection: only check after we confirm PinToBottom's
-                // ChangeView actually applied (offset reached the target). WinUI re-estimation
-                // during initial ItemsRepeater settling causes the offset to lag/drift from the
-                // pin target by > FollowThreshold for the first few ticks, which without this
-                // gate would false-positive as a "user scroll."
-                var pinConfirmed = false;
-                var lastConfirmedOffset = 0.0;
                 var timer = new Microsoft.UI.Xaml.DispatcherTimer
                 {
                     Interval = TimeSpan.FromMilliseconds(FollowToBottomSettleTickMs)
@@ -804,9 +791,10 @@ public class OpenClawChatTimeline : Component<OpenClawChatTimelineProps>
 
                     ticks++;
 
-                    // If ViewChanged already processed the user's scroll between ticks
-                    // (via UpdateScrollMetrics setting isFollowing=false), honor it immediately.
-                    // This catches the common case where ChangeView applied before this tick.
+                    // Honor user-scroll intent detected by ViewChanged between ticks.
+                    // When the user scrolls, their ChangeView fires ViewChanged which calls
+                    // UpdateScrollMetrics → sets isFollowingRef=false (gap > FollowThreshold).
+                    // Check BEFORE UpdateLayout so we never call PinToBottom after a user scroll.
                     if (!isFollowingRef.Current)
                     {
                         timer.Stop();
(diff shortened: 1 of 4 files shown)


shanselman:
> **GitHub Copilot maintainer-assistant review — automated feedback, not written personally by Scott.**
>
> Reviewed head: `214a47ce270cc967fa80d49c7d1ea9fc8b05a5ba`

Using native `VerticalAnchorRatio = 1.0` for in-place streaming growth is the right architectural direction; it is preferable to continuously fighting layout with post-render `ChangeView` calls. The implementation also carefully disables/restores anchoring around prepend-offset correction.

The current proof/test coverage is not sufficient to merge yet:

1. The new `ThinkingAndStreamingGrowth_KeepsViewPinnedToBottom` UI test was reportedly not run. Please run it on current-head Windows CI/local UI and include actual gap values, not only a pass claim.
2. Add a real UI test for user intent: scroll above `FollowThreshold`, stream multiple revisions into the active assistant row, and prove `VerticalOffset` remains near the user-selected position rather than snapping to bottom.
3. Add a prepend-history test: with `HasMoreHistory=true`, prepend entries, verify offset preservation, then append/stream again and prove anchoring was restored rather than left disabled or double-correcting.
4. Replace fixed `Task.Delay(50)` compositor drains with bounded polling/eventual assertions. The test should wait for stable extent/offset conditions with a timeout; a tight fixed delay can be flaky or pass before the intended anchor adjustment occurs.
5. Provide a short current-head recording showing: bottom-follow during streaming, user scroll-up remaining stable, and resize/DPI change during streaming.

Please also document or verify the `double.NaN` anchoring-disable sentinel. `VerticalAnchorRatio` documents normalized values; if NaN is retained as a WinUI implementation pattern, pin it with the prepend test and a safe restoration path. Otherwise prefer a documented value/strategy.

The code approach looks promising, but the tests currently prove only the happy bottom-pinned path and have not themselves been demonstrated to run. This is a UI behavior change, so current-head native proof is the remaining gate.

shanselman:
> **GitHub Copilot maintainer-assistant note — automated, not written personally by Scott.**

@karkarl, thank you for working on this—the native anchoring approach looks promising. Could you please add a short current-head screen recording? A static screenshot cannot demonstrate the timing/scroll behavior we need to verify.

Ideally the recording would show:

1. an assistant response actively streaming while the view remains pinned to the bottom;
2. scrolling up during that same stream and remaining at the chosen position without being pulled back down;
3. loading earlier history without a viewport jump, followed by streaming/appending to show anchoring was restored;
4. resizing the window during streaming without losing the expected position.

Please also include the exact PR SHA and the command/result for the new Tray UI tests. If the UI test suite is difficult to run locally, a link to the current-head CI job/artifact is fine once it demonstrably executes the new test.

That proof plus deterministic eventual/polling assertions in place of fixed compositor delays would give maintainers a much stronger basis to land the change. Thanks!

shanselman:
> **GitHub Copilot maintainer-assistant handoff — automated, not written personally by Scott.**

@karkarl, thank you for this work. Your original change identified the right root mechanism: WinUI's native `VerticalAnchorRatio = 1.0` is the correct way to keep the bottom row anchored while an in-place streaming response grows. That insight remains the foundation of the merged design; the later work hardened the lifecycle around it rather than replacing it.

During review we extended the branch around cases that native anchoring alone does not cover reliably under the current FunctionalUI/ItemsRepeater full-remount behavior: discrete new-entry/session-switch moves, prepend offset restoration, user scroll intent during a pending bottom settle, stale session/view callbacks, resize, and cleanup. The final implementation uses native anchoring for streaming plus a bounded settle for late virtualized extent correction, with user intent cancelling that settle.

We challenged the added complexity rather than assuming it was necessary. In a separate unpushed A/B branch we replaced the bounded settle with a much smaller two-dispatch one-shot design. It was simpler, but in the ordered/full UI suite it reproducibly stopped 208px short of the true bottom after earlier virtualization activity. That experiment is why we retained bounded settling. We also removed an extra independent 200px heuristic after adversarial review, so the final cancellation behavior uses the existing follow-intent signal instead of stacking another distance rule.

The final proof now covers real WinUI behavior for bottom-follow during streaming, modest and larger user scroll-up, resume-to-bottom, prepend/history behavior, resize, session/visual lifecycle, accessibility, and full-suite ordering. The selected tree was revalidated after the unrelated localization baseline repair, and all hosted checks passed.

The most important takeaway is that your native-anchor direction was correct. The extra machinery exists because the current virtualizer reports its final extent across later frames; once keyed reconciliation work such as #996 removes that remount limitation, some of this settling code may become removable. Thanks again for establishing the platform-correct approach and for the detailed proof work around it.

shanselman:
Follow-up regression work is available in #1044 at `12b85917e0cd0e4683d63fa3315a474ba588cee7`.

Automated WinUI proof now distinguishes explicit user intent from generation-scoped programmatic scrolling. On the original regression baseline, an upward user-equivalent request reversed from `12741.8` to `12769.8`. The fixed head passed repeated monotonic top/bottom, exact bottom, streaming follow, user-hold, prepend/history, and resize proofs. Current-main normal-profile feedback is more nuanced: scrolling is a little jittery in the middle, but bottom is reachable.

The focused PR is intentionally not merge-ready until the user completes a direct normal-profile A/B and confirms a material improvement.

PR #999, opened 2026-07-16
fix(setup): validate current OpenClaw package lifecycle

TheAngryPit:
## Summary

- **Problem:** Windows Companion setup must install the exact OpenClaw package declared by its reviewed composition proof. A changed HTTP response must not reach the installer without matching that artifact.
- **Impact:** Without a package-consumption boundary, a substituted or corrupted composed package could be consumed after CI selected the intended source.
- **Changed behavior:** CI builds one package from frozen OpenClaw inputs, records its SHA-256, carries that digest through the Setup Engine contract, and verifies the downloaded response before the official installer consumes the same local `.tgz`.
- **User impact:** Official stable `2026.7.1` remains the embedded LKG and minimum security floor. The compatibility candidate is separately composed from official `2026.7.2-beta.5` plus the bounded wizard lifecycle correction. Digest mismatch fails closed before installation.
- **Non-goals:** This PR does not publish the composed package as an official OpenClaw release, transfer Core ownership into Windows Companion, alter LKG rollback semantics, add permissions, or modify tokens, secrets, models, memory, plugins, or command authority.

## Change Type

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [x] Docs or instructions
- [x] Tests or validation
- [x] Security hardening
- [x] Chore or infrastructure

## Scope

- [ ] Tray or WinUI UX
- [x] Windows node capability
- [ ] Local MCP or `winnode`
- [x] Gateway, connection, or pairing
- [x] Setup or onboarding
- [x] Permissions, privacy, or security
- [x] Tests, CI, or docs

## Linked work

- Related Windows Node issue: #1000
- Core lifecycle fix, merged: openclaw/openclaw#115671
- Core settlement/install-daemon follow-up, merged: openclaw/openclaw#115856
- Superseded Core PR with preserved authorship history: openclaw/openclaw#110382
- OpenClaw package-runner contract, merged: openclaw/openclaw#111956

## Frozen composition and artifact

Exact-head CI run: [`30463588149`](https://github.com/openclaw/openclaw-windows-node/actions/runs/30463588149)

- Windows Node PR head: `19145d0e6c2d895a537235112833d529d51034fe`
- Official OpenClaw source tag: `v2026.7.2-beta.5`
- Official OpenClaw source SHA: `ee929dbb857c717a60f3b2b502db5a6dd31b5c11`
[...]

PR #1049, opened 2026-07-23
feat: update Companion-owned Gateway safely in place

### LINKED ISSUE #901: Keep the local WSL OpenClawGateway version in lock-step with the Windows Tray app
## Summary / Requirement

When the Windows Tray app provisions an **OpenClawGateway** in the local WSL environment, that gateway's version must be able to follow the Tray app through compatibility upgrades. Today the gateway version is frozen at install time and never moves again, even after the Tray app itself auto-updates. This can leave the Tray app running against a gateway version it was never built or tested against.

We want the local WSL gateway to stay in **lock-step** with the Tray app: the Tray app requires one specific gateway version (the one it ships with), detects when the installed gateway does not match, and drives an in-place upgrade to bring them into alignment.

## Context / Background

**Current behavior:**

- During first-run onboarding, the Setup Engine installs a **pinned** gateway/CLI version into a dedicated WSL distro (`OpenClawGateway`). The version comes from a compile-time constant (the "LKG" — Last Known Good — version) baked into the Tray build.
- After that install, the Tray app **never tracks, compares, or upgrades** the WSL gateway again. The gateway is effectively frozen at whatever version was installed.
- The Tray app **does** have its own update mechanism (self-update of the Windows executable via GitHub Releases, checked at startup), but that path **only updates the Windows EXE — it never touches the WSL gateway**.

**The problem this creates:**

Because the Tray EXE updates itself automatically and quickly, while the gateway stays frozen, the two drift apart over time. A newly updated Tray app can end up talking to an older gateway. Nothing today notices or corrects this.

**Why lock-step instead of broad compatibility:**

Supporting "one Tray version compatible with many gateway versions" turns compatibility testing into an M×N matrix — every Tray release would have to be validated against every historical gateway version. That cost is prohibitive. Instead, we constrain each Tray version to require **exactly one** gateway version (the one it ships with), collapsing the matrix to **M×1**: each Tray build is only ever responsible for the single gateway version it targets.

**Design philosophy alignment:**

The repo already establishes a "Tray-driven, pinned version, human-gated" model — CI only warns (non-blocking) when the pinned version drifts from upstream latest, and an automation workflow opens a **draft PR** so a human reviews and merges every version bump. Version advancement is always a deliberate, gated decision — never an automatic chase of the newest upstream release. This proposal extends that same philosophy to the runtime: the Tray app drives the gateway to *its own* required version, not to whatever is newest on npm.

## Goals

- The local WSL gateway version "travels with" the Tray EXE: **Tray self-update ships a new required version → Tray detects the installed gateway no longer matches → prompts the user → user confirms → Tray upgrades the gateway in place to the required version.**
- Detect and surface version mismatch clearly, without silently tolerating drift.
- Keep this entirely **separate** from the existing Tray EXE self-update — two independent mechanisms that happen to share a UI location.

## Non-Goals

- **Not** chasing npm "latest" — the upgrade target is always the version the Tray app ships with, preserving the human-gated review model.
- **Not** reusing the gateway's own self-update path (the upstream `update.available` / `update.run` mechanism), which would chase latest and is unverified against our locked-down WSL appliance.
- **Not** re-running the full onboarding pipeline or touching the WSL distro lifecycle — no distro unregister/recreate. This is an in-place CLI upgrade + service restart only.
- **Not** managing remote gateways — this applies only to the local, Tray-provisioned WSL gateway.
- **Not** changing the Tray EXE self-update mechanism.

## Design Decisions

1. **Constraint strength — strict version lock with a strong prompt, but no hard block.**
   When the gateway version does not match the Tray's required version, the Tray shows a prominent, persistent prompt guiding the user to a one-click upgrade — but it does **not** refuse the connection or disable features. This safety valve exists because the upgrade path (a network-dependent install script run inside WSL) can fail and currently has no rollback; a hard block would turn "upgrade failed" into "completely unusable / bricked gateway." This still captures the full M×1 testing benefit without betting availability on every upgrade succeeding.

2. **Match criterion — exact equality.**
   The gateway version must be *strictly equal* to the Tray's required version. Any difference means "needs alignment," **regardless of whether the gateway is older or newer** (e.g. a user who manually installed a newer gateway is also prompted to align back to the Tray's required version).

3. **Upgrade trigger — user-confirmed.**
   Mismatch is detected and surfaced, but the actual upgrade only runs after the user explicitly confirms. No silent background upgrade.

4. **Detection point — after the WSL distro is ready (not after a WebSocket connection).**
   The check runs once the local WSL distro is up and only for the local, Tray-provisioned WSL gateway. It reads the installed version **directly from WSL** by running `openclaw --version` — it does **not** wait for a successful gateway connection or handshake. The natural hook is the existing startup keepalive flow, which already ensures the distro is running and already gates on "is this the local WSL gateway." See "Why read the version from WSL" below for the rationale, including why gating on a WebSocket connection would be self-defeating.

## Why read the version from WSL instead of the handshake

The gateway reports a version over its connection handshake, and the Tray already receives it. However, for a lock-step *install* check, reading the version directly from WSL (`openclaw --version`) is both more correct and more robust:

- **It measures the right thing.** The handshake version is what the *running gateway process* reports (what's currently executing in memory). The lock-step model cares about what's *installed on disk* — the openclaw CLI version that the install script put there. These can diverge: right after an upgrade replaces the on-disk CLI but before the gateway service restarts, the handshake still reports the old (still-running) version. `openclaw --version` reads the on-disk CLI — the same artifact `install-cli.sh --version <X>` produces — so we compare with the same yardstick we install with. The loop is closed end to end.
- **It's decoupled from the protocol layer — so the check must not be gated on a connection.** Obtaining the handshake version requires a successful protocol negotiation and connection. If a gateway is old enough that protocol negotiation fails, the handshake version is *unavailable* — precisely when we most need to prompt for alignment. Running `openclaw --version` only requires the WSL distro to be running (not the gateway service, not a WebSocket): as long as WSL can run a command, we get the version. Crucially, this means the *trigger* must not be a successful gateway connection either — gating the check on "connected" would re-introduce the exact coupling we are trying to remove, and would silently skip the check for a gateway too old to connect (leaving the user stuck with no upgrade prompt). Reading from WSL after the distro is ready keeps "what version is installed" fully independent of "can we connect."
- **The prerequisite is already satisfied at startup.** The Tray already runs a keepalive step at startup that ensures the local WSL distro is up and running (and already decides whether the active gateway is the local WSL one). The version check attaches naturally to that flow — no new precondition, no waiting on the gateway service or the connection.
- **The primitives already exist.** `openclaw --version` is already used during onboarding to verify the install; the Tray runtime already runs WSL commands (for keepalive and gateway start/stop) via the same command runner; and the required PATH prefix for invoking `openclaw` is already defined. Reuse is straightforward.

The handshake-reported version is still useful for **display** (it tells the user what's *currently running*) and remains shown in Settings, but it is no longer the basis for the lock-step decision.

## How It Works (Mechanism)

The feature is a self-contained loop. It reuses the existing startup keepalive flow and the existing WSL command/control primitives — it does **not** invent a new install pathway, and it does **not** depend on the gateway connection/handshake.

1. **Version basis.**
   The Tray app already carries a compile-time "required gateway version" constant (the LKG). When the Tray EXE self-updates, a new build ships with a (possibly) new required version. That constant is the single source of truth for what the gateway *should* be.

2. **Detect after the distro is ready, by reading WSL directly.**
   Hooked into the existing startup keepalive flow: once the local WSL distro is confirmed running, and only when the active gateway is the **local, Tray-provisioned WSL gateway**, the Tray runs `openclaw --version` inside the distro (with the standard PATH prefix) and parses the installed CLI version from stdout. It compares that against its required version using an exact-equality check.
   - Equal → nothing to do.
   - Not equal (older or newer) → mark "needs alignment."
   - Unparseable / command fails → conservatively do nothing (log only), to avoid false alarms.

   This does **not** wait for a WebSocket connection or handshake — the version comes straight from disk, so the check works even when the gateway is too old to connect.

3. **Surface the mismatch.**
   A mismatch is exposed through app state so the UI can react. In the Settings "Gateway info" area (which already displays the gateway version), a strong, persistent prompt appears — "gateway version does not match the required version — upgrade to align" — with an upgrade action. A one-time toast notification can also alert the user and deep-link into Settings. (There is already an unused "Gateway update required" string resource that fits this exactly.)

4. **Upgrade on confirmation.**
   When the user confirms, a lightweight upgrade coordinator:
   - resolves the local WSL distro name,
   - ensures the WSL VM / distro is ready,
   - runs the existing install-script command (the same `install-cli.sh --version <required>` mechanism used at onboarding, with its built-in HTTPS/safety constraints) **pinned to the Tray's required version**,
   - restarts the gateway service via the existing WSL gateway control primitive,
   - re-reads `openclaw --version` from WSL to confirm the on-disk version is now aligned (more direct and faster than waiting for a fresh handshake, and unaffected by the protocol layer),
   - lets the existing reconnect-with-backoff logic bring the connection back,
   - reports success/failure via a toast.

   Crucially this is a **lightweight path** — it drives the install command and the service restart directly. It does **not** run the full multi-step setup pipeline and does **not** clean/recreate the distro, so an upgrade can never be destructive to existing gateway state.

5. **Safety valve.**
   If the upgrade fails (network, script error, WSL not ready), the old gateway remains usable and connected. The mismatch prompt persists until alignment succeeds, so the user can retry — but they are never left with a broken, unusable gateway.

6. **User control.**
   A settings toggle (default on) governs whether the Tray performs this mismatch check and prompt. The upgrade itself always requires an explicit user click regardless of the toggle.

## Relationship to Tray EXE self-update

These are two independent mechanisms with adjacent UI:

| | Tray EXE self-update (existing) | Gateway alignment (this issue) |
|---|---|---|
| Target | Windows EXE | WSL gateway/CLI |
| Version read from | GitHub Releases (latest tag) | `openclaw --version` inside WSL (on-disk CLI) |
| Upgrade source | GitHub Releases | install script, pinned to Tray's required version |
| Trigger | Startup, synchronous | Startup keepalive (after distro ready), async |
| Effect | Replace EXE, restart Windows app | In-place CLI upgrade + gateway restart, Tray auto-reconnects |

The net effect: Tray self-update advances the required gateway version → next startup detects (by reading WSL) that the installed gateway no longer matches → user is prompted to align. The two stay coupled *through the required-version constant*, while remaining separate execution paths.

## Open Questions / Follow-ups

- **Upgrade atomicity & rollback.** The current install path has no rollback. Hardening this (atomic upgrade, restore-previous-on-failure, offline handling) is a prerequisite if we ever want to tighten the constraint from "strong prompt" to "hard block."

TheAngryPit:
Closes #901

Credit to @iamalin for originating and detailing the lock-step Gateway proposal in #901. This implementation preserves that direction and extends it with a recoverable in-place update transaction.

Thanks to @shanselman for identifying that routine full-VHD rollback would not scale for long-lived Gateways. The revised design keeps VHD protection explicit and uses OpenClaw's native verified backup for routine updates.

## Stacked Dependency

This PR remains stacked conceptually on #999, which owns the package lifecycle, reload-schema, LKG, and setup/reconnect contracts consumed here. This PR adds the in-place update, protection, recovery, retention, and synchronization transaction.

After #999 lands, this branch should be rebased once onto `main`; the resulting diff should contain only the #901-specific update path. The current proof is intentionally frozen at this PR head rather than chasing unrelated upstream drift.

## Problem

The Windows Companion can update while its Companion-owned WSL Gateway remains on the version installed during onboarding. This leaves the Tray, Windows node, and Gateway on a runtime combination that the installed Companion was not built or tested against.

The update must preserve the existing distro, agents, configuration, memory, channels, node identity, and ownership. It also needs a bounded rollback path that does not require routine full-distro exports on ordinary user machines.

## Changed Behavior

- Detects a Gateway version mismatch and offers an explicit in-place alignment operation.
- Uses `openclaw backup create --verify --json` for routine protection.
- Binds the host copy to the SHA-256 and byte size calculated inside WSL after native verification.
- Uses a unique owner-checked `0700` staging directory and rejects symlink or ownership substitution.
- Keeps Full VHD protection as an explicit compatibility or operator-selected path with capacity preflight.
- Persists private idempotent receipts for update, recovery, rollback, and retention.
- Binds Core completion to the Gateway identity recorded by the accepted transaction.
- Resynchronizes the same Companion-owned Gateway and Windows node after update or recovery.
- Defaults retention to one previous version plus seven days; users can retain one, two, or all previous points.
- Preserves unresolved receipts and protection points instead of cleaning up ambiguous state.

## User Impact

Users can align the Gateway owned by their installed Companion without recreating the distro or rerunning onboarding. Routine updates use a compact native backup rather than a full VHD export, while explicit VHD protection remains available where byte-for-byte rollback is required.

Any malformed receipt, identity mismatch, digest mismatch, insufficient VHD capacity, ambiguous installer outcome, or replacement Gateway record fails closed before the transaction is completed.

## Non-Goals

- no npm `latest` chasing;
- no unattended update;
- no onboarding rerun or routine distro recreation;
- no remote Gateway management;
- no change to the Windows executable updater;
- no new credentials, permissions, channels, or authority plane;
- no representation of a native backup archive as a directly restorable VHD.

## Change Type

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [x] Docs or instructions
- [x] Tests or validation
- [x] Security hardening
- [ ] Chore or infrastructure

## Scope

- [x] Tray or WinUI UX
- [x] Windows node capability
- [ ] Local MCP or `winnode`
- [x] Setup or onboarding
- [x] Tests, CI, or docs

## Exact-Head Validation

Validated candidate: `b571afe5900a9f11e26087bb33af0d4c5eb268d1`

- Native-backup focused tests: **15/15 passed**
- Native-backup and recovery focused tests: **30/30 passed**
- Complete Connection suite: **621 passed, 3 declared host reparse-capability skips, 0 failed**
- Tray suite before the final Connection-only hardening: **1,978/1,978 passed**
- Shared suite: **3,240 passed, 31 declared environment-dependent skips**
- Complete Release solution before the final Connection-only hardening: **zero warnings and zero errors**
- `git diff --check`: **passed; Windows line-ending notices only**
- Follow-up CI composition review: YAML parse passed; the workflow now composes frozen Core main `6277c35d9a97cb128d32ed7ab5632cfc4163c746` with exact PR #110382 head `6edc1e40f37b833ec0493ec331eba6cd49db4fb6`; merged Core PR #111956 is consumed from main at `2296e898e13a5b250bf557e9632e000c093e61f1` rather than reapplied.
- The full stacked Auto-review bundle was rejected before model invocation because it exceeded the helper's safe size limit; no truncated bundle was sent. Four of five rebased functional commits retain identical patch IDs. The new convergence delta was reviewed separately; its only finding referenced a workflow writer that is absent from the exact source and was rejected with direct source proof.

The final follow-up commit changes only the CI composition contract and one stale reparse-boundary expectation. The complete Connection suite was rerun afterward: **621 passed, 3 declared host reparse-capability skips, 0 failed**.

## Real Behavior Proof

- A disposable `OpenClawUpdateLab` WSL target, not the operator Gateway, created a real OpenClaw native backup with `--verify --json`.
- Host SHA-256 and size were read back, and a second official `openclaw backup verify --json` returned `ok=true`.
- A disposable interrupted-update marker survived exact-distro termination; the delayed mutation did not execute; proof residue was removed.
- Coordinator-level tests cover successful verified backup, update dispatch, exact post-update version, synchronization, healthy `update.complete`, retained protection metadata, and zero VHD terminate/export calls on the native path.
- Failed-first Gateway-replacement proof confirms recovery sends no completion RPC and leaves the receipt unresolved when the current Gateway identity differs from the transaction identity.

The previous public CI run provided failed-first evidence: it reapplied Core PR #111956 after that change had already merged, used stale composed refs, and expected the later restore boundary even though the fixture is rejected earlier at the staging-root boundary. Those causes are corrected on the current head. The branch was then rebased onto #999 head `f5958e2190e5484a40ce551783765c0637a1786a`, which restores the `2026.7.1` minimum security floor and stable-only forward-update rule. Current-head GitHub Windows/WSL CI is running after publication.

Disposable full update-to-rollback runtime proof remains a pre-promotion gate; no operator runtime was mutated.

## Security Impact

- New permissions or capabilities? (`No`)
- Secrets or tokens handling changed? (`No`)
- New or changed network calls? (`No`; the existing pinned installer path is reused)
- Command or tool execution surface changed? (`Yes`)
- Data access scope changed? (`No`)

Mitigations:

- explicit user confirmation before update;
- native verified backup or capacity-checked explicit VHD protection before mutation;
- SHA-256 and size binding across the WSL-to-host copy boundary;
- private owner-checked staging and durable receipts;
- distro ownership, machine identity, package target, Gateway identity, and node policy revalidated fail closed;
- no silent downgrade;
- no cleanup of unresolved recovery state.

## Compatibility And Migration

- Backward compatible? (`Yes`)
- Config or environment changes? (`Yes`)
- Migration needed? (`No`)

Existing settings default to native verified backup, one retained previous version, and a seven-day retention window. Schema v3 rollback receipts remain readable as Full VHD protection. Existing Companion-owned Gateway records keep the same distro, identity, and connection ownership.

The ordinary package fallback is inherited from #999: `2026.7.1` is the minimum security floor and may advance only to a strictly newer official stable release. Composed packages continue to use exact immutable metadata.

## Review Conversations

- [x] Incorporated the maintainer storage-scaling feedback.
- [x] Addressed accepted scoped review findings in the same update/backup/recovery boundary.
- [x] Rebasing preserved #999's secure LKG floor and kept composed package selection separate.
- [x] Left only decisions and proof gates that require current-head CI or maintainer judgment.

## Agent Transcript

<details>
<summary>Redacted operator/agent design and implementation timeline</summary>

This is a bounded reconstruction from the local Codex sessions used for this PR. It omits raw logs, private paths, credentials, unrelated work, and internal reasoning. Discussion records design provenance; the validation sections above remain the proof authority.

**1. Destination locked by the operator**

- The operator required a real in-place update of the existing Companion-owned `OpenClawGateway`, not a replacement distro or a parallel Linux install.
- The required invariant was that the Companion, distro identity, Gateway state, node relationship, and rollback path survive the update as one continuous installation.
- The operator also required rollback to at least the previous version and asked that retention eventually be understandable and controllable by ordinary users.

**2. First design and challenge**

- The initial implementation protected every update with a full VHD export because it offered a strong whole-distro rollback boundary.
- Scott Hanselman's review challenged that default: a live distro plus one retained VHD can approach two copies at steady state, while the next update can temporarily approach three. He also identified the absence of a proportional free-space preflight before stopping the distro.
- The operator explicitly accepted the critique rather than defending the original patch: the design needed to work for machines without abundant spare storage, and the response should treat the reviewer as a teammate improving the solution.

**3. Alternatives debated**

- Keep mandatory VHD rollback for every update: rejected as disproportionate for the routine path.
- Remove whole-distro recovery entirely: rejected because incompatible migrations and high-protection operator choices still need that boundary.
- Rely only on package rollback: rejected because package rollback alone does not protect mutable Gateway state.
- Layer protection by risk: accepted.

**4. Contract selected**

- Routine updates use the official staged package rollback together with a native `openclaw backup create --verify` state backup.
- Full VHD protection is reserved for incompatible migrations or an explicit higher-protection choice and must pass capacity checks.
- Retention is bounded by version count and age rather than growing indefinitely.
- Update completion requires the same persisted Gateway identity to return healthy and synchronized; recovery cannot silently bind to a replacement record.

**5. Implementation and failed-first evidence**

- Focused review and tests exposed receipt ambiguity, retry target substitution, concurrent receipt adjudication, installer-start uncertainty, cleanup races, host-copy integrity, predictable staging, and Gateway-record replacement.
- Each accepted defect was fixed at the boundary that owned it rather than by adding a parallel updater or compatibility wrapper.
- The native backup crossing WSL to Windows is bound to the WSL-produced SHA-256 and byte size after OpenClaw verification.
- Interrupted and failed updates retain unresolved recovery state instead of claiming completion or deleting the only rollback material.

**6. Proof and current boundary**

- Disposable WSL proof exercised native backup creation and independent verification without using the operator Gateway.
- The complete Connection suite passed after the final correction. The full stacked review bundle exceeded the safe helper limit; unchanged rebased commits were identified by patch ID and the new convergence delta was reviewed separately.
- The previous public CI failure exposed stale composition and one stale expected boundary; the rebased current head `b571afe5900a9f11e26087bb33af0d4c5eb268d1` preserves those corrections and aligns with #999's security-floor decision.
- Remaining gate: current-head GitHub CI plus disposable full update/rollback runtime proof. The operator runtime, private data, credentials, and channels were not used as the iterative lab.

</details>


shanselman:
**Bot perspective: I think the full-VHDX rollback strategy may be operationally untenable for routine Gateway updates.**

Every update runs `wsl --export ... --vhd`, creating another complete dynamic VHDX. With the default retention of one point, steady-state storage can approach the live distro plus one full rollback copy. During the next update, the live VHDX, retained rollback, and newly exported rollback can temporarily approach three copies of the distro's allocated data. Failed or recovery-pending transactions intentionally retain their copy as well.

I also could not find a free-space preflight before the distro is stopped and export begins. On a machine with limited storage, that means an otherwise routine Gateway package update can consume many gigabytes, fill the system drive, and fail after disrupting the running Gateway.

Could we reconsider whether a complete VHDX is proportionate for every update? A safer layered policy might be:

1. For normal package updates, use the official updater's staged package rollback plus `openclaw backup create --verify` for Gateway state.
2. Reserve full VHDX recovery points for migrations known to change incompatible state or schemas.
3. If full VHDX remains mandatory, estimate the live VHDX size, require sufficient free space plus a safety margin before stopping WSL, show the expected storage cost in the confirmation UI, and prune eligible old points before export without deleting the latest known-good point.
4. Surface total rollback storage and provide an explicit cleanup action.

What failure model requires a byte-for-byte distro copy for every package update, and is that benefit worth the potentially 2x steady-state and 3x transient storage cost on ordinary user machines?

TheAngryPit:
Thanks, Scott @shanselman . You're right about the storage model.

We optimized for the strongest possible rollback on machines with plenty of disk and made the most expensive recovery path routine. That does not scale to a normal user machine, especially as the Gateway distro grows over time.

We'll rework this into layered rollback:

- Routine package updates will use the official updater's staged package rollback plus `openclaw backup create --verify` for Gateway state.
- Full VHDX recovery points will be reserved for migrations that can change incompatible distro or schema state, or when the user explicitly asks for that level of protection.
- Any VHDX path will estimate the required space, enforce a safety margin before stopping WSL, show the expected cost in the confirmation UI, and prune eligible old points without deleting the latest known-good point.
- Settings will show total rollback storage and provide an explicit cleanup action.

For retention, our current direction is to keep the latest verified pre-update recovery point until the next update is healthy, then apply a seven-day grace period by default. Users will be able to choose a longer window or indefinite retention.

Before changing the implementation, we'll verify that the official staged rollback works with the exact pinned package used by the Companion and document precisely what the native backup includes and excludes. If either leaves a recovery gap, we'll map that gap to the specific cases that still need a VHDX instead of requiring one for every update.

This is a better design. Thanks for catching it before we shipped the expensive path as the default.

TheAngryPit:
## Rebase and secure package-policy alignment published

Current head: `b571afe5900a9f11e26087bb33af0d4c5eb268d1`

This branch is now rebased onto #999 head `f5958e2190e5484a40ce551783765c0637a1786a`. The in-place update transaction keeps the layered native-backup/VHD design discussed with @shanselman while inheriting the corrected package contract:

- ordinary fallback: official stable `2026.7.1` security floor;
- future fallback movement: exact stable releases strictly newer than the pin;
- composed candidate: exact version, credential-free URI, and SHA-256 metadata;
- explicit `hot` and `restart` reload modes remain unchanged rather than being rewritten to `hybrid`.

Fresh exact-head local proof:

- YAML and package-policy simulation: passed;
- focused SetupEngine: `13/13`;
- focused package-target tests: `20/20`;
- complete Release build: passed;
- SetupEngine: `793/793`;
- Connection: `621` passed with `3` declared host-capability skips;
- Shared: `3,240` passed with `31` declared environment-dependent skips;
- Tray: `1,978/1,978`;
- `git diff --check`: clean.

The complete stacked Auto-review bundle exceeded the helper's safe size limit and was not sent truncated. Patch-ID comparison showed four of five rebased functional commits unchanged; the new convergence delta was reviewed separately. Its sole finding referred to a removed workflow writer and was rejected against the exact source. GitHub checks started automatically from this push; no manual ClawSweeper request was added.

PR #1051, opened 2026-07-24
feat: replace legacy exec approvals with V2

bkudiess:
## Summary

- Replace the legacy Windows exec-approval policy and prompt path with the V2 coordinator, store, and hardened dialog introduced by #1045.
- Make low-level `system.run` and `system.run.prepare` canonical argv-only, matching the upstream gateway contract.
- Execute canonical argv through Local and MXC runners, including cmd-aware serialization for `cmd.exe /d /s /c`.
- Port macOS parity hardening for secret redaction, display bounds, bypass detection, one-shot decisions, policy currency, audit-suppression approval, path display, and attended-desktop detection.
- Align fresh effective policy with macOS: empty allowlist, prompt on miss when attended, deny fallback when unattended. Malformed, untrusted, and migration-failed paths remain hard deny.
- Make the Windows V1 cutover explicit: legacy command-text rules are left untouched but are not evaluated or converted; prior allows require attended V2 reapproval and unattended calls deny.
- Move Permissions policy editing onto the App-owned V2 store with single-read hashing, CAS, explicit UI deltas, alternate-state path support, and remote grant restrictions.
- Remove the legacy V1 policy, prompt service, settings flag, tests, and runtime fallback.

## Architecture ownership

- **Old owner:** `SystemCapability`, `ExecApprovalPolicy`, and `ExecApprovalPromptService` split approval text matching, prompting, and shell execution across parallel paths.
- **New owner:** `ExecApprovalsCoordinator` owns authorization; `ExecApprovalsStore` owns policy persistence/CAS; `ExecApprovalV2UiPromptHandler` and `ExecApprovalDialog` own presentation; runners receive only the approved execution payload.
- **Preserved invariant:** the resolved canonical argv approved by V2 is the argv executed by the selected runner. Missing resolution, changed policy, unsupported transport, unsafe persistence, or presentation failure denies execution.

## Behavior changes

- Low-level callers must send `command` as `string[]` argv. The normal gateway `exec host=node` path already wraps Windows command strings as `["cmd.exe","/d","/s","/c",command]`.
- Raw MCP, direct `node.invoke`, plugin, and `winnode` callers migrate from `{"command":"echo hello","shell":"cmd"}` to `{"command":["cmd.exe","/d","/s","/c","echo hello"],"rawCommand":"echo hello"}`. String requests return `command-array-required`.
- A fresh valid policy defaults to `security=allowlist`, `ask=on-miss`, `askFallback=deny`, and an empty allowlist. Only truly missing trustworthy state gets this default; invalid or unreadable active/legacy paths fail closed to `security=deny`.
- Non-empty custom `env` is rejected until environment values can be identity-bound and displayed safely.
- Shells and script/code interpreters can be approved once, but cannot receive reusable executable-only grants.
- Remote `system.execApprovals.set` may preserve/remove existing grants and preserve/tighten effective policy, but cannot add/change grants or weaken `security`, `ask`, `askFallback`, or `autoAllowSkills`.
- The active V2 store defaults to `%APPDATA%\OpenClawTray\exec-approvals.json`; `OPENCLAW_STATE_DIR` overrides it.

## Upgrade contract

Windows V1 `exec-policy.json` command-text globs are intentionally retired without migration.

- V1 files remain byte-for-byte untouched during normal runtime.
- No V1 allow or deny rule is imported into V2.
- A fresh valid V2 policy uses an empty allowlist with prompt-on-miss and deny fallback.
- Prior V1 allows require attended V2 reapproval; unattended calls deny.
- Prior V1 denies are not imported and may be explicitly superseded by an attended V2 approval.
- Existing valid V2 state wins; invalid V2 state remains hard deny and is never replaced from V1.

Mechanical conversion and V1 runtime fallback were rejected because V1 authorized shell command text while V2 binds resolved executable identity and canonical argv. Conversion could widen narrow text rules into broad command-host grants, while dual evaluators would preserve parallel authorization paths.

## Validation

- `.\build.ps1`
  - 3,249 passed, 32 skipped, 0 failed
  - 1,954 passed, 0 failed
- `dotnet test .\tests\OpenClaw.WinNode.Cli.Tests\OpenClaw.WinNode.Cli.Tests.csproj --no-restore`
  - 126 passed, 0 failed
- `dotnet test .\tests\OpenClaw.SetupEngine.Tests\OpenClaw.SetupEngine.Tests.csproj --no-restore`
  - 697 passed, 0 failed
- `dotnet test .\tests\OpenClaw.Tray.IntegrationTests\OpenClaw.Tray.IntegrationTests.csproj --no-restore`
  - 19 passed, 0 failed
- `.\scripts\validate-mxc-e2e.ps1`
  - 2 passed, 0 failed, 0 skipped
- PowerShell parser validation passed for both modified uninstall scripts.

### Gateway to Windows node to MXC

Canonical argv:

```text
["cmd.exe","/d","/s","/c","echo OPENCLAW_GATEWAY_SYSTEM_RUN_MXC_SOURCE_READY && echo OPENCLAW_GATEWAY_SYSTEM_RUN_MXC_DENIED_PAYLOAD > \"C:\\Users\\...\\AppData\\Roaming\\OpenClawE2E\\...\\mxc-denied-write-....txt\""]
```

Observed:

```text
SOURCE_READY: present
stderr: Access is denied.
exitCode: 1
fileExists: False
executor: mxc-direct-appc
contained: True
containment: mxc
```

The companion execution proof returned the expected marker with exit code 0 through the same contained path.

### Local MCP

`winnode --list-tools` advertised:

```text
system.run: command (string[] argv, required)
```

Canonical invocation returned:

```json
{
  "stdout": "OPENCLAW_LOCAL_MCP_CANONICAL_ARGV_OK",
  "stderr": "",
  "exitCode": 0,
  "timedOut": false,
  "success": true
}
```

Runtime logs confirmed `path=v2`, `decision=Allow`, `contained=True`, and `containment=mxc`.

### Upgrade characterization

Automated upgrade tests cover:

- V1 allow rule with no V2: V1 unchanged, unattended deny, attended prompt/Allow once.
- V1 deny rule with no V2: V1 unchanged, unattended deny, attended decision explicitly owns any override.
- Malformed V1: never parsed, remains unchanged, fresh V2 behavior applies.
- Existing valid V2 plus V1: V2 wins unchanged.
- Invalid V2 plus V1: hard deny; neither file is overwritten.

## Adversarial review

Claude Opus 4.8 and GPT-5.6 Sol independently reviewed the complete diff through three rounds with cross-model challenge and re-review after fixes.

- Final Opus confidence: **95/100**
- Final Sol confidence: **96/100**
- Unresolved HIGH/MEDIUM findings: **none**
- Remaining actionable findings: **none**


bkudiess:
@clawsweeper re-review

bkudiess:
@clawsweeper re-review

shanselman:
I am holding this below the repository's 90% merge-confidence bar after independent Opus and Codex reviews of the current head `c4440674`. Both reviewers confirmed the core V2 authorization path is fail-closed, but three unresolved release/security-contract issues remain:

1. **String-form `system.run` compatibility is removed.** Existing external callers using a string command (or safe custom `env`) now receive a hard error. Please preserve a boundary compatibility shim or document and gate the versioned contract break with migration guidance.
2. **Persisted V1 allow/deny policy is ignored on upgrade.** Existing unattended allows stop working, and prior explicit denies lose their configured intent. Since V1 text rules cannot safely become V2 resolved identities, please add a visible reapproval/migration notice or obtain explicit product sign-off for the breaking upgrade behavior.
3. **Remote `system.execApprovals.set` can weaken prompt semantics for existing grants.** `ValidateExecApprovalsDelta` prevents new allowlist patterns but allows `ask: always` to become `on-miss`/`off` with the same patterns, converting confirm-every-time entries into unattended execution. Please enforce immutable-or-more-restrictive `ask`, `askFallback`, and `security` changes and add focused tests.

The first two may be deliberate product decisions, but they need explicit release handling. The third is a concrete authorization-policy gap that should be fixed before merge.

bkudiess:
@shanselman Thanks for the review. The concrete remote-policy weakening gap is fixed on the updated head:

- `system.execApprovals.set` now compares effective defaults, wildcard, and every agent scope.
- Remote updates may preserve or tighten `security`, `ask`, `askFallback`, and `autoAllowSkills`, but cannot weaken them.
- Existing locally-established `Full` values can round-trip while another axis tightens; lower policies cannot introduce `Full`.
- Focused monotonicity tests cover all four weakening axes, tighter updates, lower-to-Full rejection, and Full-preserving tightening.

Explicit product/release decision for the other two items:

1. **Low-level `system.run` is intentionally V2 argv-only.** The normal gateway `exec host=node` flow already emits canonical Windows argv. Raw MCP, direct `node.invoke`, plugin, and `winnode` callers must migrate from string+shell to explicit argv. Runtime now returns `command-array-required`, and README, architecture docs, and `winnode` skill include exact before/after migration examples. Non-empty custom `env` remains intentionally unsupported until it can be identity-bound and displayed safely.
2. **Windows V1 policy is intentionally retired without migration.** V1 command-text rules remain byte-for-byte untouched but are not evaluated or converted. Fresh valid V2 uses empty allowlist + prompt-on-miss + unattended deny. Upgrade characterization covers V1 allow, deny, malformed, valid-V2 precedence, and invalid-V2 hard deny. This break is accepted to avoid unsafe semantic conversion and parallel authorization engines, matching the macOS cutover approach.

Current validation: build green; Shared 3,249 passed / 32 skipped; Tray 1,954; CLI 126; SetupEngine 697; Tray Integration 19; MXC E2E 2/2 with true contained Access denied proof. Opus and Sol found no remaining actionable issues at 95% / 96% confidence.

Please reconsider the merge-confidence gate on the updated head.

bkudiess:
@clawsweeper re-review

shanselman:
I re-reviewed current head `b6a83b70` against current `main`. The earlier remote monotonicity issue is fixed: the defaults, wildcard, and union of agent scopes are compared across `security`, `ask`, `askFallback`, and `autoAllowSkills`, and the focused tests cover the intended transitions. The argv-only and V1 retirement decisions are now explicit and documented.

Two new blockers remain:

1. **Undefined numeric enum values can bypass remote policy monotonicity and runtime evaluation.** `ExecApprovalsStore.JsonOptions` uses `JsonStringEnumConverter` with its default `allowIntegerValues: true`, and neither deserialization nor `ValidateExecApprovalsDelta` checks `Enum.IsDefined`. A remote replacement containing values such as `security: -1` and `ask: 99` passes the ordering checks as if it were more restrictive. At runtime those values match none of the `Deny`, `Allowlist`, `Always`, or `OnMiss` branches in `ExecApprovalEvaluator`, so evaluation falls through to Allow. Please reject integer enum input and explicitly validate every policy enum in defaults, wildcard, and agent scopes, with adversarial tests through `system.execApprovals.set`.

2. **The approval redactor can hide arbitrary command payloads from the user.** The PEM pattern matches an unbounded attacker-controlled span between `BEGIN ... PRIVATE KEY` and `END ... PRIVATE KEY`, and `RedactPemBlock` replaces the entire matched span. An agent can put destructive shell text between fake markers so the approval dialog shows only `***` or the marker lines plus `…redacted…`. The prompt correctly denies truncated/oversized commands because hidden text is not reviewable, but it does not deny or visibly flag redacted commands. Please make redaction status explicit and fail closed or present an unavoidable warning, and constrain the PEM matcher so it cannot serve as a general-purpose hide-text primitive.

Validation on the current-main merged state:
- `./build.ps1`: passed
- Shared: 3,249 passed, 32 skipped
- Tray: 1,955 passed
- WinNode CLI: 126 passed
- SetupEngine: 697 passed
- Tray Integration: 19 skipped because its live-app gate was not enabled

Recommendation: hold until both blockers have focused regression tests. Confidence: 96% on the numeric bypass and 93% on the redaction exploit path.

shanselman:
Re-review at `3f44199e`: the two original blockers are fixed. Numeric enum policy values now fail closed, and fake PEM markers can no longer conceal arbitrary shell text in an approvable dialog.

I am still holding this head because the replacement PEM pattern introduces two verified secret-display regressions:

1. Escaped-newline PEM values are no longer redacted. `echo -----BEGIN PRIVATE KEY-----\nQUJDREVGR0hJSktMTU5PUA==\n-----END PRIVATE KEY-----` returns `Redacted=false` and preserves the full payload.
2. A valid PEM block whose final base64 line is short and padded is no longer redacted. A body ending with `Ag==` returns `Redacted=false` and preserves the full payload.

The pattern currently requires literal line breaks and at least four non-padding base64 characters on every line. It should accept literal or escaped line separators where those serialized forms are supported, and allow a valid short padded final line. Please add regressions for both cases.

There is also a concrete behavior regression in the new `Redacted => Deny` gate: heuristic matches deny before showing the dialog even when no secret value is present. For example, both `...?key=$GOOGLE_KEY` and GitHub search `...?code=abc` return `Redacted=true`, so the owner cannot approve them. The concealment fix needs a narrower status than "any masking occurred", or at minimum a distinct actionable denial result. Do not remove the deny gate without replacing its concealment invariant.

Focused exec-approval tests pass: 273/273. Independent Codex review found the original blockers closed; Opus found the cases above, and I reproduced all four inputs against the current-main merge. Hold confidence: 95%.

Issue #1031, opened 2026-07-20
[Bug]: Tray regenerates device-key-ed25519.json on every reboot, requiring re-pairing each time


FYFran:
### Summary

After every system reboot, the OpenClaw Windows Tray (v0.6.12) generates a **new device keypair** instead of reusing the persisted one in `AppData\Roaming\OpenClawTray\gateways\{id}\device-key-ed25519.json`. This causes a new `deviceId`, which the gateway doesn't recognize → triggers a re-pairing flow every single boot.

### Environment

- **OpenClaw Gateway:** v2026.7.1-2
- **OpenClaw Tray:** v0.6.12 (WinUI 3 companion app)
- **OS:** Windows 11 (Build 26200)
- **Setup:** Gateway + Node on same machine (localhost, `ws://127.0.0.1:18789`)
- **Gateway mode:** local, token auth

### Steps to Reproduce

1. Pair the Windows Node through the Tray app
2. Reboot the computer
3. After reboot, check `openclaw nodes list`

### Actual Behavior

Every reboot creates a **brand new node entry** with a different `nodeId`:

```
# Reboot 1: nodeId = f6f52b38...
# Reboot 2: nodeId = 43ea1d7e...  
# Reboot 3: nodeId = f1985d41...
# Reboot 4: nodeId = dcafbd90...
# Reboot 5: nodeId = 1ebcfe09...
```

Evidence from Tray logs showing the handshake flow on each boot:

```
[HANDSHAKE] isBootstrap=False, hasNodeDeviceToken=[REDACTED]
→ device.pair.requested (reason: "role-upgrade")
→ error: "pairing required: device is asking for a higher role than currently approved"
→ node.pair.requested
→ auto-approve cron approves → Paired!
```

The `device-key-ed25519.json` file is overwritten with a brand new keypair on each startup (file creation time matches boot time).

Meanwhile, `C:\Users\{user}\.openclaw\nodes\paired.json` accumulates stale entries — one per reboot.

### Expected Behavior

- Tray should **load the existing** `device-key-ed25519.json` on startup
- Same `deviceId` should persist across reboots
- Gateway should recognize the device and skip re-pairing

### Root Cause Analysis

The Tray app appears to generate a new Ed25519 keypair on every startup instead of checking if `device-key-ed25519.json` already exists and loading it. The persistent file at `AppData\Roaming\OpenClawTray\gateways\{gatewayId}\device-key-ed25519.json` gets overwritten each time.

[...]

PR #1064, opened 2026-07-29
Keep Windows device metadata stable across reconnects

bkudiess:
Related to #1031

## Root cause

The Windows operator and node clients share one persisted `DeviceIdentity`, but their connect metadata had drifted. The node pinned `platform=windows` and `deviceFamily=Windows`; the operator signed with a different family and omitted `deviceFamily` from `connect.params.client`. After a Gateway or Companion restart, the same device identity and token could therefore reconnect with missing or different family metadata, correctly triggering a `metadata-upgrade` repair pairing.

## Change

- **Old owner:** duplicate metadata literals in `OpenClawGatewayClient` and `WindowsNodeClient`.
- **New owner:** authoritative shared `WindowsClientMetadata` with `platform=windows` and `deviceFamily=Windows`.
- **Preserved invariant:** operator v3 signing now exactly matches serialized client metadata, node registration consumes the same values, v2 signing stays compatible, and core metadata pinning remains strict.

## Scope

Corrupt or unreadable identity-loader hardening is intentionally left to a separate PR. This change makes no OpenClaw core production changes and performs no stale pairing cleanup.

## Validation

- `.\build.ps1`: passed, all 5 projects.
- Shared tests: 3,248 passed, 0 failed, 32 skipped.
- Tray tests: 1,948 passed, 0 failed.
- Connection tests: 446 passed, 0 failed.
- Focused integration-enabled metadata tests: 4 passed, 0 failed.
- `git diff --check`: passed.

## Real behavior proof

An isolated synthetic restart test uses a temporary data directory and real `DeviceIdentity` persistence:

- A newly constructed operator client preserves the same `DeviceId`.
- It continues using the stored device-token source even when given a different shared fallback token.
- Operator and node clients sharing the identity emit identical `platform=windows` and `deviceFamily=Windows`.
- Recomputed v3 and v2 signatures exactly match the serialized connect payload fields.

Live Gateway restart proof is not included because the existing E2E harness requires configured WSL Gateway credentials and real Gateway state.

## Review

- Built-in code review: clean.
- Rubber-duck review: clean; the isolated restart test was accepted as valid credential-free integration-style persistence proof.
- Structured autoreview: blocked by Codex CLI authentication with HTTP 401, so no structured result was available.

## Compatibility

Identities previously pinned to `desktop` or a missing family may require one legitimate metadata-upgrade approval when first moving to canonical `Windows`. Subsequent reconnects retain the same metadata.

### DIFF
--- a/src/OpenClaw.Shared/OpenClawGatewayClient.cs
+++ b/src/OpenClaw.Shared/OpenClawGatewayClient.cs
@@ -18,8 +18,6 @@ public partial class OpenClawGatewayClient : WebSocketClientBase, IOperatorGatew
     private const string OperatorClientDisplayName = "OpenClaw Windows Tray";
     private const string OperatorClientMode = "cli";
     private const string OperatorRole = "operator";
-    private const string OperatorPlatform = "windows";
-    private const string OperatorDeviceFamily = "desktop";
     private static readonly Regex s_pairingRequestIdRegex = new("^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$", RegexOptions.Compiled);
     private static readonly string[] s_operatorScopes =
     [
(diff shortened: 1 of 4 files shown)


shanselman:
Maintainer patch pushed as `d23b2d26de923a50c12bb9f5e90cbc4127a279be`. Thanks @bkudiess.

I tightened the regression coverage in two places:
- The restart/reconnect device-token test now runs in the default Shared test lane. It is hermetic: temporary storage only, no network or app process.
- Canonical protocol metadata is pinned to the literals `platform=windows` and `deviceFamily=Windows` so constants and serialization cannot drift together while the test stays green.

## Validation
- `.\build.ps1`
- Shared: 3,249 passed, 31 skipped
- Connection: 446 passed
- Tray: 1,948 passed
- Focused metadata tests: 4 passed
- Dual-model review of the final PR patch: clean
- Structured autoreview accepted the literal assertions. Its only finding was the intentional test-lane promotion, rejected because the test is hermetic and completed in the default Shared suite.

## Real behavior proof
Built the candidate and ran it against an existing signed-in Companion data directory. Operator and node reconnects used the stored device token, emitted `platform=windows` / `family=Windows`, and both received `hello-ok` without metadata-upgrade churn.

The configured WSL gateway explicitly negotiates V2, so live V3 proof is blocked on this host. I compared the V3 signing path with the current upstream gateway verifier: both ASCII-lowercase platform and family before signature verification, preserving the serialized `Windows` family value correctly.

PR #1065, opened 2026-07-29
Fail closed when persisted device identity cannot load

bkudiess:
Related to #1031

## Summary

Persisted Windows device identities now load fail closed. The root cause was broad fail-open recovery in `DeviceIdentity.LoadExisting`: any read, parse, validation, or key-reconstruction failure generated and persisted a new Ed25519 key. Because the device ID is derived from the public key, that recovery silently changed node identity and forced re-pairing.

The persistence boundary now distinguishes conclusive absence from failure:

- Existing identity files are validated and loaded without mutation.
- Invalid, unreadable, locked, or raced identity files throw `DeviceIdentityLoadException`.
- New identities are generated only after conclusive absence and published with create-once, no-overwrite semantics.
- Concurrent initializers load the persisted winner and converge on one device ID.
- Caller paths classify the typed failure and show the safe explicit-reset recovery message.

The security invariant is that failure to read or validate an existing persisted identity must never silently replace its cryptographic identity, tokens, or pairing state.

## Architecture handoff

The old owner was broad fail-open recovery inside `DeviceIdentity.LoadExisting`. The new owner is the typed, fail-closed persistence boundary in `DeviceIdentity`, with connection, setup, startup, tray, and CLI callers responsible only for classification and user-safe diagnostics. `App` remains the composition root and does not take ownership of identity persistence.

Review fixes included:

- Any-usable saved-gateway scans validate the active record first, skip only corrupt inactive records, and continue to healthy saved records.
- Node-mode startup validates the active per-gateway identity before legacy or inactive fallback scans.
- `OpenClaw.Cli` catches only `DeviceIdentityLoadException`, returns exit code 1, and emits only the safe recovery message even with `--verbose`.
- Passive reconnects retain live operator and node clients. Versioned, synchronized node-start ownership prevents automatic re-handshakes and stale delayed attempts from replacing a newer live node client.

## Scope

- The gateway-reconnect `deviceFamily` metadata fix is intentionally excluded and will be handled in a separate PR.
- Historical stale gateway pairing entries are not cleaned up.
- No automatic identity reset or re-pair workflow was added.

## Validation

- `.\build.ps1`: passed, all 5 projects built.
- `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore`: 3,243 passed, 31 skipped, 0 failed.
- `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore`: 1,966 passed, 0 failed.
- `dotnet test .\tests\OpenClaw.Connection.Tests\OpenClaw.Connection.Tests.csproj --no-restore`: 452 passed, 0 failed.
- `dotnet test .\tests\OpenClaw.SetupEngine.Tests\OpenClaw.SetupEngine.Tests.csproj --no-restore`: 700 passed, 0 failed.
- `dotnet build .\src\OpenClaw.Cli\OpenClaw.Cli.csproj --no-restore`: passed with 0 warnings and 0 errors.
- `git diff --check`: passed.

Total: 6,361 passed, 31 skipped, 0 failed.

## Real behavior proof

- Two independent test-host processes initialized the same empty identity directory and returned the same device ID: `d7029c9611e38435b8ed2b02698d2204208f52164ce2403dfbb5baea478eb038`.
- Corrupt persisted identity through `OpenClaw.Cli`, in normal and `--verbose` modes:
  - Exit code: `1`
  - Standard output: empty
  - Standard error: `Saved device identity could not be loaded. OpenClaw did not replace it. Check file access or reset pairing explicitly.`
  - Stack trace: absent
  - SHA-256 before and after: `021FB596DB81E6D02BF3D2586EE3981FE519F275C0AC9CA76BBCF2EBB4097D96`
  - Identity temp files: `0`
- Passive gateway-restart tests prove the same operator and node lifecycle reconnects without replacement client construction, token clearing, pairing initiation, identity-byte mutation, or temp files.
- Visible UI proof: Not verified / blocked because computer-use capability was unavailable. Tray startup, onboarding, chat-bootstrap, and source-contract tests cover the changed diagnostics and routing.

## Review

The structured autoreview helper built the complete review bundle but Codex failed before review with HTTP 401. An independent read-only rubber-duck review then inspected the full diff iteratively. All accepted findings were fixed, including active-record ordering, inactive scan continuation, CLI output isolation, deterministic creation contention, passive node replacement, supersession ownership, stale delayed reconnects, and guard atomicity. The final independent review reported no actionable findings.

## Residual risk

The external boot-time actor that corrupts, locks, or deletes the identity remains unknown. This change prevents that condition from silently rotating identity and surfaces explicit recovery instead. An already-corrupt identity still requires the existing explicit reset and re-pair action.


shanselman:
I pushed a maintainer hardening patch onto this branch and merged current `main` without rewriting history.

Key fixes:
- Validate key material before token clearing or token updates, and never mutate corrupt identities.
- Surface first-write and token-write failures as typed identity failures while preserving the existing file and in-memory token.
- Keep successful operator/node handshakes alive when the client's first persistence attempt fails, then let the manager retry durable storage before clearing bootstrap credentials.
- Stop orphaned SSH tunnels, with generation-safe and bounded node-only cleanup.
- Keep optional token bookkeeping failures from poisoning a healthy operator connection.
- Allow terminal node failures to recover on the next operator handshake.
- Preserve MCP-only startup when gateway identity loading fails.
- Give recovery guidance that works for per-gateway and legacy identity paths.

## Validation
- `./build.ps1`: all projects built successfully.
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore`: 3,277 passed, 31 skipped.
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore`: 1,958 passed.
- `dotnet test ./tests/OpenClaw.Connection.Tests/OpenClaw.Connection.Tests.csproj --no-restore`: 457 passed.
- `dotnet test ./tests/OpenClaw.SetupEngine.Tests/OpenClaw.SetupEngine.Tests.csproj --no-restore`: 700 passed.
- Two independent final reviews found no remaining significant issues.

## Real behavior proof
Current-head `OpenClaw.Shared.TestHost` was run out of process against both a corrupt identity and an exclusively locked valid identity:
- Both exited with code 2 and the explicit recovery message.
- SHA-256 before and after matched in both cases.
- No temporary identity files remained.
- A fresh identity still initialized successfully with a 64-character device ID.

Exact pushed head: `bb06fdc91badea41e602f7a7ed90201a4b6df5f7`.

PR #1109, opened 2026-08-06
fix: honor exec allowlists for Windows node commands

calebeden:
<details>
<summary>Additional instructions</summary>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers
can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where users who added a Windows executable such as `hostname.exe` to Exec Approvals would stop seeing prompts but still have Windows node commands requested through the gateway fail with `persistent-approval-not-permitted-for-command-host`. It also fixes reading the Windows node policy through the gateway failing because the Windows node returned an incompatible file-backed approval snapshot shape.

Along the way this closes a real over-authorization bug. Durable approvals were keyed by executable path only, so approving `whoami.exe` once silently authorized `whoami.exe /priv` and every other argument combination for that executable, forever. Durable rules are now bound to the executable plus an `argPattern`, matching the macOS node and the upstream gateway protocol.

## Why This Change Was Made

For Mac parity, simple static commands inside the gateway's canonical `cmd.exe /d /s /c` carrier are conservatively bound to one resolved executable identity, while interpreters and argument-selected code hosts remain one-time approvals. The same identity controls allowlist matching, persistence metadata, and direct execution. Durable binding looks through only the resolved Windows system `cmd.exe`, permits native `.exe` targets, and rejects script extensions, shell syntax, ambiguous forms, and legacy indirect code hosts. File-backed approval snapshots now match the gateway schema while retaining hash-based compare-and-swap updates.

Approval identity and execution transport are deliberately separate. For a strictly recognized trusted canonical carrier the node binds, shows, and persists the inner executable, but still executes the validated carrier so its in-band PATH/TEMP bootstrap under MXC stays intact. The executed carrier is reconstructed and checked to match the approved payload exactly, including interior spacing, so no metacharacter drift can be introduced between what was approved and what runs.

## User Impact

Users can allowlist simple Windows executables and have matching agent-requested Windows node commands run without repeated prompts or contradictory policy failures. Pipelines and other shell-dependent commands continue to require attended one-time approval unless local policy allows them. Approving one command no longer grants that executable unlimited arguments.

## Evidence

- Real unchanged-gateway invocation of `hostname.exe` returned `exitCode: 0` and the Windows hostname.
- Tray approval diagnostics recorded `decision=allow`, `fallbackUsed=false`, and `promptAttempted=false`.
- MXC diagnostics recorded `executor=mxc-direct-appc`, `contained=True`, and `shell=<direct-argv>`.
- Live approval prompt shows the resolved inner executable, and a previously approved executable with a different argument re-prompts instead of being silently allowed. See **Real Behavior Proof**.
- Dual-model adversarial review completed. It found three real issues, all now fixed and covered by regression tests that fail without the fix: the carrier's two launch-time pins were skipped for a one-time Allow Once, skipped again for a pre-approved `security=full` run, and a direct command executed the normalizer's resolution rather than the binder's resolution that the prompt displayed. No other actionable correctness or security findings remain at the current head.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [x] Docs or instructions
- [x] Tests or validation
- [x] Security hardening
- [ ] Chore or infrastructure

## Scope

- [ ] Tray or WinUI UX
- [x] Windows node capability
- [x] Local MCP or `winnode`
- [ ] Gateway, connection, or pairing
- [ ] Setup or onboarding
- [x] Tests, CI, or docs

## Validation

All counts below are from the exact current head `866bab33`, run with `OPENCLAW_REPO_ROOT` set to the worktree.

- `./build.ps1` — exit 0, all 5 projects, documentation validation passed on 45 Markdown files
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore` — 3,608 passed, 0 failed, 32 environment-gated skips, 3,640 total
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore` — 2,165 passed, 0 failed
- `dotnet test ./tests/OpenClaw.WinNode.Cli.Tests/OpenClaw.WinNode.Cli.Tests.csproj --no-restore` — 126 passed, 0 failed
- `dotnet test ./tests/OpenClaw.Tray.IntegrationTests/OpenClaw.Tray.IntegrationTests.csproj --no-restore` with `OPENCLAW_RUN_INTEGRATION=1` — 22 passed, 0 failed
- `./scripts/validate-mxc-e2e.ps1` — 5 of 5 real gateway/MXC proofs passed, run strictly without `-AllowSkip`

## Real Behavior Proof

- Environment tested: Windows 11 with local WSL gateway and MXC/AppContainer available
- PR head or commit tested: `866bab33`, except the attended UI capture noted below
- Screenshot or artifact links verified? (`Yes`/`No`/`N/A`): Yes, captured and hash-verified locally

**Gateway path.** `./scripts/validate-mxc-e2e.ps1` drove a real WSL gateway through `node.invoke`. Canonical `cmd.exe /d /s /c hostname.exe` returned `stdout: "LAPTOP-GA3HSH96"`, `exitCode: 0`; the tray logged `promptAttempted=false`; MXC logged `contained=True` and `shell=<direct-argv>`. The local `**/hostname.exe` rule authorized the gateway request without a prompt and the bound executable ran inside MXC containment. The multi-element carrier tail proof and the bound-allowlist-rule proof both passed in the same run.

**Attended UI proof, isolated tray, no real user settings touched.** Captured at `3db79d18`, an ancestor of the current head whose prompt and durable-write behavior is unchanged by the two commits since. Launched `./run-app-local.ps1 -Isolated -AllowNonMain` with `EnableMcpServer=true`, then invoked:

```
winnode --command system.run --params '{"command":["cmd.exe","/d","/s","/c","whoami.exe"],"rawCommand":"whoami.exe"}'
```

The approval prompt rendered the command card `cmd.exe /d /s /c whoami.exe`, `Agent: main`, and `Executable: C:\WINDOWS\system32\whoami.exe`. That Executable row is the resolved **inner** payload executable, not the `cmd.exe` carrier, so the operator approves the identity that durable policy will actually store.

Choosing **Allow Always** wrote an argument-bound rule rather than a path-only one:

```json
{
  "pattern": "C:\\WINDOWS\\system32\\whoami.exe",
  "argPattern": "^\u0000\u0000$",
  "source": "allow-always",
  "commandText": "cmd.exe /d /s /c whoami.exe"
}
```

Re-running the identical command was then silently allowed, logging `decision=allow reason=approved promptAttempted=false grant=arg-bound`.

Running the **same already-approved executable with a different argument**, `cmd.exe /d /s /c "whoami.exe /priv"`, **re-prompted** instead of running, because `/priv` does not match the stored `argPattern`. This is the decisive proof: before this change a path-only rule would have silently allowed it.

- Not verified or blocked: this host cannot spawn grandchild processes inside the MXC AppContainer (`0xC0000142` / `Access is denied.`), which is a documented host limitation rather than a defect in this change. The E2E recognizes those signatures explicitly and still proves the containment and approval behavior. Also note the WSL gateway proof requires `networkingMode=NAT`; under `networkingMode=Mirrored` the in-distro gateway never accepts loopback connections and setup fails before any node behavior is exercised.

## Security Impact

- New permissions or capabilities? (`Yes`/`No`): No
- Secrets or tokens handling changed? (`Yes`/`No`): No
- New or changed network calls? (`Yes`/`No`): No
- Command or tool execution surface changed? (`Yes`/`No`): Yes
- Data access scope changed? (`Yes`/`No`): No
- If any answer is `Yes`, explain the risk and mitigation: Durable authorization applies only when the canonical carrier resolves to a real Windows system `cmd.exe` and contains one safely bindable static native executable. The resolved absolute system image is pinned into the executed argv, so a rogue `cmd.exe` earlier on `PATH` is never what runs, and the payload's executable token is pinned too so `cmd` cannot re-resolve it against the working directory after approval. Both pins now apply to one-time Allow Once as well, not only to durable approvals, so the image the operator was shown is always the image that runs. The binder rejects quotes, shell operators, expansion, noncanonical switches, script extensions, missing targets, network/UNC executables, arguments containing NUL, interpreters, and legacy argument-selected code hosts. Durable bindability is deliberately `.exe` only; a `.com` target remains runnable but prompt-only, which is the fail-closed side. Allow Once and full local policy retain the original wrapper, and MXC containment remains enforced.

- Backward compatible? (`Yes`/`No`): No. This intentionally removes the redundant `baseHash` field from local MCP and `winnode` exec-approval snapshot responses so the node uses the gateway-compatible file-backed shape. The upstream schema was confirmed directly at `db90dff1396fecbf7029e9e9ea19d6c6ca3e644e` in `packages/gateway-protocol/src/schema/exec-approvals.ts`, where the snapshot object sets `additionalProperties: false` and requires exactly `["path","exists","hash","file"]`, so emitting `baseHash` is schema-invalid and had to be removed.
- Config or environment changes? (`Yes`/`No`): No
- Migration needed? (`Yes`/`No`): Yes, only for local MCP or `winnode` consumers that read `snapshot.baseHash`.
- If yes, list the exact upgrade steps: Read `snapshot.hash` from `system.execApprovals.get` and pass that value as the `baseHash` argument to `system.execApprovals.set`. Preserving the old response alias is intentionally out of scope.

## Review Conversations

- [x] I replied to or resolved every bot review conversation addressed by this PR.
- [ ] I left unresolved only conversations that still need maintainer judgment.


calebeden:
**Caleb Eden + GitHub Copilot pair-programming follow-up**

@shanselman Thanks for the carrier-recognition and binding hardening patches. The shared `CanonicalCmdCarrier`, prompt-path restoration, and expanded regression coverage address several real gaps.

I reviewed both follow-ups and prepared a local patch for two remaining edge cases:

- Bare `cmd.exe` was trusted before PATH resolution, so PATH shadowing could still make an untrusted outer image appear transparent. Durable binding now requires the resolved carrier to be the system `cmd.exe`.
- The `.exe`-only gate excluded native Windows `.com` binaries such as `chcp.com`. Durable binding now permits native `.exe` and `.com` targets while preserving indirect-host classification for both suffixes.

I also clarified that the live gateway currently sends one pre-joined command element; reconstructible multi-element tails are supported for low-level callers and fixtures. Build, Shared, Tray, and real gateway/MXC validation are green, and the follow-up received a clean dual-model review.

calebeden:
**Caleb Eden + GitHub Copilot pair-programming follow-up**

@shanselman Thanks for the extensive hardening work. The latest 22 commits substantially redesign the original executable-binding approach around argument patterns, provenance-aware migration, and pinned carrier execution.

Since you are now driving the underlying authorization model, could you take ownership of the remaining implementation and validation? My review found three follow-ups in the current head:

- Node-side `rawCommand` validation is stricter than upstream for trimmed text and safe PowerShell/POSIX wrapper previews.
- Handwritten `argPattern` rules use .NET regex semantics, which can differ from the gateway’s JavaScript regex semantics.
- Allowlist normalization deduplicates by `(pattern, argPattern)` even though `source` affects whether an entry authorizes anything, which can suppress a valid manual rule.

These appear fail-closed or limited to handwritten/low-level callers, but they are consequences of the redesigned model rather than the original fix. I would prefer not to keep layering contributor patches over an architecture you are actively revising.

PR #1177, opened 2026-08-19
fix(connection): trust MSIX-packaged wslrelay.exe when Authenticode check fails

larroy:
## Problem

Setup fails on machines running the modern MSIX-distributed WSL (`Program Files\WSL`, e.g. the Store/GitHub-released package) with:

```
Loopback port 18789 listener verification failed: wslrelay (PID ...): WSL relay Authenticode verification failed (Unsigned).
```

MSIX packages don't embed per-file Authenticode/catalog signatures on their EXEs — trust is established at the package level instead (verified by Windows at install time). So `wslrelay.exe`, `wsl.exe`, `wslhost.exe`, etc. from this distribution report `Unsigned` via `WinVerifyTrust`/`FileSignatureInfo`, even though they're genuine, unmodified Microsoft binaries (confirmed via `Get-AppxPackage`: `Publisher = CN=Microsoft Corporation, O=Microsoft Corporation, ...`).

## Fix

Add a fallback to `WindowsAuthenticodeVerifier.VerifyMicrosoftSignedFile`: when the primary Authenticode/catalog check fails **and** the file is `wslrelay.exe`, corroborate trust via the installed AppX package instead of rejecting outright. The fallback requires all of:

- Exact match on the known WSL package family name `MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe` (the suffix is derived from the publisher's signing certificate, so an impostor package re-signed with a different cert would get a different family name).
- A real `SignatureKind` (not `None`) — proves Windows cryptographically validated the package at install time.
- A Microsoft publisher, reusing the existing `HasMicrosoftPublisherIdentity` check unchanged.

Looked up via a `powershell.exe -Command "Get-AppxPackage ... | ConvertTo-Json"` shell-out, matching this file's existing pattern for invoking `schtasks.exe`/`wsl.exe` elsewhere in `ManagedLocalGatewayPortProvenanceService.cs`.

The primary Authenticode/catalog check is unchanged, always runs first, and is never bypassed — the fallback only narrows scope to `wslrelay.exe` and never weakens the trust bar for any other binary. No `.csproj`/TFM changes; no changes to the canonical-path allowlist.

## Testing

- `dotnet build src/OpenClaw.Connection/OpenClaw.Connection.csproj` — succeeds, 0 warnings/errors.
- `dotnet test tests/OpenClaw.Connection.Tests/OpenClaw.Connection.Tests.csproj` — 638/638 passing, including 6 new tests covering: trusted-via-fallback, no-package-found, family-name mismatch, unsigned package, non-Microsoft publisher, and a regression test proving the fallback is never consulted when Authenticode already succeeds.

### DIFF
--- a/src/OpenClaw.Connection/WindowsAuthenticodeVerifier.cs
+++ b/src/OpenClaw.Connection/WindowsAuthenticodeVerifier.cs
@@ -1,5 +1,9 @@
 using System;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
+using System.Text.Json;
+using System.Threading;
 using Microsoft.Security.Extensions;
 
 namespace OpenClaw.Connection;
@@ -11,9 +15,51 @@ internal readonly record struct AuthenticodeTrustResult(bool IsTrusted, string? 
     public static AuthenticodeTrustResult Rejected(string detail) => new(false, detail);
 }
 
+/// <summary>
+/// Minimal projection of a Windows AppX/MSIX package as reported by
+/// <c>Get-AppxPackage</c>, used to corroborate genuine Microsoft WSL binaries that
+/// ship without a per-file Authenticode/catalog signature (MSIX packages are
+/// trusted at the package level, not per-file).
+/// </summary>
+internal readonly record struct AppxPackageInfo(
+    string? PackageFamilyName,
+    string? Publisher,
+    string? SignatureKind);
+
 internal static class WindowsAuthenticodeVerifier
 {
-    public static AuthenticodeTrustResult VerifyMicrosoftSignedFile(string path)
+    // Modern WSL ships as an MSIX package. The trailing suffix is a hash derived
+    // from the publisher's signing certificate/public key, so an impostor package
+    // re-signed with a different certificate would get a different family name.
+    private const string WslPackageFamilyName =
+        "MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe";
+
+    public static AuthenticodeTrustResult VerifyMicrosoftSignedFile(string path) =>
+        VerifyMicrosoftSignedFile(path, LookupWslAppxPackageViaPowerShell);
+
+    /// <summary>
+    /// Test seam: allows callers to substitute a fake WSL AppX package lookup
+    /// instead of shelling out to PowerShell.
+    /// </summary>
+    internal static AuthenticodeTrustResult VerifyMicrosoftSignedFile(
+        string path,
+        Func<AppxPackageInfo?> lookupWslPackage)
+    {
+        var primary = VerifyAuthenticodeSignature(path);
+        if (primary.IsTrusted)
+            return primary;
+
+        // MSIX packages (modern WSL) don't carry a per-file Authenticode/catalog
+        // signature, so a failed classic check isn't conclusive for wslrelay.exe.
+        // Only consult the package-level fallback for the WSL binaries this
+        // codebase actually cares about; never widen trust for arbitrary files.
+        if (!string.Equals(Path.GetFileName(path), "wslrelay.exe", StringComparison.OrdinalIgnoreCase))
+            return primary;
+
+        return VerifyWslPackageFallback(primary, lookupWslPackage);
+    }
+
+    private static AuthenticodeTrustResult VerifyAuthenticodeSignature(string path)
     {
         try
         {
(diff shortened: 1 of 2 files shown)


shanselman:
The compatibility issue is real: some Store/MSIX WSL installations can report the genuine `wslrelay.exe` as unsigned at the individual-file layer even though Windows trusted the package at installation.

The current fallback cannot merge because it disconnects the file being trusted from the package being verified. A genuine WSL package can be installed while a different `wslrelay.exe` under the broad WindowsApps canonical-path rule owns the listener. The fallback validates only the separately queried package family/publisher/signature kind, then trusts the unrelated relay path. This provenance result gates shared/bootstrap credentials, so that is a security-boundary regression rather than only a setup compatibility issue.

Please change the fallback contract to receive the inspected path and prove package ownership:

1. Keep `FileSignatureInfo` as the primary path.
2. Query authoritative package location data in addition to family, publisher, and signature kind.
3. For WindowsApps layouts, require the inspected canonical path to be under that exact package's installed/effective location.
4. For the external `C:\Program Files\WSL\wslrelay.exe` layout, require the exact canonical path, no reparse-point ambiguity, protected SYSTEM/Admin ownership and write ACLs, and a package/file version binding. Fail closed if the external bytes cannot be bound to the installed package.
5. Add a sibling-WindowsApps relay rejection test and an external-path mismatch rejection test.
6. Inject the primary Authenticode result into tests. On WSL 2.9.4 here, `C:\Program Files\WSL\wslrelay.exe` is already validly Microsoft-signed, so the fallback is never called and 4 of the 638 tests fail: no package, family mismatch, unsigned package, and non-Microsoft publisher all unexpectedly return trusted.
7. Run full required validation and provide redacted current-head proof through an actually affected MSIX WSL setup/provenance path.

Observed on this host:

```text
WSL package: MicrosoftCorporationII.WindowsSubsystemForLinux_2.9.4.0_x64__8wekyb3d8bbwe
Package family: MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe
Package install location: C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_2.9.4.0_x64__8wekyb3d8bbwe
Relay path: C:\Program Files\WSL\wslrelay.exe
Relay Authenticode: Valid, Microsoft Corporation
Relay file/product version: 2.9.4.0
Relay ACL: writable only by SYSTEM and Administrators; Users and app packages are read/execute
```

The goal is good. The package-level fallback just needs to prove that the package owns or authoritatively installed the exact relay bytes before those bytes can pass the credential-protection gate.

shanselman:
Pedro, thank you for finding and addressing the underlying WSL compatibility problem. The important observation in this PR is correct: Windows can trust an MSIX package even when one of its executable files does not present the per-file Authenticode evidence we normally expect from a Win32 binary.

We took this over in maintainer PR #1201 because this check sits on a credential boundary and the remaining work needed fairly deep Windows provenance and ACL handling. Your original commit and authorship are preserved as the first commit in that PR.

The key Windows mental model is:

1. File trust and package trust answer different questions.
2. `FileSignatureInfo` tells us about the exact file or its catalog.
3. `Get-AppxPackage` tells us that Windows trusts an installed package identity.
4. Proving that the package is installed does not automatically prove that an arbitrary file named `wslrelay.exe` belongs to it.
5. At a credential boundary, we need an unbroken chain from listener, to process, to executable path, to package ownership.

The original fallback established package identity, publisher, and signature kind, but the inspected relay path was not passed into that decision. That meant a genuine WSL package could be installed while a different relay under the broad canonical-path rule inherited the package's trust.

The maintainer patch keeps your primary design and adds the missing binding:

- File-level Authenticode remains first.
- Fallback remains limited to `wslrelay.exe`.
- A WindowsApps relay must be beneath the exact install location reported for the genuine WSL package.
- A sibling WindowsApps package is rejected.
- Reparse points are rejected through the protected path ancestry.
- The external `C:\Program Files\WSL\wslrelay.exe` layout must match exactly.
- External path ownership and effective write ACLs are restricted to SYSTEM, Administrators, and TrustedInstaller.
- The external relay file version must match the installed WSL package version.
- Missing or ambiguous evidence fails closed.

There were also two subtle Windows ACL details worth calling out:

- An inherit-only `CREATOR OWNER` ACE describes rights that may be materialized on descendants. It does not grant those rights on the current directory.
- A null DACL and an empty DACL are opposites. A null DACL grants everyone full access; an empty DACL denies access. High-level rule enumeration can look empty in both cases, so security code must inspect the raw descriptor.

We also changed the tests so the primary Authenticode outcome is injected. On the maintainer host, WSL 2.9.4's external relay is already validly Microsoft-signed. Without injection, tests intended for package fallback can accidentally pass through the primary path and never test the fallback.

#1201 includes the full history, threat model, implementation walkthrough, validation counts, current-head live package/ACL/version proof, and the proof limitation. The final structured security review is clean.

Closing this PR as superseded by #1201. Thank you again for surfacing the compatibility issue and for providing the initial implementation that the maintainer patch builds on.

PR #1186, opened 2026-08-20
feat: add OpenClaw Gateway MSIX packaging

xlinush:
## What Problem This Solves

OpenClaw does not currently have a public, independently buildable MSIX path for packaging a pinned OpenClaw Gateway revision together with its Node.js runtime and a first-launch bootstrap experience.

## What This Adds

This PR adds a standalone Gateway MSIX implementation under `packaging/gateway-msix`:

- a .NET 10 NativeAOT bootstrap host for x64 and ARM64;
- a pinned and hash-verified OpenClaw payload plus the matching official Node.js runtime;
- transparent forwarding of every explicit command-line argument to the bundled OpenClaw CLI;
- package-owned payload preparation, repair, and diagnostics;
- unsigned, temporary test-signed, and policy-gated official-signing workflow modes;
- immutable release-policy checks for official signing with Azure OIDC;
- unit tests and public documentation for the host, staging, signing, and launch behavior.

The package is separate from the Companion application and uses the `OpenClaw.Gateway` package identity.

## Runtime Behavior

Launching `openclaw` without arguments prepares the bundled payload under `%USERPROFILE%\.openclaw-msix\app` and offers fast or full verification. Preparation uses a temporary directory and atomic promotion, so an existing payload is replaced only after the new payload is extracted and verified. Explicit arguments are forwarded unchanged to the bundled `openclaw.mjs`; the host does not reserve, consume, reject, rewrite, or implement OpenClaw commands.

Every child process receives:

```text
OPENCLAW_SUPERVISOR_MODE=external
OPENCLAW_NO_AUTO_UPDATE=1
```

These settings make the MSIX the owner of packaged code updates while leaving command behavior and enforcement to OpenClaw itself.

For first-run local setup, the documented command is:

```powershell
openclaw setup --classic --mode local --no-install-daemon
```

## Payload Integrity Boundary

The immutable payload archive inside the MSIX is hash-verified on launch. The prepared payload marker is then used to avoid re-hashing every extracted file during normal startup because complete per-file verification substantially delays launch. Users can explicitly request full verification and repair from the bootstrap menu.

The prepared directory is writable by the current user and is treated as user-owned application state, not as a tamper-resistant boundary against that same user context. The longer-term design is to run the Gateway payload in a dedicated isolated agent session rather than the interactive human session. That isolation is not part of this initial MSIX implementation.

## Signing and Release Controls

Pull-request and push builds produce unsigned validation packages. Manual runs support:

- `unsigned`: any branch, tag, or commit;
- `test`: any ref, signed with a temporary runner-generated certificate;
- `official`: only the immutable approved commit in `release-policy.json`, only from `main`, using the protected `release-signing` environment and Azure OIDC.

No signing private key is stored in the repository or uploaded as an artifact.

## Current-Head Validation

Tested packaging head: `fbc07162eb3409975d78a83c299f72fa0f428f5a`

Bundled OpenClaw commit: `0790d9f593ad30c940ed93b5872a8cf6d6f3cf8c`

- `./build.ps1`: passed. All repository projects built and documentation validation passed for 47 Markdown files.
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore`: passed. 3,791 passed, 0 failed, 32 skipped.
- `dotnet test ./packaging/gateway-msix/OpenClaw.Gateway.MSIX.slnx --configuration Release --no-restore`: passed. 42 passed, 0 failed, 0 skipped.
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore`: blocked locally. The assembly started and discovered the test file but produced no results for ten minutes, reproducing the existing local hang.
- Gateway MSIX signing-policy fixtures: passed.
- Local x64 NativeAOT MSIX `0.1.13.0`: built with 0 warnings and 0 errors and was test-signed successfully. SHA-256: `AD6E06A64435F68A400CDBD99E9C634480063C8797092B16C107937FEE340155`.
- Current-head GitHub Actions execution: not available because this first-time-contributor workflow still requires upstream approval.
## Windows Installation Proof

The following evidence was captured from the test-signed x64 package built from the current PR head, `fbc07162eb3409975d78a83c299f72fa0f428f5a`.

Tested package metadata:

- Packaging commit: `fbc07162eb3409975d78a83c299f72fa0f428f5a`
- Bundled OpenClaw commit: `0790d9f593ad30c940ed93b5872a8cf6d6f3cf8c`
- Architecture: x64
- Package version: `0.1.13.0`
- Test-signed MSIX SHA-256: `AD6E06A64435F68A400CDBD99E9C634480063C8797092B16C107937FEE340155`

### Artifact provenance and matching package hash

The generated metadata identifies the clean packaging commit, bundled OpenClaw commit, package version, signing type, and expected package hash. `Get-FileHash` independently reports the same SHA-256 for the tested MSIX.

Current-head MSIX provenance and matching SHA-256

### Installed package identity

Windows reports the current package as `OpenClaw.Gateway` version `0.1.13.0`, x64, installed under the packaged WindowsApps location with the expected OpenClaw Foundation publisher.

Current-head installed OpenClaw Gateway package identity

### Windows command resolution

Both `Get-Command openclaw -All` and `where.exe openclaw` resolve the tested command to the packaged Windows App Execution Alias.

Current-head Windows openclaw command resolution

### Real packaged CLI execution and argument forwarding

`openclaw --version` loads and verifies the packaged payload, forwards the explicit argument to the bundled OpenClaw CLI, reports bundled version `2026.7.1 (0790d9f)`, and observes child exit code 0. The final `Press Enter to close this window...` line also reproduces the outstanding review finding that explicit forwarded commands currently wait for an extra Enter.

Current-head packaged OpenClaw version execution and exit prompt

`openclaw --help` likewise reaches the bundled CLI and displays its real command and option output through the installed package.

Current-head forwarded OpenClaw help output

## Compatibility and State Ownership
OpenClaw currently assumes one default `openclaw` command and one default state root per user. Multiple Gateway instances are supported only through explicit isolation with `--profile` or separate `OPENCLAW_STATE_DIR`, `OPENCLAW_CONFIG_PATH`, and port settings. This MSIX follows that existing single-default-install model rather than attempting to introduce a general side-by-side installation solution.

The package currently registers the `openclaw.exe` App Execution Alias. If maintainers require simultaneous npm/native and MSIX installations, the project needs an explicit product-level command-routing, state-isolation, upgrade, and uninstall contract. If only one default installation is supported, the alias should instead be reviewed as the intended replacement and servicing path.

The bootstrapper deliberately performs no Gateway shutdown or state deletion. OpenClaw shutdown and shared-state cleanup are delegated to the documented [`openclaw uninstall`](https://docs.openclaw.ai/install/uninstall) flow. Removing the MSIX may leave the prepared `%USERPROFILE%\.openclaw-msix` directory, which users can remove manually after OpenClaw is stopped.

## Maintainer Decisions Requested

1. Whether this standalone Gateway MSIX and official-signing path should be added while the existing release documentation keeps MSIX distribution paused.
2. Whether the MSIX should own the global `openclaw.exe` alias or use a separate coexistence route.
3. Whether the documented same-user staged-payload integrity boundary is acceptable for the initial release pending isolated agent sessions.

## Change Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [x] Docs or instructions
- [x] Tests or validation
- [x] Security hardening
- [x] Chore or infrastructure

## Scope

- [ ] Tray or WinUI UX
- [ ] Windows node capability
- [ ] Local MCP or `winnode`
- [x] Setup or onboarding
- [x] Tests, CI, or docs

## Review Conversations

- [ ] I replied to or resolved every review conversation addressed by this PR.
- [x] Remaining unresolved conversations require maintainer product or ownership decisions described above.


packaging/gateway-msix/src/OpenClaw.Gateway.MSIXHost/OpenClawResetter.cs
```diff
@@ -0,0 +1,142 @@
+namespace OpenClaw.MSIXHost;
+
+public static class OpenClawResetter
+{
+    public static async Task ResetAsync(
```
MythiliMur: Do we really need Reset? With MSIX, uninstall path should actually stop gateway. No?

packaging/gateway-msix/src/OpenClaw.Gateway.MSIXHost/HostOptions.cs
```diff
@@ -0,0 +1,128 @@
+using System.Runtime.InteropServices;
+
+namespace OpenClaw.MSIXHost;
+
+public sealed record HostOptions(
```
MythiliMur, line 5: where do you disable "gateway /update"? Gateway started via MSIX should have disabled direct update through gitclone. Updates should happen through MSIX.

xlinush:
## Current-head Windows installation proof

Tested package metadata:

- Packaging commit: `af680854b9461ef995cf50c528259e5bb2ee38da`
- Bundled OpenClaw commit: `0790d9f593ad30c940ed93b5872a8cf6d6f3cf8c`
- Architecture: x64
- Package version: `0.1.12.0`
- Test-signed MSIX SHA-256: `86D7D30D49A4FD0B1204CC9743AD7BD035CDABC00F12117AD721840F307FC681`

### 1. Artifact provenance and matching package hash

Artifact provenance and matching SHA-256

### 2. Installed package identity

Installed OpenClaw Gateway package identity

### 3. Windows command resolution

Windows openclaw command resolution

### 4. Real packaged CLI execution and transparent argument forwarding

Packaged OpenClaw version execution

Forwarded OpenClaw gateway help output

### 5. Gateway startup

The packaged Gateway reached the ready state on the Windows test VM. The warning shown is from the VM's existing local development configuration and is not introduced by the package.

Packaged OpenClaw Gateway running and ready

packaging/gateway-msix/src/OpenClaw.Gateway.MSIXHost/HostOptions.cs
xlinush: Added the flags here: https://github.com/openclaw/openclaw-windows-node/pull/1186/changes#diff-6015b6e880674621da93cda64026a72259899b413a55ecd214f508f120bb8f0bR99

packaging/gateway-msix/src/OpenClaw.Gateway.MSIXHost/OpenClawResetter.cs
xlinush: > uninstall path should actually stop gateway. No?

Uninstalling the MSIX does not reliably stop a running gateway, nor does it clean up user states; uninstallation removes the package from WindowsApps and frees up the `openclaw` AEA. It leaves these traces on user's device:
1.  %USERPROFILE%\.openclaw-msix  => where we extract the payload
2.  %USERPROFILE%\.openclaw           => where openclaw keeps its state and configs

For 1, we need to have a reset button for users to delete the extract payload; for 2, users need to rely on `openclaw gateway stop` + `openclaw uninstall` as documented in https://docs.openclaw.ai/install/uninstall. I'd prefer we keep the reset logic, to provide some convenience for users in case they want a clean uninstallation.

As a side note, I was told earlier by Jeff/Dom that there's some uninstallation work being done when a package is installed for an agent user/session. The main idea is when such a package is uninstalled, the agent session will be teared down and wiped out, so there won't be any 'manual cleanup' needed on our end. I don't know how true that statement is as of today, and it's outside the scope of this PR.

xlinush:
@clawsweeper re-review

packaging/gateway-msix/src/OpenClaw.Gateway.MSIXHost/OpenClawResetter.cs
xlinush: Resolved in `fbc07162`. I removed the Reset actions, `OpenClawResetter`, and host-side Gateway process registration entirely. The bootstrapper now only prepares or fully verifies/repairs the packaged payload.

Gateway shutdown and shared `.openclaw` cleanup are delegated to OpenClaw's documented `openclaw uninstall` flow. MSIX removal may leave `%USERPROFILE%\.openclaw-msix`, which is now documented as manual cleanup after OpenClaw is stopped. Payload upgrades remain safe because preparation already uses a temporary directory, atomic promotion, rollback, and deletion of the previous payload only after successful promotion.

xlinush:
@clawsweeper re-review

xlinush:
@clawsweeper re-review

shanselman:
Thank you for putting this together. This is a substantial and thoughtful first contribution. The provenance metadata, payload validation, x64/ARM64 structure, signing-policy work, tests, and documentation show a lot of care. The feedback below is intended to help us settle the Windows product and lifecycle contract, rather than to discount the implementation work already here.

## Overall direction

I think a small Windows launcher is reasonable. Windows needs a real executable for an MSIX App Execution Alias, and the launcher has useful work to do before invoking the Node-based OpenClaw CLI. NativeAOT is also a reasonable implementation choice for a small, stable launcher because it gives us a self-contained native executable with no installed .NET dependency.

NativeAOT is not itself required by MSIX, though. A self-contained managed .NET executable could do the same job. The important architectural decision is the launcher and package lifecycle, not AOT specifically.

## Naming and product identity

I think `openclaw.exe` is the right executable name if this package is intended to become the canonical native-Windows distribution of the complete OpenClaw CLI. Users should be able to type `openclaw` on Windows, macOS, Linux, and WSL even though each platform reaches `openclaw.mjs` differently.

The internal project name `OpenClaw.Gateway.MSIXHost` is less clear. This process does not host the Gateway or Node in-process. It verifies/prepares files and launches a child process. Something like `OpenClaw.MsixLauncher` or `OpenClaw.WindowsLauncher` would describe that responsibility more directly.

There is also a product-identity question: the package is named `OpenClaw.Gateway`, while explicit arguments expose the entire OpenClaw CLI. If this distributes the complete CLI, `OpenClaw.CLI` or `OpenClaw` may be the more accurate package identity. If it is intentionally Gateway-only, a distinct `openclaw-gateway.exe` command may be more honest.

## App Execution Alias

An App Execution Alias is the native MSIX mechanism for exposing a CLI command, so the manifest approach is correct in isolation. The unresolved concern is ownership of the global `openclaw` command.

Today an npm installation provides `openclaw.cmd` and `openclaw.ps1`. This MSIX would add `%LOCALAPPDATA%\Microsoft\WindowsApps\openclaw.exe`. Which command wins can depend on the shell, PATH ordering, PATHEXT, and whether the user has disabled an alias in Windows Settings. We could end up with two package managers owning the same command while both use the same `~\.openclaw` state.

Before taking the global alias, I think we need an explicit decision between these models:

1. The MSIX replaces npm as the supported native-Windows installation. In that case, `openclaw.exe` and the alias are correct, but we need migration, conflict detection, upgrade, uninstall, and state-ownership behavior.
2. The MSIX coexists with npm. In that case, a preview alias such as `openclaw-msix.exe` is safer until coexistence is designed and proven.
3. The package installs only the Gateway. In that case, it should probably use an explicit Gateway command instead of claiming the general CLI name.

My preference is model 1 as the eventual user experience, but model 2 while this is still an experimental distribution path.

## Keep command behavior cross-platform

Packaging can be platform-specific, but `openclaw` should mean the same thing everywhere:

```text
macOS/Linux: openclaw -> node openclaw.mjs
Windows:     openclaw -> openclaw.exe -> node openclaw.mjs
```

The Windows launcher should be as transparent as possible. Currently, a zero-argument invocation is intercepted for package preparation instead of being forwarded to OpenClaw as it is on other platforms. That makes behavior depend on the installation channel.

I would prefer explicit setup and repair experiences, such as `openclaw setup` and `openclaw doctor`, or a separate Start menu application named **OpenClaw Setup**. A pure CLI package should otherwise use `AppListEntry="none"`, as recommended by Microsoft's CLI MSIX guidance.

## Reconsider extracting the application from the MSIX

The largest architectural concern is not NativeAOT. It is packaging a tar archive inside the MSIX and then copying the application into `%USERPROFILE%\.openclaw-msix\app`.

That gives us two copies of the application and requires custom hashing, inventory markers, locks, rollback, repair, cleanup, and update coordination. It also moves the executable code out of the immutable, tamper-protected MSIX package into a same-user-writable directory. Some of the strongest benefits of MSIX are therefore lost.

Before committing to that design, I think we should prove whether bundled `node.exe` and `openclaw.mjs` can run directly from the read-only package installation. OpenClaw configuration and mutable state can remain under the user profile, but packaged application code ideally remains package-owned and immutable.

If extraction is genuinely required because of Node/npm layout constraints, that requirement should be documented with evidence. The extracted layout should then use immutable, content-addressed version directories. New processes can select the new version without renaming or deleting the directory used by an already-running Gateway.

## Merge-blocking correctness issues

I independently reviewed the current head and agree with these three blockers:

1. `.github/workflows/gateway-msix.yml:348` interpolates `openclaw_ref` directly into a PowerShell script. A crafted workflow-dispatch value can inject PowerShell and bypass the official-signing authorization check. Please pass the value through an environment variable and read it as data.
2. `Program.cs:151` waits for Enter after every interactive invocation. This makes ordinary commands such as `openclaw --version` and `openclaw status` appear to hang. The pause should apply only to a deliberate Explorer/Start-menu bootstrap experience, not forwarded CLI commands.
3. `PayloadStager.cs:205` renames the installed payload directory without coordinating with a running Gateway. On Windows this can fail on locked files or disrupt the live process. Versioned payload directories or explicit lifecycle coordination are needed.

## Proof still needed

Before treating this as a production distribution path, I would want to see:

- A successful current-head GitHub Actions run.
- Real ARM64 installation and runtime proof, not only an ARM64 build matrix.
- Windows App Certification Kit coverage.
- Install, update, rollback, and uninstall proof while a Gateway is running.
- Coexistence or migration proof for an existing npm installation.
- End-to-end proof that `OPENCLAW_NO_AUTO_UPDATE=1` actually prevents an upstream self-update.
- A maintainer decision on the repository's currently paused MSIX distribution policy.

The focused host tests and signing-policy fixtures pass, which is a good foundation. They do not currently cover the three lifecycle/signing blockers above.

## Suggested landing path

I would preserve this work but separate experimentation from global product ownership:

1. Land or iterate on the packaging path using a preview-specific identity/alias and unsigned or test-signed artifacts.
2. Resolve whether application code can run directly from the immutable package.
3. Define npm replacement/coexistence, Gateway lifecycle, state, update, and uninstall contracts.
4. Prove both architectures and the real installed behavior.
5. Only then claim the global `openclaw.exe` alias and enable official signing/distribution.

Again, there is a lot of solid engineering here. The remaining questions are mostly about choosing a durable Windows product boundary so this launcher stays small and understandable instead of becoming a second package manager around OpenClaw.

shanselman:
Important clarification on my review above: the intended model is that the **Companion/Windows Node app is distributed through the Microsoft Store**, while the **Gateway is a separate, non-Store MSIX that the Companion optionally acquires and installs**.

That context changes a few parts of the assessment:

- The `OpenClaw.Gateway` package identity makes more sense as a separately installed component.
- The Gateway MSIX does not itself need to be a Store submission, although it should still be validated as a production MSIX.
- A global `openclaw.exe` App Execution Alias can be appropriate if installing the optional Gateway is also the explicit act that installs the native-Windows OpenClaw CLI.
- The Companion, rather than the launcher, should ideally own acquisition, installation, update, repair, rollback, and uninstall of the Gateway package.

It also makes the boundary between the two packages more important. Before treating this as the release architecture, I think we need a written and tested contract for:

1. How the Store-delivered Companion obtains and installs the independently signed Gateway MSIX, including user consent and the supported Windows deployment API or App Installer flow.
2. Stable publisher and package identity, trusted signing, version discovery, update source, rollback, and uninstall behavior.
3. Store certification/policy confirmation for a Store app that downloads and initiates installation of a separate non-Store executable package. I would confirm this exact scenario with Partner Center rather than assume it is accepted.
4. Whether the Companion prevents, detects, or explains conflicts with an existing npm `openclaw` installation before enabling the `openclaw.exe` alias.
5. Which component stops and restarts a running Gateway during package updates.
6. Which component owns shared state under `~\.openclaw`, and what remains after either package is removed.

This model strengthens the argument for making the Gateway MSIX itself simple and immutable. Ideally it contains the runnable Node/OpenClaw payload, while the Companion updates the whole signed package. Extracting another mutable application copy and implementing a second repair/update system inside `openclaw.exe` duplicates responsibilities that the Companion plus MSIX deployment layer could own.

So my adjusted recommendation is:

- Keep `OpenClaw.Gateway` as the separate component identity.
- Keep `openclaw.exe` and the App Execution Alias if the optional install is explicitly intended to provide the canonical Windows CLI.
- Rename the internal `MSIXHost` project to `Launcher` or `Bootstrapper` for clarity.
- Have the Companion own Gateway package lifecycle.
- Keep the launcher transparent: set environment, forward arguments, launch packaged Node/OpenClaw, and return the exit code.
- Resolve npm coexistence and Store-policy validation before enabling this as a production acquisition path.

Thank you for the architectural context. It makes the two-package direction much clearer and narrows the key design question to the contract between Companion, Gateway MSIX, and the existing cross-platform CLI.

MythiliMur:
Thanks for the detailed review @shanselman. The worry that I have with app execution alias if that we will always have to do catchup with whatever happens in the actual openclaw gateway which I want to avoid if possible. Seems like you already made it optional in your adjusted recommendation.

Things that will be captured in next set of PRs specifically for MSIX:
1. Where will these MSIXs be stored - CDN, github releases or something else. We have alignment that it won't be in Store as companion app will go there.
2. MSIX update story for prosumer - will companion app take care of it? Should we add it part of existing /update command and/or provide a new generic mechanism like baked in updateprovider in gateway
3. Companion app consuming this MSIX instead of WSL or even npm for native windows gateway install.

MythiliMur:
- Make sure we are not bundling any other package (e.g., node.js)
- see if you can do winget install as part of MSIX install. Can it run a setup script of some sort?

RomneyDa:
Closing to be safe, this has been moved to https://github.com/openclaw/openclaw-msix-packaging

PR #1201, opened 2026-08-21
fix(connection): bind WSL relay trust to installed package

shanselman:
## Summary

This is a maintainer follow-up to #1177. It preserves Pedro's original compatibility fix as the first commit, then completes the package-to-file provenance checks requested in the security review.

The compatibility problem in #1177 is real. Some Store/MSIX WSL installations can report a genuine `wslrelay.exe` as unsigned when inspected as an individual PE file, even though Windows trusted the MSIX package that installed it. The original PR correctly recognized that package trust can be relevant when per-file Authenticode is unavailable.

The missing piece was proving that the exact relay process owning the local listener came from that trusted package. It is not sufficient to prove these two facts independently:

1. A genuine Microsoft WSL package is installed.
2. A file named `wslrelay.exe` exists somewhere under a broadly accepted path.

The trusted package must own, or must be authoritatively bound to, the exact relay bytes being trusted.

## Credit

Pedro identified the Store/MSIX compatibility failure and implemented the initial package-level fallback in #1177. His original commit is preserved on this branch, including its authorship.

This follow-up does not replace that contribution. It builds on it by tightening the provenance contract at the credential-protection boundary.

## Background: why Windows signing has two relevant layers

Traditional Windows desktop binaries are often authenticated at the file layer:

- An executable can carry an embedded Authenticode signature.
- Windows can also establish file trust through a signed catalog.
- `FileSignatureInfo` gives us a useful file-level answer for those cases.

MSIX adds a package-level trust model:

- Windows validates the signed package during deployment.
- The package identity includes a package family, publisher, version, and installed location.
- Individual executable files inside or associated with the package do not always present the same file-level signature evidence that a traditional Win32 application would.

That means a failed per-file Authenticode result is not always proof that the WSL installation is malicious. However, package trust is not automatically transferable to an arbitrary file with the right name. Package trust answers "Windows trusts this package." It does not, by itself, answer "this exact process image came from that package."

## Why this is a security boundary

`ManagedLocalGatewayPortProvenanceService` is not merely displaying diagnostics. Its result helps decide whether OpenClaw may send shared or bootstrap credentials to a loopback listener.

Loopback is a network location, not an identity. A different local process can bind the expected port. If we trust a relay only because:

- its path looks generally plausible, and
- a genuine WSL package happens to be installed,

then an unrelated relay can inherit the package's trust without being owned by it. That creates a path for local credential disclosure.

The rule for this boundary is therefore:

> Before strong credentials cross the loopback boundary, every link from listener to process to executable to package must be positively established. Missing, ambiguous, or inconsistent evidence fails closed.

### 1. Keep file-level Authenticode first

The normal `FileSignatureInfo` path remains the primary check.

If the relay is signed and trusted by Microsoft at the file or catalog layer, the result returns immediately. Package lookup and filesystem fallback logic are not consulted.

Tests now inject this primary result explicitly. This matters because the relay on the maintainer test host is already validly Microsoft-signed. Without injection, fallback tests silently exercise the fast path and can appear to pass without testing package fallback at all.

### 2. Limit fallback to `wslrelay.exe`

A failed signature on any other filename returns the original failure. Installing WSL does not widen trust for arbitrary executables.

### 3. Validate the installed package identity

The fallback queries `Get-AppxPackage` for the active WSL package and reads:

- package family name
- publisher
- signature kind
- install location
- package version

The package must:

- match the known WSL package family exactly
- have a non-empty, non-`None` signature kind
- contain the exact `O=Microsoft Corporation` publisher identity
- report an absolute installed location

The PowerShell projection converts enum and version values to strings before JSON serialization. This was verified against the live `Get-AppxPackage` object because `SignatureKind` otherwise serializes in a shape the parser does not accept.

### 4. Bind WindowsApps relays to the exact package location

For a WindowsApps layout:

- the reported package install location must itself be below the real `Program Files\WindowsApps` root
- the inspected relay must be below that exact package install location
- a sibling package directory is rejected
- every path component from the relay through `WindowsApps` and `Program Files` is checked for reparse points
- ownership and effective write ACLs are checked on every component

This closes the original disconnect between "the WSL package is trusted" and "this relay path belongs to that package."

### 5. Bind the external Program Files relay

Current WSL packages can also use:

`C:\Program Files\WSL\wslrelay.exe`

That file is outside the package's immutable WindowsApps install directory, so package-location containment cannot prove ownership. The external path receives a stricter composite proof:

- the path must exactly equal the canonical `Program Files\WSL\wslrelay.exe`
- the relay and every parent through `Program Files` must contain no reparse point
- each component must be owned by SYSTEM, Administrators, or TrustedInstaller
- no other principal may have an effective allow ACE containing write, append, attribute-write, delete, DACL-change, or ownership-change rights
- the relay's file version must exactly equal the installed WSL package version

The exact version check binds the protected external relay to the installed package generation. A mismatch or an unreadable version fails closed.

## Windows ACL details worth preserving

Windows ACLs have several details that are easy to miss in security-sensitive code.

### Ownership is not the same as access

A SYSTEM-owned file can still be writable by another principal if its DACL grants that access. The implementation checks both owner and effective allow rules.

### Read access is expected

Users and application packages normally have read and execute access to files under `Program Files`. That is safe for this decision. The policy rejects modification rights, not ordinary read and execute rights.

### Inherit-only ACEs do not apply to the current object

A standard `Program Files` ACL can contain a `CREATOR OWNER` ACE marked `InheritOnly`. It describes permissions to materialize on descendants. It does not grant the creator rights on the current directory.

The policy ignores inherit-only ACEs while evaluating the current object. Descendants are each inspected separately, so effective inherited permissions are still checked where they apply.

### A null DACL is not an empty DACL

This distinction is especially important:

- An empty DACL denies access.
- A null or absent DACL grants full access to everyone.

High-level access-rule enumeration can look empty in both cases. The implementation reads the raw security descriptor and requires the `DiscretionaryAclPresent` control flag plus a non-null DACL before evaluating rules.

## Tests added or corrected

The focused tests now cover:

- primary Authenticode success never invokes fallback
- non-`wslrelay.exe` failures never invoke fallback
- a package-owned WindowsApps relay is accepted
- a sibling WindowsApps package relay is rejected before path inspection
- the protected external relay is accepted only with a package-version match
- an external relay version mismatch is rejected
- an unprotected external path is rejected
- a relative or unavailable package location is rejected
- missing package, wrong family, unsigned package, and non-Microsoft publisher are rejected
- Users may retain read and execute access
- a Users write ACE is rejected
- an untrusted owner is rejected
- an inherit-only `CREATOR OWNER` ACE does not create a false rejection
- a null DACL is rejected
- an empty DACL remains safe

## Validation

Executed from the isolated `C:\p1177` maintainer checkout with `OPENCLAW_REPO_ROOT=C:\p1177`:

```text
.\build.ps1
All builds succeeded.

Passed: 3791, Skipped: 32, Failed: 0

Passed: 2684, Skipped: 0, Failed: 0

dotnet test .\tests\OpenClaw.Connection.Tests\OpenClaw.Connection.Tests.csproj --no-restore
Passed: 673, Skipped: 0, Failed: 0

Focused ManagedLocalGatewayPortProvenanceServiceTests
Passed: 44, Skipped: 0, Failed: 0
```

One unrelated telemetry timing test failed once during closeout:

`McpHttpServerTelemetryTests.ShutdownWhileWaitingForHandler_RecordsShutdownNotBusyOrTimeout`

The exact test passed immediately on rerun. The complete required sequence then passed.

## Real behavior proof

The maintainer host has WSL 2.9.4 and a validly Microsoft-signed external relay, so normal execution returns through the primary Authenticode path. To exercise the fallback itself, the current-head assembly was invoked with an injected primary failure while retaining the production package lookup and production filesystem/ACL inspector.

```text
RelayPath              : C:\Program Files\WSL\wslrelay.exe
PackageFamilyName      : MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe
PackageSignatureKind   : Developer
PackageInstallLocation : C:\Program Files\WindowsApps\
                         MicrosoftCorporationII.WindowsSubsystemForLinux_2.9.4.0_x64__8wekyb3d8bbwe
PackageVersion         : 2.9.4.0
RelayPathSecure        : True
RelayFileVersion       : 2.9.4.0
FallbackTrusted        : True
Detail                 :
```

This proves that the current-head package query, JSON projection, canonical external path, reparse walk, raw DACL handling, ownership/write policy, and version binding compose successfully on a real WSL installation.

### Proof limitation

A host where the genuine relay naturally reports an unsigned file-level result was not available for this maintainer pass. The fallback was therefore forced only at the injected primary-result seam. All downstream evidence came from the real installed package and real relay filesystem state.

## Review

The security-focused rubber-duck pass found no trust bypass. Its compatibility and path-ancestor suggestions were incorporated.

The structured ship review then found two Windows ACL edge cases:

1. Inherit-only `CREATOR OWNER` ACEs must not be treated as effective rights on the current object.
2. A null DACL must be distinguished from an empty DACL and rejected.

Both findings were fixed and covered by tests. The final structured review result was:

```text
autoreview clean: no accepted/actionable findings reported
overall: patch is correct
```

## Relationship to #1177

This PR supersedes #1177 for landing purposes while preserving and crediting its original compatibility contribution. The main change is not the goal of the fallback, which remains correct. The change is the proof required before package trust can be transferred to the exact listener-owning relay at a credential boundary.

### DIFF
--- a/src/OpenClaw.Connection/WindowsAuthenticodeVerifier.cs
+++ b/src/OpenClaw.Connection/WindowsAuthenticodeVerifier.cs
@@ -2,6 +2,9 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Runtime.Versioning;
+using System.Security.AccessControl;
+using System.Security.Principal;
 using System.Text.Json;
 using System.Threading;
 using Microsoft.Security.Extensions;
@@ -24,7 +27,21 @@ internal readonly record struct AuthenticodeTrustResult(bool IsTrusted, string? 
 internal readonly record struct AppxPackageInfo(
     string? PackageFamilyName,
     string? Publisher,
-    string? SignatureKind);
+    string? SignatureKind,
+    string? InstallLocation,
+    string? Version);
+
+internal readonly record struct WslRelayPathInspection(
+    bool IsSecure,
+    string? FileVersion,
+    string? Detail)
+{
+    public static WslRelayPathInspection Secure(string? fileVersion) =>
+        new(true, fileVersion, null);
+
+    public static WslRelayPathInspection Rejected(string detail) =>
+        new(false, null, detail);
+}
 
 internal static class WindowsAuthenticodeVerifier
 {
@@ -33,19 +50,37 @@ internal static class WindowsAuthenticodeVerifier
     // re-signed with a different certificate would get a different family name.
     private const string WslPackageFamilyName =
         "MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe";
+    private const string TrustedInstallerSid =
+        "S-1-5-80-956008885-3418522649-1831038044-1853292631-2271478464";
 
     public static AuthenticodeTrustResult VerifyMicrosoftSignedFile(string path) =>
-        VerifyMicrosoftSignedFile(path, LookupWslAppxPackageViaPowerShell);
+        VerifyMicrosoftSignedFile(
+            path,
+            VerifyAuthenticodeSignature,
+            LookupWslAppxPackageViaPowerShell,
+            InspectWslRelayPath);
 
     /// <summary>
-    /// Test seam: allows callers to substitute a fake WSL AppX package lookup
-    /// instead of shelling out to PowerShell.
+    /// Test seam for the primary signature, WSL package, and protected-path
+    /// evidence used by the package fallback.
     /// </summary>
     internal static AuthenticodeTrustResult VerifyMicrosoftSignedFile(
         string path,
-        Func<AppxPackageInfo?> lookupWslPackage)
+        Func<string, AuthenticodeTrustResult> verifyAuthenticode,
+        Func<AppxPackageInfo?> lookupWslPackage,
+        Func<string, string, WslRelayPathInspection> inspectRelayPath)
     {
-        var primary = VerifyAuthenticodeSignature(path);
+        AuthenticodeTrustResult primary;
+        try
+        {
+            primary = verifyAuthenticode(path);
+        }
+        catch
+        {
+            return AuthenticodeTrustResult.Rejected(
+                "WSL relay Authenticode verification could not complete.");
+        }
+
         if (primary.IsTrusted)
             return primary;
 
@@ -56,7 +91,7 @@ internal static class WindowsAuthenticodeVerifier
         if (!string.Equals(Path.GetFileName(path), "wslrelay.exe", StringComparison.OrdinalIgnoreCase))
             return primary;
 
-        return VerifyWslPackageFallback(primary, lookupWslPackage);
+        return VerifyWslPackageFallback(path, primary, lookupWslPackage, inspectRelayPath);
     }
 
     private static AuthenticodeTrustResult VerifyAuthenticodeSignature(string path)
@@ -92,8 +127,10 @@ internal static class WindowsAuthenticodeVerifier
     }
 
     private static AuthenticodeTrustResult VerifyWslPackageFallback(
+        string path,
         AuthenticodeTrustResult primaryFailure,
-        Func<AppxPackageInfo?> lookupWslPackage)
+        Func<AppxPackageInfo?> lookupWslPackage,
+        Func<string, string, WslRelayPathInspection> inspectRelayPath)
     {
         AppxPackageInfo? package;
         try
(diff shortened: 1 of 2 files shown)


shanselman:
We are pausing landing while we identify why the external relay appeared unsigned. This is not a rejection of the compatibility report.

The upstream WSL release pipeline is designed to:

1. Authenticode-sign and verify `wslrelay.exe`.
2. Copy the signed executable into the MSI staging directory.
3. Install that MSI payload under `C:\Program Files\WSL`.

The current maintainer host has WSL 2.9.4, and its external `wslrelay.exe` has valid Microsoft Authenticode. If another official installation reports `Unsigned`, we need to determine which of these cases occurred:

- `Get-AuthenticodeSignature` also considers the file unsigned, indicating a WSL build, channel, installation, update, or file-integrity problem.
- PowerShell considers it valid but OpenClaw's `FileSignatureInfo` reports `Unsigned`, indicating a verifier compatibility bug we should fix.
- The file came from a development, nightly, preview, or locally built WSL package where release signing was not applied.

@larroy, could you please run the following read-only PowerShell and paste the output? It contains no credentials. You may redact usernames if one unexpectedly appears.

```powershell
$relay = 'C:\Program Files\WSL\wslrelay.exe'
$package = Get-AppxPackage -Name MicrosoftCorporationII.WindowsSubsystemForLinux |
    Sort-Object Version -Descending |
    Select-Object -First 1

Write-Output '=== WSL version ==='
wsl.exe --version

Write-Output '=== WSL package ==='
$package |
    Select-Object Name, PackageFullName, PackageFamilyName, Version,
        SignatureKind, Publisher, InstallLocation |
    Format-List

Write-Output '=== Relay file ==='
Get-Item -LiteralPath $relay |
    Select-Object FullName, Length, CreationTimeUtc, LastWriteTimeUtc,
        Attributes, LinkType,
        @{Name='FileVersion'; Expression={$_.VersionInfo.FileVersion}},
        @{Name='ProductVersion'; Expression={$_.VersionInfo.ProductVersion}} |
    Format-List

Write-Output '=== Relay hash ==='
Get-FileHash -LiteralPath $relay -Algorithm SHA256 | Format-List

Write-Output '=== PowerShell Authenticode ==='
$signature = Get-AuthenticodeSignature -LiteralPath $relay
$signature |
    Select-Object Status, StatusMessage, SignatureType,
        @{Name='SignerSubject'; Expression={$_.SignerCertificate.Subject}},
        @{Name='SignerThumbprint'; Expression={$_.SignerCertificate.Thumbprint}} |
    Format-List

Write-Output '=== Optional signtool verification ==='
$signtool = Get-Command signtool.exe -ErrorAction SilentlyContinue
if ($signtool) {
    & $signtool.Source verify /pa /all /v $relay
} else {
    Write-Output 'signtool.exe is not on PATH; skipped.'
}
```

Please also tell us how this WSL build was installed or updated:

- Microsoft Store
- `wsl --install` or `wsl --update`
- `wsl --update --web-download`
- GitHub release MSI
- Insider, preview, nightly, or development build
- locally built WSL

The key comparison is the existing OpenClaw error (`FileSignatureInfo` reported `Unsigned`) versus `Get-AuthenticodeSignature` and, when available, `signtool verify`.

We will keep this PR as a draft investigation until that distinction is clear. If the exact Microsoft file is valid and only our API path disagrees, we will fix the verifier narrowly. If the file itself is unsigned, we should not transfer package trust to unrelated external bytes; we will document the repair/update path and report an upstream WSL signing defect when appropriate.

larroy:
```

=== WSL version ===
WSL version: 2.6.3.0
Kernel version: 6.6.87.2-1
WSLg version: 1.0.71
MSRDC version: 1.2.6353
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.28000.2704
=== WSL package ===


Name              : MicrosoftCorporationII.WindowsSubsystemForLinux
PackageFullName   : MicrosoftCorporationII.WindowsSubsystemForLinux_2.6.3.0_arm64__8wekyb3d8bbwe
PackageFamilyName : MicrosoftCorporationII.WindowsSubsystemForLinux_8wekyb3d8bbwe
Version           : 2.6.3.0
SignatureKind     : Developer
Publisher         : CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US
InstallLocation   : C:\Program Files\WindowsApps\MicrosoftCorporationII.WindowsSubsystemForLinux_2.6.3.0_arm64__8wekyb3d8bbwe



=== Relay file ===


FullName         : C:\Program Files\WSL\wslrelay.exe
Length           : 3749888
CreationTimeUtc  : 12/12/2025 9:24:08 AM
LastWriteTimeUtc : 12/12/2025 9:24:08 AM
Attributes       : Archive
LinkType         :
FileVersion      : 2.6.3.0
ProductVersion   : 2.6.3.0



=== Relay hash ===


Algorithm : SHA256
Hash      : D4EB41F8E383A8EE8CE005DF1F23AF54978139164C862F31C6B4E62D118C1D58
Path      : C:\Program Files\WSL\wslrelay.exe



=== PowerShell Authenticode ===


Status           : NotSigned
StatusMessage    : The file C:\Program Files\WSL\wslrelay.exe is not digitally signed. You cannot run this script on the current system. For more information about running scripts and setting execution
                   policy, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170
SignatureType    : None
SignerSubject    :
SignerThumbprint :



=== Optional signtool verification ===
signtool.exe is not on PATH; skipped.

```

shanselman:
Closing this draft because it weakens the current credential-boundary invariant: the canonical wslrelay executable must be Authenticode SignedAndTrusted by Microsoft. Package metadata, ACLs, and a Developer package signature do not establish equivalent file-level provenance. If an officially released Microsoft-signed relay is misclassified, please open a narrow verifier fix with that signed artifact as proof.

PR #1207, opened 2026-08-22
fix(setup): preserve trailing command output while bounding inherited-pipe waits

### LINKED ISSUE #1194: Preserve trailing command output while bounding inherited-pipe waits
## Problem

`CommandRunner` waits for redirected stdout and stderr EOF after the direct child exits, but gives every command only 250 ms for the asynchronous output callbacks to finish draining.

https://github.com/openclaw/openclaw-windows-node/blob/31dffdb0dbd9eebc8b98ce35cf13ab60762410a9/src/OpenClaw.SetupEngine/CommandRunner.cs#L193-L199

The bounded drain was introduced to prevent setup from hanging when a surviving descendant inherits the redirected pipe handles. The immediate motivating command was the Welcome-page WSL inspection:

```text
wsl.exe --list --quiet
```

https://github.com/openclaw/openclaw-windows-node/blob/31dffdb0dbd9eebc8b98ce35cf13ab60762410a9/src/OpenClaw.SetupEngine/ExistingConfigDetector.cs#L32-L37

That inherited-handle scenario is legitimate, but the 250 ms escape now applies to all `CommandRunner` and `RunInWslAsync` calls, including commands whose complete output is parsed as JSON or checked for trailing success markers. Examples include:

```text
wsl.exe -d <distro> -- bash -c "<path-prefix> && openclaw nodes list --json"
wsl.exe -d <distro> -- bash -c "<path-prefix> && openclaw devices approve --latest --json"
wsl.exe -d <distro> -- bash -c "<path-prefix> && openclaw qr --json"
```

The setup flow also checks trailing markers such as `GATEWAY_CONFIGURED`, `CONFIGURED_OK`, `OPENCLAW_FRESH_WSL_READY`, and `WINDOWS_NODE_CONTEXT_READY`.

## Impact

If output callbacks are still draining when the 250 ms delay wins, setup can receive truncated JSON or miss a final success marker despite a successful child exit. This can produce parse failures or false setup failures.

## Expected behavior

- Preserve complete stdout and stderr when the direct child closed its pipe handles normally.
- Keep a bounded escape for the case where surviving descendants retain inherited handles.
- Preserve existing process timeout and caller-cancellation behavior.
- Do not regress the no-hang behavior for orphan or background WSL descendants.

## Tests

- Add a high-volume command whose required JSON or marker is emitted at the end and assert that it is captured.
- Retain the existing descendant-keeps-pipes-open test.
- Cover stdout and stderr draining.
- Cover timeout and caller cancellation.

Raised by @larroy during the PR #1178 review follow-up.

karkarl:
## Summary

`CommandRunner.RunAsync` always applied a short bounded grace (`s_outputDrainGrace`) to the stdout/stderr drain after the child process exited. On high-volume commands whose output was still draining when that grace elapsed, trailing output could be truncated even though the process exited cleanly.

This change preserves complete output on the normal path and keeps the bounded wait only where it is actually needed:

- On the normal exit path, `RunAsync` now awaits full stdout/stderr close so all trailing output is captured.
- The bounded grace is applied only on timeout cleanup, or when a caller opts in via the new `allowInheritedPipeHandleEscape` parameter for WSL inspection commands whose descendants can keep inherited pipe handles open after the direct child exits.
- The `wsl.exe --list` / `--verbose` / `--version` / `--status` inspection callers (`CleanupStaleDistroStep`, `CreateWslInstanceStep`, `ExistingConfigDetector`, `StartGatewayStep`, `PreflightWslStep`) opt into the bounded escape, since those are the calls that can leave a descendant holding inherited handles.

Timeout and cancellation behavior are unchanged.

Fixes #1194

## Validation

- `dotnet test tests/OpenClaw.SetupEngine.Tests` (subsystem touched by this change): full suite **949/949 passed** after restoring against a local NuGet feed.
- Focused `CommandRunnerTests`: **7/7 passed**, including the new high-volume drain and bounded-timeout tests.
- Required `./build.ps1` and the Shared/Tray closeout suites are **Not verified / blocked**: NuGet cannot reach `https://api.nuget.org/v3/index.json` on this host (NU1900 "Unable to load the service index"; underlying TLS/service-index handshake failure), and the global cache lacks `Microsoft.ML.OnnxRuntime 1.29.0`. A maintainer on a network-enabled host should run `./build.ps1`, `dotnet test tests/OpenClaw.Shared.Tests`, and `dotnet test tests/OpenClaw.Tray.Tests` to complete closeout before merge.

## Real behavior proof

- New regression test `RunAsync_DrainsHighVolumeStdoutAndStderrThroughTrailingMarkers` runs a command emitting 8,000 lines on each of stdout and stderr followed by `STDOUT_MARKER` / `STDERR_MARKER`, on the **default** (non-opt-in) path. It asserts all 8,000 lines on each stream are present and that the output ends with the trailing marker, directly proving trailing output is no longer truncated.
- `RunAsync_ProcessTimeoutRemainsBounded` proves the timeout path still returns bounded (`TimedOut`, exit `-1`) within the grace window.
- `RunAsync_ExitsLeavingPipeHolder...` now passes `allowInheritedPipeHandleEscape: true`, proving the opt-in path stays bounded when a descendant holds an inherited pipe handle open.

This PR is intentionally left as a **draft** for a human to complete full-repo `./build.ps1` + Shared/Tray validation on a network-enabled host before merge.

### DIFF
--- a/src/OpenClaw.SetupEngine/CleanupStaleDistroStep.cs
+++ b/src/OpenClaw.SetupEngine/CleanupStaleDistroStep.cs
@@ -24,7 +24,12 @@ public sealed class CleanupStaleDistroStep : SetupStep
         if (!DistroInstallPathPolicy.TryGetManagedInstallPath(ctx.LocalDataDir, distro, out var wslDir, out var pathError))
             return StepResult.Terminal(pathError);
 
-        var list = await ctx.Commands.RunAsync(WslConstants.WslExePath, ["--list", "--quiet"], TimeSpan.FromSeconds(15), ct: ct);
+        var list = await ctx.Commands.RunAsync(
+            WslConstants.WslExePath,
+            ["--list", "--quiet"],
+            TimeSpan.FromSeconds(15),
+            ct: ct,
+            allowInheritedPipeHandleEscape: true);
         if (list.ExitCode != 0)
             return StepResult.Ok("WSL not available or no distros - nothing to clean");
 
--- a/src/OpenClaw.SetupEngine/CommandRunner.cs
+++ b/src/OpenClaw.SetupEngine/CommandRunner.cs
@@ -17,7 +17,8 @@ public interface ICommandRunner
         string? workingDirectory = null,
         string? stdinInput = null,
         CancellationToken ct = default,
-        Stream? stdinStream = null);
+        Stream? stdinStream = null,
+        bool allowInheritedPipeHandleEscape = false);
 
     /// <summary>
     /// Run a command inside a WSL distro.
@@ -86,7 +87,8 @@ public sealed class CommandRunner : ICommandRunner
         string? workingDirectory = null,
         string? stdinInput = null,
         CancellationToken ct = default,
-        Stream? stdinStream = null)
+        Stream? stdinStream = null,
+        bool allowInheritedPipeHandleEscape = false)
     {
         ArgumentNullException.ThrowIfNull(executable);
         ArgumentNullException.ThrowIfNull(arguments);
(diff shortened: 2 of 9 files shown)


shanselman:
Closing the obsolete conflicting output-drain branch. Reconfirm the issue against current main before creating a new narrow repair.

PR #1239, opened 2026-08-27
fix(inference,setup): shared-memory capacity semantics and destructive rollback (PR #1237 follow-ups)

larroy:
## Summary

Builds on #1237's NVML/DXGI PCI-identity join fix and addresses the two remaining issues that PR's description flagged but deliberately left unfixed:

1. **Shared memory wrongly counted as CUDA-allocatable capacity.** `LocalInferenceQualificationPolicy.GetEffectiveTotalMemoryBytes`/`GetEffectiveFreeMemoryBytes` used to add DXGI-reported shared memory directly into NVML dedicated memory. DXGI shared memory is a WDDM host-RAM budget the GPU may borrow for graphics use — it does not back CUDA device allocations. Counting it as capacity can make a model appear eligible when the GPU cannot actually load it. Eligibility/free-memory checks now use NVML dedicated memory only.

2. **Capability-verification failure destroyed unrelated completed work.** `SetupPipeline` rolled back every previously-completed step whenever any step failed with rollback enabled. A failed Local AI GPU/inference verification — which runs downstream of expensive model/runtime acquisition — would delete the just-downloaded multi-gigabyte GGUF and llama-server runtime. Adds `SetupStep.RollsBackPrecedingStepsOnFailure` (default `true`, preserving existing behavior everywhere else) and opts `VerifyLocalAiGpuLoadStep`/`VerifyLocalAiInferenceStep` out of it, since both are read-only verification steps whose failure shouldn't force re-acquisition of artifacts that are otherwise safely reused via each acquisition step's own `CreatedThisRun`-gated skip logic. The failed step itself is still rolled back regardless.

## Testing

- New/updated unit tests for both fixes, plus regression coverage for #1237's PCI-identity join (name-mismatch join, ambiguous-match rejection).
- Full suite: `OpenClaw.Shared.Tests` (3821 passed) and `OpenClaw.SetupEngine.Tests` (1002 passed, including the pipeline step-count/ordering lock-down test) — no regressions.
- Verified against real DGX Spark (GB10) hardware: `LocalInferenceEligibility.Evaluate` now reports dedicated-only capacity (47.94 GiB, down from the previous 59.38 GiB that wrongly included shared memory), and all catalog models remain `Eligible`.

### DIFF
--- a/src/OpenClaw.SetupEngine/LocalAiGpuVerification.cs
+++ b/src/OpenClaw.SetupEngine/LocalAiGpuVerification.cs
@@ -254,6 +254,11 @@ public sealed class VerifyLocalAiGpuLoadStep : SetupStep
     public override string DisplayName => "Verifying Local AI GPU placement";
     public override bool CanRetry => false;
     public override RetryPolicy Retry => RetryPolicy.None;
+
+    // Read-only verification downstream of expensive model/runtime acquisition.
+    // A failure here shouldn't force those artifacts to be re-downloaded.
+    public override bool RollsBackPrecedingStepsOnFailure => false;
+
     public override bool CanSkip(SetupContext ctx) => !ctx.Config.LocalAi.Enabled;
 
     public override async Task<StepResult> ExecuteAsync(SetupContext ctx, CancellationToken ct)
--- a/src/OpenClaw.SetupEngine/LocalAiSetupSteps.cs
+++ b/src/OpenClaw.SetupEngine/LocalAiSetupSteps.cs
@@ -758,6 +758,10 @@ public sealed class VerifyLocalAiInferenceStep : SetupStep
     public override bool CanRetry => false;
     public override RetryPolicy Retry => RetryPolicy.None;
 
+    // Read-only verification downstream of expensive model/runtime acquisition.
+    // A failure here shouldn't force those artifacts to be re-downloaded.
+    public override bool RollsBackPrecedingStepsOnFailure => false;
+
     public override bool CanSkip(SetupContext ctx) => !ctx.Config.LocalAi.Enabled;
 
     public override async Task<StepResult> ExecuteAsync(SetupContext ctx, CancellationToken ct)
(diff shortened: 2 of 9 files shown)


joelagnel:
This breaks UMA architectures. Lets not merge. The correct fix is likely to use Cuda API to query available Cuda memory (which will include the shared memory)

karkarl:
## Adversarial review consensus

**Recommendation: request changes.** PR #1239 can leave its managed `llama-server` running after verification failure, retaining GPU memory and the listening port and potentially blocking an immediate retry.

### Both models agree: HIGH consensus

| Issue | Opus | Codex | Fix confidence |
|---|---|---|---|
| Rollback opt-out at `SetupPipeline.cs:223-235` preserves the running process along with file artifacts. Both verification steps have no failed-step cleanup, so `StartLocalAiRuntimeStep.RollbackAsync` at `LocalAiSetupSteps.cs:715-736` is bypassed. Dispose and clear the transaction-owned runtime while preserving runtime/model/manifest files. Add a regression test asserting runtime disposal and `ctx.LocalAiRuntime == null` while artifact rollback remains suppressed. | MEDIUM | HIGH | 95% |

### Only one model flagged: LOW consensus

| Issue | Opus | Codex | Fix confidence |
|---|---|---|---|
| `VerifyLocalAiInferenceStep` returns from the invalid post-refresh state at `LocalAiSetupSteps.cs:818-824` without resetting the router, unlike its exception paths at `799-815`. Reset before returning failure and test that `RestartAsync` is invoked. | — | MEDIUM | 99% |
| Cancellation still performs full rollback at `SetupPipeline.cs:305-315`, deleting resumable artifacts. This may be intentional cancellation semantics and needs a policy regression test. | LOW | — | 60% (disputed) |

The NVML dedicated-memory eligibility change, DXGI shared-memory exclusion, PCI packing/subsystem matching, ambiguity handling, native struct layout, and fallback behavior were validated as correct and fail-closed. No concrete security vulnerability was found.

shanselman:
Closing the current implementation. Its dedicated-memory direction is superseded and rollback can leave llama-server running; retain those requirements for a safer replacement.

Issue #1242, opened 2026-08-27
Local AI setup intermittently fails (~30%): llama-server CUDA crash (cudaErrorSharedObjectInitFailed in flash-attn) on RTX Spark N1X during warm-up


larroy:
> **Updated 2026-08-31** after direct reproduction testing. The crash is **intermittent (~30% of identical runs)**, not deterministic, and is **present on the latest build tested (b10655)** — it was never fixed. Earlier "it works now" observations were sampling luck. `--flash-attn off` is **not** a usable workaround at the shipped context size. See [Reproducing this (for agents)](#reproducing-this-for-agents) for a scripted repro.

## Summary

Local AI setup fails at the `verify-local-ai-inference` step with `HTTP 500 (InternalServerError)`. The root cause is a **CUDA driver fault inside llama-server's flash-attention kernel during model warm-up**, not a setup-engine or VRAM-capacity problem — GPU memory allocation for the model, KV cache, and compute buffers all succeed first.

The fault is **nondeterministic**: with every argument held constant it fires on roughly 3 of 10 launches. A green setup run is therefore *not* evidence the problem is gone.

## Hardware / environment

| | |
|---|---|
| GPU | NVIDIA RTX Spark N1X (5120-core Blackwell RTX GPU), compute capability 12.1 (sm_121) |
| NVIDIA driver | 616.55 (KMD 616.55, CUDA UMD 13.4) |
| System | ARM64-based PC (NVIDIA "Spark" class device) |
| OS | Windows 11 Enterprise, 10.0.28000 (Build 28000) |
| llama-server builds affected | b10488 (`9d77fa172`) **and** b10655 (`cb300598d`), Clang 20.1.8 for Windows arm64 |
| Models affected | Qwen3.6-35B-A3B-MTP UD-Q4_K_M (`qwen35moe`) **and** Qwen3.6-27B-MTP Q4_K_M (`qwen35`, dense) |
| Runtime flags of note | `--flash-attn on`, `--n-gpu-layers all`, `--ctx-size 262144`, `--cache-type-k f16`, `--cache-type-v f16` |

### Memory topology (this is important)

This is a **unified-memory** device: one physical pool, statically split between the GPU carveout and the OS.

| | |
|---|---|
| Installed physical memory | **65,536 MB** (64 GiB, single SK Hynix DIMM0) |
| GPU carveout (`nvidia-smi` FB total) | **44,480 MiB** |
[...]

PR #1252, opened 2026-08-27
chore(local-ai): bump llama.cpp runtime to b10655

larroy:
## Summary

Bumps the pinned llama-server runtime from `b10488` to `b10655` (`ggml-org/llama.cpp` commit `cb300598d`) and updates the pinned SHA-256 hashes/sizes for the x64 and arm64 CUDA runtime binaries. The `cudart` dependency archives are byte-identical between the two releases, so their hashes are unchanged.

## Motivation

Closes/addresses #1242: on the RTX Spark N1X (Blackwell, sm_121), `b10488` crashes with a fatal CUDA error (`cudaErrorSharedObjectInitFailed`) inside the flash-attention kernel's shared-memory setup during model warm-up. This surfaces during setup as `verify-local-ai-inference` failing with HTTP 500.

I reproduced the crash manually by running `llama-server.exe` directly with the same flags/model as the failing setup run, against both b10488 (crashes, confirming the report) and b10655 (does not crash — loading completes and a real chat-completion round-trips successfully). Full details and logs are in the issue and its comment thread.

Note: at the full `--ctx-size 262144` used in the original failing run, b10655 now gets further in loading but hits an unrelated CUDA OOM allocating the speculative/MTP draft buffers — plausibly related to the GPU-capacity/DXGI-shared-memory misdetection already being addressed elsewhere in this repo (`ba487943`, `05994b93`). That's a separate follow-up, not blocking this bump — it's an improvement either way, and worth re-checking against whatever context size the app's own capacity planner actually selects on this hardware.

## Changes

- `LlamaRuntimeCatalog`: `ReleaseTag` → `b10655`, `ReleaseCommitSha` → `cb300598d5f90189cb69d2702f4930aaf99d32a2`, runtime ids, artifact filenames/sizes/hashes for both x64 (CUDA 13.3) and arm64 (CUDA 13.4) variants.
- `LlamaRuntimeInstaller.ValidateVersionOutput`: was checking for the hardcoded literal `"build 10488"` instead of deriving it from `LlamaRuntimeCatalog.ReleaseTag`. Left as-is, this bump would have made every install permanently fail capability verification (the installed binary reports `build 10655`, which would never match the hardcoded check). Now derives the expected string from the catalog constant so future bumps don't hit the same trap.

## Testing

- `dotnet build` on `OpenClaw.SetupEngine` (which pulls in `OpenClaw.Shared`/`OpenClaw.Connection`): clean, 0 warnings/errors.
- `dotnet test tests/OpenClaw.Shared.Tests` (`LocalInferenceQualificationTests`): 21/21 passed.
- `dotnet test tests/OpenClaw.SetupEngine.Tests` (full suite): 998/998 passed.
- Manually verified all four new pinned artifacts against the actual `ggml-org/llama.cpp` b10655 release assets (downloaded fresh, hashed locally, cross-checked sizes against the GitHub release API).
- Manually ran the extracted b10655 `llama-server.exe` + this model end-to-end (outside the app) and confirmed a successful `/v1/chat/completions` response.

### DIFF
--- a/src/OpenClaw.SetupEngine/LlamaRuntimeInstaller.cs
+++ b/src/OpenClaw.SetupEngine/LlamaRuntimeInstaller.cs
@@ -277,7 +277,9 @@ internal sealed class WindowsLlamaRuntimeInspector : ILlamaRuntimeInspector
 
     internal static LlamaRuntimeInspection ValidateVersionOutput(string output)
     {
-        bool buildMatches = output.Contains("build 10488", StringComparison.OrdinalIgnoreCase);
+        bool buildMatches = output.Contains(
+            $"build {LlamaRuntimeCatalog.ReleaseTag[1..]}",
+            StringComparison.OrdinalIgnoreCase);
         bool commitMatches = output.Contains(
             LlamaRuntimeCatalog.ReleaseCommitSha[..9],
             StringComparison.OrdinalIgnoreCase);
@@ -286,7 +288,7 @@ internal sealed class WindowsLlamaRuntimeInspector : ILlamaRuntimeInspector
             : new LlamaRuntimeInspection(
                 false,
                 output,
-                "llama-server did not report the pinned b10488 build and source commit.");
+                $"llama-server did not report the pinned {LlamaRuntimeCatalog.ReleaseTag} build and source commit.");
     }
 
     private static void KillProcessTree(Process process)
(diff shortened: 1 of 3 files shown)


joelagnel:
LGTM. @karkarl could we merge this?

PR #1253, opened 2026-08-28
fix(local-ai): bound GPU admission by dedicated device memory

### LINKED ISSUE #1191: Do not count DXGI shared system memory as CUDA capacity on discrete GPUs
## Problem

Local AI qualification currently adds DXGI `SharedSystemMemory` to NVML-visible GPU memory for every NVIDIA adapter. This is unsafe for ordinary discrete GPUs.

DXGI defines `SharedSystemMemory` as the maximum amount of system memory the adapter may consume. It is not evidence that the memory is CUDA-visible VRAM suitable for the configured full-offload llama-server recipe. A discrete GPU can therefore be qualified for a model that it cannot fully offload, causing a multi-gigabyte download followed by startup or verification failure.

Microsoft contract: https://learn.microsoft.com/windows/win32/api/dxgi/ns-dxgi-dxgi_adapter_desc1

## Current behavior

`GetEffectiveTotalMemoryBytes` unconditionally adds positive shared memory to NVML-visible memory:

https://github.com/openclaw/openclaw-windows-node/blob/31dffdb0dbd9eebc8b98ce35cf13ab60762410a9/src/OpenClaw.Shared/Inference/Catalog/LocalInferenceSelector.cs#L144-L149

The current test explicitly expects shared memory to count for any NVIDIA GPU:

https://github.com/openclaw/openclaw-windows-node/blob/31dffdb0dbd9eebc8b98ce35cf13ab60762410a9/tests/OpenClaw.Shared.Tests/LocalInferenceQualificationTests.cs#L92-L108

## Expected behavior

- Use NVML-visible total and free memory by default.
- Count shared system memory only when unified-memory capability is positively established for that device.
- Preserve stable per-adapter identity when correlating NVML and DXGI evidence. Prefer an OS or driver identity such as LUID rather than display-name containment where practical.
- Add regression coverage for discrete, unified-memory, and multiple-GPU systems.

## Impact

A false-positive qualification can waste bandwidth and disk space, then fail only after artifact acquisition and runtime startup. This is especially likely on systems with substantial RAM and a discrete GPU below the model's actual VRAM requirement.

Raised by @larroy during the PR #1178 review follow-up.

joelagnel:
## Summary

- Query the CUDA driver directly for NVIDIA GPU identity, CUDA-major compatibility, and total/free memory.
- Bound Local AI admission capacity by a **driver-reported dedicated-memory source joined on device identity**, so a WDDM device cannot be qualified on memory that does not back device allocations: DXGI dedicated video memory joined by adapter LUID and NVML joined by GPU UUID, combined by taking the most conservative value.
- Take free memory from the adapter's local WDDM budget when available, so launch readiness is not inflated by the shared host pool.
- Preserve retryable partial GPU facts instead of collapsing a read failure into a definitive no-GPU verdict, and restrict `nvcuda.dll` loading to System32.

Closes #1191. Related investigation: #1237 and #1239.

The intended merge order places this before #1281 and #1290.

### Why capacity is capped rather than taken raw from CUDA

`cuMemGetInfo` is not a safe admission bound on WDDM. On the DGX Spark recorded in #1237, CUDA advertised **46,332 MiB** because it surfaces the WDDM shared host pool, while llama-server still died with `CUDA error: out of memory` at **16,186 MiB (15.81 GiB)** against roughly 15.9 GiB of real device memory. Shared system memory is host RAM and never backs a device allocation, so it is used for nothing here: the dedicated figure only ever *lowers* reported capacity, never raises it.

`capacity = min(cudaTotal, dedicated)`. No arbitrary constants: every bound is a driver or OS reported fact.

### Why there is no allocation-proven gate

An allocation gate was implemented and then **empirically disproved on real hardware at this head** (RTX 4080 SUPER, 16,375 MiB CUDA total, driver 620.02). WDDM reservations do not fail on oversubscription:

```
cuMemAlloc  32,751 MiB : Allocatable   (2x the reported total)
cuMemCreate 16,375 MiB : status=0 (ALLOCATED)
cuMemCreate 32,751 MiB : status=0 (ALLOCATED)
cuMemCreate 65,536 MiB : status=0 (ALLOCATED)   (4x the physical device)
```

Neither `cuMemAlloc` nor the VMM `cuMemCreate` path discriminates allocatable capacity on WDDM, so a reservation gate would have reported a false "proven" result while adding a multi-gigabyte allocation to setup. The conservative dedicated-memory bound is used instead, per the recommendation in #1237 that shared memory must be kept separate from the number eligibility is judged against.

### Non-DXGI hosts stay supported

A DXGI-only bound would contract support. A unified-memory DGX or Blackwell part that reports no dedicated video memory, a TCC device, or a headless GPU may have no usable DXGI adapter, and would resolve no bound at all.

Resolution order is therefore:

1. **DXGI** dedicated video memory, joined by adapter LUID (`cuDeviceGetLuid` against `DXGI_ADAPTER_DESC1.AdapterLuid`). Preferred, so the conservative WDDM budget behavior is unchanged wherever it applies.
2. **NVML** `nvmlDeviceGetMemoryInfo` total, joined by the same `GPU-...` UUID the CUDA driver reports. NVML reports dedicated device memory only, never the shared host pool, which is why it reported 16,320 MiB on the DGX Spark where CUDA advertised 46,332 MiB.

Both joins are on device identity, never on adapter name. `nvml.dll` is loaded only from fully qualified driver-owned paths, matching the System32 pinning used for `nvcuda.dll` and `dxgi.dll`.

### Fail-closed behavior

When neither source resolves a bound, `GpuVisibleMemoryBytes` stays null. That yields the retryable `HardwareFactsIncomplete` state rather than an over-qualification, so the worst case is "Local AI unavailable, recheck available" instead of a large download that cannot start.

## Required proof pools

- `windows-11-arm64`: native ARM64 CUDA detection and app build behavior changed.
- `windows-wsl-dgx-blackwell`: NVIDIA DGX/Blackwell detection, Local AI setup, and inference qualification changed.
- `windows-winui-interactive`: onboarding eligibility and Local AI presentation depend on the new hardware facts.

## Validation

Current head: `fb0260599dcc287411354fc33c156acf2c8f2ce5`

Windows 11 x64, .NET SDK 10.0.303, NVIDIA driver 620.02:

- `.\build.ps1`: passed. Documentation gate (48 Markdown files, 7 proof pools) plus Shared, Cli, WinNodeCli, SetupEngine, and WinUI all built successfully.
- `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj`: **3,904 passed**, 32 environment-gated skips.
- `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj`: **2,829 passed**.
- `dotnet test .\tests\OpenClaw.SetupEngine.Tests\OpenClaw.SetupEngine.Tests.csproj`: **1,028 passed**.
- `dotnet test .\tests\OpenClaw.Connection.Tests\OpenClaw.Connection.Tests.csproj`: **706 passed**.
- Focused `LocalInferenceQualificationTests`: **50 passed**, including the DGX capacity regression, the CUDA-free-inflation regression, the missing/failing UUID regressions, and the no-LUID, no-DXGI, zero-dedicated, DXGI-preferred, identity-mismatch, and both-sources-failing fallback regressions.
- Hardware-gated `LocalInferenceHardwareProbeTests` with `OPENCLAW_RUN_GPU_PROOF=1`: **2 passed** on a real NVIDIA GPU. These skip automatically on hosts with no NVIDIA device, so GPU-less CI stays green.

CI on the previous head `03975b60` was green, including the aggregate `test` job and all three E2E suites.

Rubber-duck review was run before publication and raised three blocking issues, all fixed:

1. A failed `cuInit` still became a definitive `NoNvidiaGpu`. Driver availability is now tri-state, and only an absent driver or a driver reporting no device is definitive.
2. An inconclusive GPU could be hidden by a second, definitively unsupported GPU, showing a permanent verdict and disabling recheck. Inconclusive candidates now outrank definitive ones.
3. Clamping CUDA free memory did not make it trustworthy. The adapter's WDDM budget is now the authority for free memory.

That pass also surfaced two of my own defects, both fixed before publication: an incorrect `IDXGIAdapter3::QueryVideoMemoryInfo` vtable slot, and an over-strict free-memory rule that blanked out capacity on an ordinary discrete GPU whose CUDA total slightly exceeds the DXGI dedicated figure.

## Real behavior proof

Standalone current-head hardware proof, collected on this branch (not from integration SHA `38a172f1`). Windows 11 x64, NVIDIA GeForce RTX 4080 SUPER, driver 620.02, compute capability 8.9, redacted GPU UUID.

### Default path

```
== CudaHostHardwareProbe (current head) ==
os.arch                 : X64
gpus.detected           : 1
gpus.nvidia             : 1
  vendor                : Nvidia
  name                  : NVIDIA GeForce RTX 4080 SUPER
  stableId              : GPU-3c102124-<redacted>
  cudaMajorVersion      : 13
  admissionCapacity     : 16,045 MiB
  freeWithinCapacity    : 15,277 MiB
  sharedGpuMemory       : <null> (never counted)

== LocalInferenceEligibility (current head) ==
status                  : Unsupported
failureCode             : InsufficientGpuMemory
selectionFailureCode    : None
plan.model              : qwen3.5-9b-mtp-q4-k-m
plan.runtime            : b10655-cuda13-x64
requiredTotalMemory     : 21,980 MiB
detectedTotalMemory     : 16,045 MiB
availableFreeMemory     : 15,277 MiB
```

### Non-DXGI fallback path, proven on the same hardware at this head

`LocalInferenceHardwareProbeTests` forces the DXGI source to supply nothing, so resolution must fall back to the NVML identity join exactly as it would on a host DXGI cannot describe:

```
nvml GPU-3c102124-<redacted> dedicated=16,376 MiB free=8,913 MiB

dxgi  capacity=16,045 MiB free=15,277 MiB
nvml  capacity=16,375 MiB free=8,889 MiB
dxgi  Unsupported/InsufficientGpuMemory detected=16,045 MiB
nvml  Unsupported/InsufficientGpuMemory detected=16,375 MiB
```

`nvidia-smi` on the same host reports `16376 MiB` total and `8900 MiB` free, so the NVML dedicated figure matches the driver exactly.

What this demonstrates on real hardware at this head:

- **The non-DXGI path keeps the device supported.** With DXGI supplying nothing, capacity resolves to 16,375 MiB rather than blanking out, and the failure code is not `HardwareFactsIncomplete`.
- **Both paths reach the same qualification decision** (`Unsupported/InsufficientGpuMemory`), so the fallback does not change the outcome on a host where both sources exist.
- **The identity join is exact.** NVML keys on `GPU-3c102124-...`, the same UUID `cuDeviceGetUuid_v2` produced.
- **The LUID join is exact.** DXGI reported adapter LUID `00000000:00018980` and `cuDeviceGetLuid` returned the identical value.
- **The cap is active and conservative.** CUDA reported 16,375 MiB total; the DXGI path admits 16,045 MiB.
- **The shared pool is excluded.** DXGI reported `SharedSystemMemory` of 104,673 MiB on this adapter and none of it reaches qualification. Under the pre-#1191 accounting this device would have been credited with well over 100 GiB.
- **The stable `GPU-...` manifest identity is unchanged**, so existing installs keep their GPU selector.

The reservation results quoted in the Summary were captured from the same host and head.

**Not verified / blocked:** full model download, llama-server launch, GPU-placement verification, and fixed-prompt inference were not run at this head. This 16 GiB device correctly qualifies for no catalog model (the smallest requires 21,980 MiB), so end-to-end inference proof needs the `windows-wsl-dgx-blackwell` pool. DGX/Blackwell capacity behavior is covered here by the recorded #1237 numbers replayed as unit regressions plus the non-DXGI fallback proven on the hardware available, not by a live DGX run.

## Security and compatibility

- All `nvcuda.dll` imports are restricted to `DllImportSearchPath.System32`, enforced by a test that scans every nvcuda import in the assembly rather than one type.
- `dxgi.dll` is likewise pinned to System32, and `nvml.dll` is loaded only from fully qualified driver-owned paths, enforced by a test.
- CUDA UUID formatting preserves the existing `GPU-...` manifest identity, so GPU selection and upgrade behavior are stable.
- Capacity can only decrease relative to the pre-#1191 accounting, so no device becomes newly eligible.
- No permissions, tokens, or network surfaces are added.


src/OpenClaw.Shared/Inference/CudaHostHardwareProbe.cs
```diff
@@ -0,0 +1,134 @@
...
+            physicalMemory?.AvailableBytes,
+            gpus,
+            VulkanAvailable: false);
+    }
+
+    private static IReadOnlyList<GpuInfo> CaptureCudaGpus()
+    {
+        if (!OperatingSystem.IsWindows() || CuInit(0) != CudaSuccess ||
+            CuDeviceGetCount(out int count) != CudaSuccess)
+        {
```
larroy, line 115: we should surface an error to the user here if cuda is not found or can't be initialized

src/OpenClaw.Shared/Inference/CudaHostHardwareProbe.cs
```diff
@@ -0,0 +1,134 @@
...
+            return [];
+        }
+
+        int? cudaMajorVersion = CuDriverGetVersion(out int driverVersion) == CudaSuccess && driverVersion > 0
+            ? driverVersion / 1000
+            : null;
+        var gpus = new List<GpuInfo>();
+        for (int ordinal = 0; ordinal < count; ordinal++)
+        {
```
larroy: Suggest to refactor this in a more functional idiom:


```
private static IReadOnlyList<GpuInfo> CaptureCudaGpus()
{
    if (!OperatingSystem.IsWindows() || CuInit(0) != CudaSuccess ||
        CuDeviceGetCount(out int count) != CudaSuccess)
    {
        return [];
    }

    int? cudaMajorVersion =
        CuDriverGetVersion(out int driverVersion) == CudaSuccess && driverVersion > 0
            ? driverVersion / 1000
            : null;

    return Enumerable.Range(0, count)
        .Select(ordinal => TryCaptureGpu(ordinal, cudaMajorVersion))
        .Where(gpu => gpu is not null)
        .Select(gpu => gpu!)
        .ToList();
}

private static GpuInfo? TryCaptureGpu(int ordinal, int? cudaMajorVersion)
{
    if (CuDeviceGet(out int device, ordinal) != CudaSuccess)
        return null;

    string? name = ReadDeviceName(device);
    string? pciBusId = ReadPciBusId(device);
    if (string.IsNullOrWhiteSpace(name) || string.IsNullOrWhiteSpace(pciBusId))
        return null;

    return WithCudaContext(device, () =>
    {
        if (CuMemGetInfo(out nuint freeBytes, out nuint totalBytes) != CudaSuccess ||
            totalBytes == 0 || totalBytes > long.MaxValue || freeBytes > totalBytes)
        {
            return null;
        }

        return new GpuInfo(
            GpuVendor.Nvidia,
            name,
            GpuVisibleMemoryBytes: (long)totalBytes,
            FreeGpuVisibleMemoryBytes: (long)freeBytes,
            CudaMajorVersion: cudaMajorVersion,
            StableId: ToCudaVisibleDevicesSelector(pciBusId));
    });
}

private static GpuInfo? WithCudaContext(int device, Func<GpuInfo?> action)
{
    if (CuCtxCreate(out IntPtr context, 0, device) != CudaSuccess)
        return null;

    try
    {
        return action();
    }
    finally
    {
        _ = CuCtxDestroy(context);
    }
}
```

larroy (commented):
Approach looks good.

We should add a test if possible for the case where Cuda detection fails or can't be initialized.

karkarl:
## Adversarial review consensus

Reviewed the complete 18-file diff from base `f46400aab24e835d9f7339b3e94821260a42d3c0` to head `fa285061e37c4082d5815955d7a9e7d379bd4f64` independently with Claude Opus and GPT Codex, then cross-referenced and adjudicated their findings.

**Verdict:** one actionable **MEDIUM-risk** defect should be fixed before merge. No CRITICAL findings.

### Both models agree: HIGH consensus

| Issue | Opus | Codex | Fix confidence |
|---|---:|---:|---:|
| `CudaHostHardwareProbe.cs:23-24,47-62,116-123,165-166`: UUID capture failures erase detected NVIDIA devices. A CUDA error makes `TryCaptureGpu` return `null` and filter the device; missing `cuDeviceGetUuid_v2` throws through enumeration and the outer catch replaces the entire list with `[]`. `LocalInferenceSelector.cs:77-78` then emits definitive `NoNvidiaGpu`, bypassing the retryable `HardwareFactsIncomplete` UX. | LOW | HIGH | 95% |

**Adjudication:** practical severity is MEDIUM. On CUDA 13 drivers, UUID failure should be rare, so HIGH is aggressive. The behavior is nevertheless real and directly violates the PR's partial-facts and retry contract. On CUDA 9.2 through 11.3, the `_v2` entrypoint problem affects already-ineligible drivers, but gives a false `NoNvidiaGpu` diagnosis instead of an accurate compatibility result.

**Suggested fix:** isolate native failures per device, fall back to legacy `cuDeviceGetUuid` when `_v2` is unavailable, and preserve a partial NVIDIA fact for retryable failures. Add injected/native-seam tests for UUID non-success, missing `_v2`, and one failing device on a multi-GPU host.

### Only one model flagged: LOW consensus

| Issue | Opus | Codex | Fix confidence |
|---|---:|---:|---:|
| `LocalAiInstallReconciler.cs:70-73,181-184`: `GpuIdsMatch` treats the `cuda:` prefix and UUID hex case-sensitively. | Not flagged | MEDIUM | 98% (disputed) |
| `LocalInferenceEligibility.cs:10-17,44`, `LocalInferenceEligibilityDiagnostics.cs:45-65`, and both UI formatters retain `DriverTooOld`, minimum version `615.0`, and presentation branches after driver-version qualification was removed. | MEDIUM | Not flagged | 80% (disputed) |

- **GPU ID casing:** valid hardening, but no supported migration failure was demonstrated. Base NVML persisted NVIDIA's canonical `GPU-...` string, the CUDA formatter emits lowercase hex, and the known legacy test covers exact lowercase `cuda:`. This is non-blocking unless prerelease or hand-edited manifests are supported. `OrdinalIgnoreCase` for both prefix and UUID is a straightforward fix.
- **Stale driver path:** dead code, not a current user-visible defect. Removing the public property or enum outright could create compatibility risk. Any cleanup should retain the enum numeric value if serialized or external consumers exist.

Cross-checks found no DLL hijack path because all CUDA imports use `DllImportSearchPath.System32`; no substantiated ABI, calling-convention, integer-overflow, context-lifetime, UUID-format, WDDM/UMA, setup rollback, downgrade identity, or additional security defect.

Validation was review-only. No files were changed and no tests were run as part of this review.

joelagnel:
@clawsweeper re-review

joelagnel:
Karen: will look into test flakiness/test failure issues.

joelagnel:
Joel: screenshots , tests for proof.

joelagnel:
## Test and runtime proof

### Standalone PR head `79ef5c54ad5871f1611a0e4edf7218f02fed834d`

- `.\build.ps1`: passed on Windows 11 ARM64; Shared, CLI, WinNode CLI, SetupEngine, and WinUI built successfully.
- `OpenClaw.Shared.Tests`: 3,869 passed; 32 environment-gated skips.
- `OpenClaw.Tray.Tests`: 2,824 passed.
- `git diff --check origin/main...HEAD`: passed.

### Integrated Local AI validation at `38a172f12bf1ed591dc1a0f1169b3b100f3e53d5`

- Full build: passed.
- Shared: 3,869 passed; 32 environment-gated skips.
- Tray: 2,813 passed.
- SetupEngine: 1,043 passed.
- Connection: 713 passed.

The Windows 11 ARM64 RTX Spark run also recorded:

- Qwen 3.8 27B selected with 45.4 GiB CUDA-visible memory.
- Fixed-prompt inference completed successfully with 32 generated tokens.
- GPU placement verification passed with 66/66 layers and 15,738,286,899 bytes in CUDA model buffers.
- The complete setup pipeline finished successfully in 564.3 seconds.

Hosted CI's aggregate test job is currently red only on the same two capture-consent integration timeouts seen across the refreshed PRs: `CameraSnap_ReturnsImageOrWellFormedError` and `ScreenSnapshot_ReturnsImageOrWellFormedError`, each timing out after 30 seconds. This PR does not change camera, screen capture, consent, or MCP integration code.

joelagnel:
@clawsweeper re-review

src/OpenClaw.Shared/Inference/CudaHostHardwareProbe.cs
shanselman: Addressed in `03975b60`. CUDA availability is now tri-state (`CudaDriverAvailability`), so an initialization failure is no longer silently equivalent to "no NVIDIA GPU":

- `Absent` (no `nvcuda.dll`) or `NoDevice` (`CUDA_ERROR_NO_DEVICE`) are the only definitive no-GPU results.
- Any other `cuInit` failure, such as a driver/runtime mismatch, keeps a detected NVIDIA entry with its facts null, which surfaces as the retryable `HardwareFactsIncomplete` state.

`LocalAiPageViewModel` already routes `HardwareFactsIncomplete` to the same inconclusive path as a thrown probe failure, so the user sees a recheckable message instead of Local AI being permanently disabled. `PreflightLocalAiHardwareStep` reports the failure code in its terminal message.

This also covers the per-device case you were pointing at: name, UUID, LUID, and memory are read independently, and a failure in one no longer discards the others or the other devices. New regressions: `CudaProbe_DriverInitializationFailureStaysRetryableRatherThanNoGpu`, `CudaProbe_AbsentDriverIsDefinitiveNoNvidiaGpu`, `CudaProbe_DriverReportingNoDeviceIsDefinitiveNoNvidiaGpu`, plus the missing and failing UUID cases.

src/OpenClaw.Shared/Inference/CudaHostHardwareProbe.cs
shanselman: The functional shape you suggested landed in `79ef5c54` and is still the structure here: `CaptureCudaGpus` builds the list with `Enumerable.Range(...).Select(...)`, and the native interop moved into a dedicated `NvcudaDriver` type so both the identity read and the memory read go through one System32-pinned surface.

Two deliberate departures from the exact snippet, both required by the correctness fixes in `03975b60`:

- `TryCaptureGpu` no longer returns null and is not filtered out. Dropping a device on a failed name or UUID read was the bug that turned a partial read into a definitive `NoNvidiaGpu`, so every enumerated device now yields a `GpuInfo` with unavailable facts left null.
- The CUDA context wrapper is gone from this path. Memory is read through the `ICudaDeviceReader` seam, which is what makes the failure handling unit-testable without NVIDIA hardware.

`pciBusId` also stayed as the CUDA UUID rather than a PCI bus selector, to preserve the existing `GPU-...` manifest identity for installed devices.

shanselman:
## Both blockers addressed at `03975b60ac59d96641e46aa3c536e56eea90ee09`

### 1. WDDM admission bound (P1)

I took the "retain a conservative admission bound" option, because I **empirically disproved the allocation-gate option on real hardware at this head**. I implemented a gate first, then found WDDM reservations do not fail on oversubscription (RTX 4080 SUPER, 16,375 MiB CUDA total):

```
cuMemAlloc  32,751 MiB : Allocatable   (2x reported total)
cuMemCreate 16,375 MiB : status=0 (ALLOCATED)
cuMemCreate 32,751 MiB : status=0 (ALLOCATED)
cuMemCreate 65,536 MiB : status=0 (ALLOCATED)   (4x the physical device)
```

Neither `cuMemAlloc` nor the VMM `cuMemCreate` path discriminates allocatable capacity, so a reservation gate would have reported a false "proven" result while adding a multi-gigabyte allocation to setup. That approach was removed.

The bound now used is the adapter's DXGI `DedicatedVideoMemory`, joined to the CUDA device by adapter LUID (`cuDeviceGetLuid` against `DXGI_ADAPTER_DESC1.AdapterLuid`):

- `capacity = min(cudaTotal, dedicatedVideoMemory)`. The dedicated figure can only lower capacity, never raise it.
- Free memory comes from the adapter's local WDDM budget (`IDXGIAdapter3::QueryVideoMemoryInfo`) when available, so launch readiness is not inflated by the shared host pool either.
- `SharedSystemMemory` is read for nothing. It is host RAM and never backs a device allocation, which is exactly the conclusion in #1237.
- If no bound resolves (no LUID, no matching adapter, DXGI unavailable, multi-node device), capacity stays null, which yields the retryable `HardwareFactsIncomplete` state. It fails closed and cannot over-qualify.

Applied to the recorded DGX Spark numbers, CUDA's 46,332 MiB is capped to 16,320 MiB and **no catalog model qualifies**, matching #1237's own conclusion that with about 15.9 GiB usable nothing in the catalog fits. That is a unit regression (`Evaluate_DoesNotQualifyAnyModelOnTheRecordedDgxSparkCapacity`), replayed from the recorded numbers, not a live DGX run.

No arbitrary constants were introduced. Every bound is a driver or OS reported fact.

### 2. Partial NVIDIA facts on UUID or entry-point failure (P2)

Only an absent driver or a driver reporting `CUDA_ERROR_NO_DEVICE` is now a definitive `NoNvidiaGpu`. Everything else keeps the detected device with unavailable facts null, producing retryable `HardwareFactsIncomplete`:

- Driver availability is tri-state, so a failed `cuInit` (driver/runtime mismatch) no longer means "no GPU".
- A failed device count keeps a retryable NVIDIA entry.
- Capture is contained per device, so one bad device cannot discard the ones that read cleanly.
- Name, UUID, LUID, and memory are read independently, so one missing entry point cannot discard the other facts.

Regressions added for missing UUID, throwing UUID entry point, mixed healthy/failing multi-GPU, failed device count, failed device handle, and missing name.

### Also fixed from rubber-duck review

That review raised three blocking issues beyond the two above, all fixed here:

1. A failed `cuInit` still became definitive `NoNvidiaGpu` (folded into the tri-state fix above).
2. An inconclusive GPU could be **hidden by a second, definitively unsupported GPU**, so a 48 GiB unreadable adapter plus a 16 GiB adapter reported a permanent `InsufficientGpuMemory` and disabled recheck. Inconclusive candidates now outrank definitive ones.
3. Clamping CUDA free memory did not make it trustworthy; the WDDM budget is now the authority.

It also caught two defects of my own before publication: a wrong `IDXGIAdapter3::QueryVideoMemoryInfo` vtable slot, and an over-strict free-memory rule that blanked out capacity on an ordinary discrete GPU whose CUDA total slightly exceeds the DXGI dedicated figure.

### Validation and proof

Full details are in the PR body. Windows 11 x64, driver 620.02, at this head:

- `.\build.ps1` passed; Shared 3,894 passed / 32 skipped; Tray 2,824 passed; SetupEngine 1,028 passed; Connection 706 passed; focused qualification suite 42 passed.
- Standalone current-head hardware proof (**not** integration SHA `38a172f1`) on an RTX 4080 SUPER: exact LUID join (`00000000:00018980` on both sides), admission capacity 16,045 MiB capped down from CUDA's 16,375 MiB, 104,673 MiB shared pool excluded, free 15,277 MiB from the WDDM budget, stable `GPU-...` identity preserved.

**Not verified / blocked:** full model download, llama-server launch, and fixed-prompt inference were not run at this head. This 16 GiB device correctly qualifies for no catalog model, so end-to-end inference still needs `windows-wsl-dgx-blackwell`. One known limitation is called out in the PR body: a UMA adapter reporting zero `DedicatedVideoMemory`, or a TCC/headless device with no DXGI adapter, resolves no bound and stays retryable-incomplete. If a DGX run shows that, the follow-up is a UUID-joined NVML dedicated fallback, which reported 16,320 MiB correctly on that hardware. I did not add it without hardware to prove it.

I did not merge this. Intended order remains #1253, then #1281, then #1290.

shanselman:
@clawsweeper re-review

shanselman:
Maintainer decision: preserve supported non-DXGI CUDA hosts. We are not accepting a DXGI-only Local AI support contraction for DGX/Blackwell UMA, TCC, or headless devices. Keep the conservative DXGI/WDDM bound, add an allocation-safe identity-joined fallback or explicit allocation-proven gate for supported non-DXGI platforms, and prove it on exact-head hardware. Raw CUDA total and arbitrary capacity constants are not acceptable.

shanselman:
## Maintainer decision addressed: no DXGI-only support contraction

Head is now `415dc7bfafb7a9e59932112db496eaf4c126561e`.

### What changed

Capacity resolution is now a two-source, identity-joined chain rather than DXGI-only:

1. **DXGI** dedicated video memory joined by adapter LUID (`cuDeviceGetLuid` against `DXGI_ADAPTER_DESC1.AdapterLuid`). Still preferred, so the conservative WDDM budget behavior is unchanged wherever it applies.
2. **NVML** `nvmlDeviceGetMemoryInfo` total joined by the same `GPU-...` UUID the CUDA driver reports. Consulted when DXGI supplies no adapter or reports zero dedicated memory, which covers unified-memory DGX and Blackwell parts, TCC devices, and headless GPUs.

NVML reports dedicated device memory only and never the WDDM shared host pool, which is exactly why it reported 16,320 MiB on the DGX Spark in #1237 where CUDA advertised 46,332 MiB and llama-server died at 16,186 MiB. So the fallback is allocation-safe on the platform that motivated this work.

Constraints held: `capacity = min(cudaTotal, dedicated)` from a driver-reported source, no raw CUDA total, no invented constants, both joins on device identity rather than adapter name, and if neither source resolves the device stays retryable rather than over-qualified. `nvml.dll` loads only from fully qualified driver-owned paths, matching the System32 pinning on `nvcuda.dll` and `dxgi.dll`.

### Exact-head hardware proof of the non-DXGI path

`LocalInferenceHardwareProbeTests` forces the DXGI source to supply nothing so resolution must fall back to the NVML identity join, exactly as on a host DXGI cannot describe. On an RTX 4080 SUPER, driver 620.02:

```
nvml GPU-3c102124-<redacted> dedicated=16,376 MiB free=8,913 MiB

dxgi  capacity=16,045 MiB free=15,277 MiB
nvml  capacity=16,375 MiB free=8,889 MiB
dxgi  Unsupported/InsufficientGpuMemory detected=16,045 MiB
nvml  Unsupported/InsufficientGpuMemory detected=16,375 MiB
```

`nvidia-smi` reports `16376 MiB` total on the same host, so the NVML dedicated figure matches the driver exactly. The fallback keeps the device supported (capacity resolves rather than blanking out, and the result is not `HardwareFactsIncomplete`), and both paths reach the same qualification decision.

### Regressions added

No-LUID, no-DXGI-adapter, zero-dedicated-memory, DXGI-preferred-over-NVML, identity-mismatch (NVML entry for a different UUID must not be used), and both-sources-throwing. Plus the two hardware-gated proofs above under `OPENCLAW_RUN_INTEGRATION=1`.

### Validation at this head

- `.\build.ps1` passed.
- Shared **3,900 passed** / 34 gated skips; Tray **2,824 passed**; SetupEngine **1,028 passed**; Connection **706 passed**.
- Focused `LocalInferenceQualificationTests`: **48 passed**. Hardware-gated `LocalInferenceHardwareProbeTests`: **2 passed** on a real NVIDIA GPU.
- CI on the previous head `03975b60` was green, including the aggregate `test` job and all three E2E suites. CI is running on this head now.

### Still blocked

Full model download, llama-server launch, and fixed-prompt inference remain unverified at this head. The available GPU correctly qualifies for no catalog model (smallest requires 21,980 MiB), so end-to-end inference proof needs the `windows-wsl-dgx-blackwell` pool. DGX/Blackwell capacity behavior is covered here by the recorded #1237 numbers as unit regressions plus the non-DXGI fallback proven on available hardware, not by a live DGX run.

I have not merged this. Intended order remains #1253, then #1281, then #1290.

shanselman:
@clawsweeper re-review

shanselman:
## Second rubber-duck pass applied: head is now `8439f3d17a851604843e3fc7980aab5cd7a7fa85`

A second review of the NVML fallback found a real defect in how the two sources were combined, now fixed.

### Free-memory semantic collision (blocking, fixed)

DXGI reports this process's WDDM budget; NVML reports device-wide unallocated memory. Both were feeding `FreeGpuVisibleMemoryBytes`, which drives `EligibleButBusy` and can terminate setup. On the test host the two disagreed sharply (DXGI 15,277 MiB vs NVML 8,889 MiB), so the same GPU could be called ready or busy purely by which source resolved.

Fixed by combining rather than preferring: when both sources identify the device, the smaller dedicated bound and the smaller free reading win. The decision is now deterministic and conservative regardless of which sources a host exposes, and a device is never admitted on the larger of two bounds.

### Also fixed

- The NVML loader reference is now freed, and cleanup stays reachable when `nvmlShutdown` throws.
- The interop catch again covers `MarshalDirectiveException` and `SEHException`, matching the implementation this PR had deleted.
- The NVML remarks no longer over-claim general UMA coverage. The recorded 16,320 MiB result proves the Windows driver tested in #1237, not the device class in general. A device whose driver returns `NOT_SUPPORTED` for memory info contributes no bound and stays retryable rather than over-qualified, which is the intended fail-closed outcome. MIG compute instances are explicitly out of scope.
- The allowed-path test now asserts the exact two driver-owned roots instead of any absolute path.
- The hardware tests fail loudly instead of passing silently if integration is enabled on a host with no NVIDIA GPU, and the cross-source test compares capacity rather than requiring equal status, since the combined path is allowed to be strictly more conservative.

### Validation at this head

- `.\build.ps1` passed.
- Shared **3,902 passed** / 34 gated skips; Tray **2,824 passed**; SetupEngine **1,028 passed**; Connection **706 passed**.
- Focused qualification suite **50 passed**; hardware-gated suite **2 passed** on a real NVIDIA GPU (52 together).
- CI was green on `03975b60`; it is running on this head now.

Known remaining gap, unchanged: end-to-end model download, llama-server launch, and fixed-prompt inference still need the `windows-wsl-dgx-blackwell` pool. The available GPU correctly qualifies for no catalog model.

shanselman:
@clawsweeper re-review

shanselman:
## CI gating fix: head is now `8d1c9e35fa690fd5ddcebd8916251c9436923dd6`

The exact-head CI failure was mine and narrow. The two hardware proofs were gated only by `OPENCLAW_RUN_INTEGRATION`, so hosted GPU-less runners executed them and failed on the explicit "requires an NVIDIA GPU" assertions. No production logic was involved.

Gating is now on the device rather than the integration flag. `NvidiaHardwareFactAttribute` skips when no NVIDIA GPU is detected, so ordinary CI stays green, while an explicit hardware proof run with `OPENCLAW_RUN_GPU_PROOF=1` still requires the device and fails loudly if it is missing. That keeps the proof meaningful instead of silently passing on a GPU-less host.

### Validation at this head

- Shared with `OPENCLAW_RUN_INTEGRATION=1` and no GPU-proof flag: **3,936 passed, 0 skipped**, reproducing the CI configuration that failed.
- Hardware proof mode `OPENCLAW_RUN_GPU_PROOF=1`: **2 passed** on a real NVIDIA GPU.
- Earlier at `8439f3d1`: build.ps1 passed; Tray 2,824; SetupEngine 1,028; Connection 706; all three E2E suites green on that head, with only the aggregate `test` job red on these two tests.

Re-running CI on this head.

shanselman:
## Merged current `origin/main`: head is now `fb0260599dcc287411354fc33c156acf2c8f2ce5`

Regular merge, no rebase and no force. `origin/main` at `0b5d8606` (which includes #1302 and #1300) merged cleanly with **no conflicts**: that work touches capture consent, camera preflight, and accessibility fixtures, which are disjoint from the Local AI inference files in this PR.

### Validation at this exact head

- `.\build.ps1`: passed, including the documentation gate.
- Shared: **3,904 passed**, 32 environment-gated skips.
- Tray: **2,829 passed** (up 5 from the merged consent-gate tests).
- SetupEngine: **1,028 passed**.
- Connection: **706 passed**.
- Focused qualification plus hardware proofs with `OPENCLAW_RUN_GPU_PROOF=1`: **52 passed** on a real NVIDIA GPU.

The previous head's aggregate `test` job was cancelled and superseded; final CI is running on this merged head. All three E2E suites were already green on `8d1c9e35`.

Nothing else changed in this push: it is the merge plus the validation above.

PR #1234, opened 2026-08-26
chore: use MXC SDK 0.8.0

shanselman:
@karkarl please review this MXC executor upgrade.

## What Problem This Solves

OpenClaw shipped the older MXC 0.7 executor even though MXC 0.8 is now the supported npm release with native executor fixes and stronger fail-closed behavior.

## Why This Change Was Made

Upgrade the packaged `@microsoft/mxc-sdk` and shipped `wxc-exec.exe` to 0.8.0. OpenClaw intentionally retains the supported `0.7.0-alpha` policy contract so this PR does not also introduce schema 0.8 directional-network behavior changes.

## User Impact

Sandboxed `system.run` uses the MXC 0.8 native executor while preserving the existing OpenClaw policy and permission behavior.

## Evidence

- `npm ls @microsoft/mxc-sdk --depth=0` resolves `@microsoft/mxc-sdk@0.8.0`.
- The built tray `tools\mxc\x64\wxc-exec.exe` SHA-256 matches the npm 0.8.0 package binary: `6049C64723AF1173C3739DC6CD6B2F33F6C021BB2832C4216233CBA7F71AEE9A`.
- Structured autoreview was clean with no actionable findings.
- Rubber-duck review found no blocking issues and confirmed the build guard, binary hash, and live containment proof form a coherent evidence chain.

## Change Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs or instructions
- [x] Tests or validation
- [ ] Security hardening
- [x] Chore or infrastructure

## Scope

- [ ] Tray or WinUI UX
- [x] Windows node capability
- [ ] Local MCP or `winnode`
- [x] Gateway, connection, or pairing
- [ ] Setup or onboarding
- [ ] Permissions, privacy, or security
- [x] Tests, CI, or docs

## Validation

- `./build.ps1`: passed, all five projects built successfully.
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore`: 3,813 passed, 32 skipped, 0 failed.
- `dotnet test ./tests/OpenClaw.Tray.Tests/OpenClaw.Tray.Tests.csproj --no-restore`: 2,724 passed, 0 failed.
- `./scripts/validate-mxc-e2e.ps1 -NoBuild`: passed all required non-skipping MXC proofs.

## Real Behavior Proof

- Environment tested: Windows x64, Windows SDK 10.0.26100, local WSL Gateway.
[...]

PR #1204, opened 2026-08-22
Clarify unavailable Local AI setup states

karkarl:
## Summary

- show Local AI availability as a green Local AI icon badge beside **Recommended** on the local WSL gateway choice
- place unavailable Local AI guidance above dependent setup content and keep **See why** independently usable
- use the native InfoBar action slot so **See why** stays on the message line at the far right with standard Fluent padding
- center the Local AI Hub page at wide window sizes using the same finite-width Grid and 900 px content-column pattern as Settings, Connection, Skills, Voice, and Sessions
- gray and disable unavailable setup choices while preserving applicable controls for an already installed managed runtime
- hide Local AI-only progress groups when Local AI is unavailable and keep the visible WSL installation stage active
- share eligibility diagnostics between setup and the Hub, with localized copy and focused accessibility/state tests

## Validation

Validated at PR head `1f21dbd6` with `OPENCLAW_REPO_ROOT` set to this worktree:

- `.\build.ps1` - passed (Shared, CLI, WinNode CLI, SetupEngine, WinUI)
- `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore` - passed: 3,817, skipped: 32, failed: 0
- `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore` - passed: 2,750, failed: 0
- `dotnet test .\tests\OpenClaw.Tray.UITests\OpenClaw.Tray.UITests.csproj -c Debug -r win-arm64 --no-restore` - passed: 134, failed: 0
- focused `LocalAiSetupUxContractTests` - passed: 3, failed: 0
- one unrelated Shared telemetry timing test initially timed out, passed immediately in isolation, and the full required suite then passed
- final rubber-duck review found no correctness, accessibility, localization, layout, or regression issues

## Real behavior proof

Current PR head `1f21dbd6` is running directly on the Local AI page in an isolated data directory, bypassing the setup wizard:

```text
OPENCLAW_TRAY_DATA_DIR=C:\Users\biilai\AppData\Local\Temp\OpenClawTray\karkarl-probable-garbanzo-karkarl-refine-local-ai-setup-ux-137bb16e
OpenClaw.Tray.WinUI.exe openclaw://hub/local-ai --post-setup-restart
PID: 36872
Window: OpenClaw Companion
Responding: True
Local AI router startup state: NotInstalled
```

[...]

PR #1203, opened 2026-08-21
fix(setup): gate Local AI on WSL readiness

karkarl:
## Summary

- merge current `main` at `4c078f9ee5f9308744b2180a41bad361ceb0371d`, which includes prerequisite PR #1202
- preserve #1202's known-state, one-shot elevated recovery for missing or uninitialized WSL
- keep the read-only WSL readiness gate before Capabilities can present Local AI choices
- keep `WslViabilityInspector` as the single classifier and `WslViabilityProbe` as the single UI cache and refresh owner
- keep WSL platform failures out of Local AI hardware and networking availability aggregation

Adapted from the ordering intent in joelagnel/openclaw-windows-node#4 for the post-#1178 setup architecture.

## Validation

Current head: `63de5640b74f1247f799352474f31b745187dacc`

- `.\build.ps1` passed. Shared, CLI, WinNode CLI, SetupEngine, and WinUI built successfully. 46 Markdown files validated.
- `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore` passed: 3,819 passed, 32 skipped, 0 failed.
- `dotnet test .\tests\OpenClaw.SetupEngine.Tests\OpenClaw.SetupEngine.Tests.csproj --no-restore` passed: 1,025 passed, 0 failed.
- focused combined #1202/#1203 SetupEngine matrix passed: 32 passed, 0 failed. Coverage includes disabled virtualization, missing WSL, optional-component and uninitialized WSL, ready WSL, one-shot install and reinspection, cancellation, reboot guidance, probe refresh and concurrency, existing-config classification, pipeline ordering, Local AI availability aggregation, and installed-runtime behavior in the full SetupEngine suite.
- focused Tray setup and UI contracts passed: 6 passed, 0 failed. Coverage includes readiness before Local AI decisions, fresh retry inspection, existing-config ordering, progress ordering, and Capabilities ownership.
- hosted CI passed on the current head. The first network-recovery attempt timed out before either test body while runner `netsh.exe` captured dynamic TCP port ranges. The isolated job rerun passed.
- `git diff --check` passed.

Full local Tray validation is blocked by a known host testhost hang. The exact full Tray command was stopped after 24 minutes without a test result. Focused changed-path Tray tests pass, and hosted CI passed the full test job.

## Real behavior proof

Current-head computer-use proof on Windows x64 used an isolated Debug build:

[...]

PR #1111, opened 2026-08-06
fix(chat): preserve persisted tool activity details

TheAngryPit:
Related: #1092

## What Problem This Solves

Fixes an issue where users reopening persisted chat history could lose important tool activity details: failed tool calls could be hidden inside a successful activity group, tool results using `callId` could lose their correlation, message-level errors could be ignored, and JSON-string-encoded tool arguments could render without their safe display fields.

## Why This Change Was Made

The history parser now accepts the persisted aliases and shapes already emitted by the Gateway, while the existing `NativeToolProjector` remains authoritative for decoding and safe argument projection. Failed tools remain individually visible; only neighboring successful tools are compacted into activity groups. This is a narrow follow-up to the compact tool activity work in #1092.

## User Impact

Users can reopen older conversations and still inspect failed tools, correlated results, and safe command arguments without losing the compact presentation for successful activity.

## Evidence

- Failed-first regressions reproduce and cover persisted `callId`, message-level errors, JSON-string arguments, result correlation, and failure presentation.
- Fresh exact-head build passed for Shared, CLI, WinNodeCLI, SetupEngine, and WinUI.
- Fresh exact-head Shared suite: 3,632 passed, 0 failed, 32 environment-only skips.
- Fresh exact-head Tray suite: 2,250 passed, 0 failed.
- All required GitHub test, build, E2E, hygiene, and CodeQL gates are green on the current head.
- Exact-head persisted Gateway-to-WinUI replay was exercised twice: initially and again after terminating/relaunching the disposable app without rewriting history.

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs or instructions
- [x] Tests or validation
- [ ] Security hardening
- [ ] Chore or infrastructure

## Scope

- [x] Tray or WinUI UX
- [ ] Windows node capability
- [ ] Local MCP or `winnode`
- [x] Gateway, connection, or pairing
- [ ] Setup or onboarding
- [ ] Permissions, privacy, or security
- [x] Tests, CI, or docs

## Validation

- `./build.ps1` - passed all 5 projects on exact head `e05d86c5151bae9e6aa286231975c99616c577ce`
- `dotnet test ./tests/OpenClaw.Shared.Tests/OpenClaw.Shared.Tests.csproj --no-restore` - 3,632 passed, 0 failed, 32 skipped
[...]

PR #1263, opened 2026-08-28
fix(local-ai): clarify unavailable Local AI setup states (clean replacement for #1204)

shanselman:
## Summary

Clean replacement for #1204 (`karkarl-refine-local-ai-setup-ux`), built directly on current `main`. That branch's history mixes unrelated, already-landed commits with the real contribution, so this PR recomputes the true diff (merge-base `911f3686` to PR head `568db5c1`) and rebases it onto today's `main` instead of rewriting/force-pushing the original branch. Rebased twice more since as `main` kept advancing on the same files (`#1203`, `#1254`, and others); each rebase was validated fresh.

- share Local AI eligibility diagnostics between setup and the Hub via a new `LocalAiSetupAvailabilityCoordinator` and locale-neutral `LocalInferenceEligibilityDiagnostics`
- place unavailable Local AI guidance above dependent setup content; the native `InfoBar.ActionButton` keeps **See why** independently usable and on the message line
- show a green Local AI availability badge beside **Recommended** on the local WSL gateway choice
- center the Local AI Hub page at wide window sizes using the same finite-width Grid/900px pattern as other Hub pages
- gray/disable unavailable setup choices while preserving controls for an already installed managed runtime
- hide Local AI-only progress groups when Local AI is unavailable, keeping the WSL installation stage active
- localize all new setup/Hub copy (en-us, fr-fr, nl-nl, zh-cn, zh-tw) — see below
- gate Local AI availability on device-level eligibility, not the selected model, and never dead-end step 3 — see "Round 2" below

## What changed vs. #1204 (retained vs. discarded)

**Retained (all 24 files, ~1.6k LOC):** the entire real contribution from #1204 applies here, nothing was dropped as "already landed" or unrelated. Verified by grepping current `main` for the PR's new symbols (`LocalAiSetupAvailabilityCoordinator`, `IsLocalAiOnlyGroup`, `LocalInferenceEligibilityDiagnostics`, `LocalAiUnavailablePanel`) before starting: none existed, confirming every touched file carries a real, non-trivial diff against `main`.

**Discarded:** only the stale, already-merged commits that #1204's branch carried from its outdated base (e.g. #1111, #1226, #1229, #1210, #1230, #1234, already on `main` independently). None of that reappears here.

**Fixed beyond #1204's last pushed head**, all covered by focused tests:

1. **Delayed-dispatch race in the Hub view model.** A queued UI-thread update could be silently dropped if `DispatcherQueue.TryEnqueue` genuinely deferred the callback; the result is now applied and the probe's cancellation released together, gated by one currency check.
2. **Setup toggle clobbered a configured selection while checking.** The toggle now mirrors the current config value during a check instead of forcing off; only a completed probe changes the selection.
3. **Primary button could stay stuck disabled** after a definitive unsupported result (`ShowLocalAiUnavailable` was missing the `UpdatePrimaryButtonState()` call its siblings had).
4. **Availability tied to the wrong model.** The Hub's device-availability probe now evaluates the best catalog model this hardware can run instead of the currently selected/installed model, so a stale/removed model no longer blocks Retry Setup.
5. **Incomplete hardware facts treated as permanent.** A partial/transient NVML read is now routed through the same retryable/probe-error path as a thrown probe failure, in both the setup page and the Hub.
6. **Change Model bypassed the availability gate** (`CanChangeModel`, added independently by `main`'s #1254 while this branch was being rebased, didn't check `IsSetupAvailable` like `CanRetrySetup` does).
7. **Setup toggle also clobbered a configured selection on a retryable probe *failure*** (`ShowLocalAiProbeUnknown`, the sibling of fix #2 above): a transient failure unconditionally cleared `LocalAi.Enabled`, so a successful recheck afterward initialized from the already-cleared value instead of restoring the prior selection.
8. **All new availability copy was hardcoded English** (ClawSweeper review on this PR): the setup InfoBar title/message/action, the checking/probe-error state messages and accessibility help text, the detailed diagnostic reason shown in the setup dialog and the Hub tip, and the Welcome-page "Local AI available" badge/accessible name. See the dedicated section below.

Each fix has a dedicated regression test.

## Round 2 (external Codex + Opus review on this PR)

- **Device-level gating, not model-level (Codex HIGH).** `CapabilitiesPage`/`WelcomePage` now gate device availability on `LocalInferenceEligibility.Evaluate(hardware)` first, then reconcile the configured `SelectedModelId` separately: a model that no longer exists in the catalog, or exists but this specific hardware cannot install at all (`!CanInstall`, i.e. truly `Unsupported`, not merely `EligibleButBusy`), falls back to the recommended/device-eligible model instead of leaving setup stuck on a stale or hardware-incompatible selection. `WelcomePage`'s badge is evaluated device-only (no model parameter).
- **Remaining hardcoded `.wslconfig` failure reason localized (Codex MEDIUM)** through `SetupLocalization` across all 5 locales.
- **Availability-probe CTS could leak or, on a worker-thread path, race a UI-thread cancel (Codex LOW → hardened further by review).** `LocalAiPageViewModel` now guarantees the probe's `CancellationTokenSource` is released/disposed exactly once on every path (dispatched apply, non-enqueued, rejected-enqueue, or already-disposed/inactive), using `Interlocked.Exchange`/`CompareExchange` so a worker thread releasing the token can never race the UI thread's `CancelAvailabilityRefresh()` into a double-cancel/double-dispose. `RecheckAvailability()` also now enforces its own `CanRecheckAvailability` gate instead of only the weaker "not already in flight" check.
- **Step-3 dead end when Local AI is enabled and availability is Checking/ProbeUnknown (Opus HIGH).** Both states now keep the toggle itself genuinely clickable (including restoring `LocalAiOptionContent.IsHitTestVisible`, which a naive re-enable of just the toggle's own `IsEnabled` would not undo) so turning Local AI off is always an escape hatch. Continue's own gate is unchanged and never bypasses eligibility or an as-yet-undetermined WSL networking-consent decision merely because availability is still pending.
- **Hub InfoBar didn't distinguish definitive-unavailable from checking/probe-unknown (Opus MEDIUM x2).** The Hub now shows an explicit "checking" title/message/progress ring while a probe (initial or recheck) is in flight, and a distinct localized title/severity (Warning) for a retryable probe error vs. Informational for a definitive unavailable result — instead of one static title/message and silently hiding all chrome mid-check.
- **`SetupLocalization`/probe failure diagnostics now use `Trace.TraceWarning`, not `Debug.WriteLine` (Opus LOW),** since `Debug.WriteLine` is compiled out entirely in Release builds; `Trace` is not, and its default listener reaches DebugView/ETW.
- **Welcome badge accessible-name suffix made idempotent (Opus LOW):** the base name is captured once so repeated detections don't re-append the suffix, and a `LiveRegionChanged` automation event (with peer creation forced, since one may not exist yet) announces the change.

## Localization (fix #8)

`OpenClaw.Shared`'s `LocalInferenceEligibilityDiagnostics` now returns a `LocalInferenceUnavailableReason` (a kind enum plus plain facts: model name, required/detected GPU memory in GB, driver version) instead of a formatted English sentence, keeping Shared locale-neutral. Each UI owner localizes at its own boundary:

- **`OpenClaw.SetupEngine.UI`** has no `.resw` files of its own and cannot reference the Tray app project (which references it, so the reverse would be circular). It gets a small new `SetupLocalization` helper that resolves resource keys against the same merged app resource map the Tray app's `LocalizationHelper` already reads (the existing `x:Uid="Onboarding_LocalAi_RecheckAvailabilityButton"` binding relies on the same mechanism). `CapabilitiesPage.xaml.cs` now routes every checking/probe-error/unavailable title, message, help text, the "why unavailable" dialog, the `.wslconfig` inspection failure, and the Welcome-page badge/accessible name through it.
- **`LocalAiPageViewModel`** (source-linked into `OpenClaw.Tray.Tests` without a WinUI resource host) exposes the locale-neutral `LocalInferenceUnavailableReason` instead of a formatted string, so it stays testable without `LocalizationHelper`. `LocalAiPage.xaml.cs` (the View, which already has full WinUI/`LocalizationHelper` access) formats it into localized text, including the new checking/probe-unknown title distinction — matching the existing resource-key-on-ViewModel / format-in-View pattern already used for engine/model/gateway status.

Resource keys were added to all five supported locales (en-us, fr-fr, nl-nl, zh-cn, zh-tw), and localization regression coverage was added: contract tests proving the setup page's checking/probe-error/unavailable copy, the "why unavailable" dialog, the Welcome-page badge/accessible name, and the Hub's checking/probe-unknown titles all route through `SetupLocalization`/`LocalizationHelper`/`x:Uid` (not hardcoded English) with matching keys in every locale; a test proving the shared diagnostics stay fact-only while both UI owners localize the same reason kinds; and a source test proving `SetupLocalization` implements the same "Key.Property" → "Key/Property" XAML-resource fallback `LocalizationHelper` already uses.

## Owner boundary (why all these files are one cohesive change)

Everything here belongs to a single seam: "is Local AI available, and does the UI correctly reflect, escape, and localize that during checking/unsupported/error states." `LocalAiSetupAvailabilityCoordinator`, `LocalInferenceEligibilityDiagnostics`, and `SetupLocalization` are the shared policy/diagnostics/localization layer; `CapabilitiesPage`/`WelcomePage`/`ProgressPage` are the setup-time consumers; `LocalAiPage`/`LocalAiPageViewModel` are the Hub-time consumer; the `.resw` files are the copy for both; the test files are 1:1 with the production files they cover.

## Concurrency note

Fetched and diffed against #1204's live head repeatedly during this work; its branch was never touched, read, or force-pushed. #1204 should be closed/superseded only after this PR is reviewed, green, and ready.

## Validation

```
$env:OPENCLAW_REPO_ROOT = '<worktree path>'
.\build.ps1                                                                    # Shared, Cli, WinNodeCli, SetupEngine, WinUI all built (clean, 0 warnings/errors, including a from-scratch obj/bin rebuild of OpenClaw.SetupEngine.UI)
dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj         # 3819 passed, 0 failed, 32 skipped
dotnet test .\tests\OpenClaw.SetupEngine.Tests\OpenClaw.SetupEngine.Tests.csproj # 1025 passed, 0 failed
dotnet test .\tests\OpenClaw.Tray.Tests\...csproj --filter "FullyQualifiedName~LocalAiSetupUxContractTests|...LocalAiPageViewModelTests|...AppRefactorContractTests|...LocalizationValidationTests|...CapabilitiesPageLocalizationCoverageTests|...LocalizationHelperSourceTests"
                                                                                # 154 passed, 0 failed (every file this PR touches, run repeatedly for stability)
```

Full, untargeted `OpenClaw.Tray.Tests` run: still blocked in this environment (hangs with no test output after ~6 minutes; reproduces the same way on a clean pre-fix checkout, so it is a pre-existing environmental blocker in this sandbox, not a regression from this change). Recommend a maintainer/CI run of the full untargeted suite before merge.

## Real behavior proof

- **Setup wizard, fresh isolated data dir:** the WSL gateway choice on the Welcome page shows the "Local AI available" badge beside "Recommended" (accessibility-tree confirmed: `WelcomeInstallLocalGatewayChoice` -> "Recommended" + "Local AI available").
- **Local AI Hub, real hardware (this host has an NVIDIA GeForce RTX 4080 SUPER):** launched directly to `openclaw://hub/local-ai` in an isolated data dir with no prior config. `NotInstalled` state renders correctly, no spurious unavailable banner, "Retry setup or download" enabled, engine/model status all correct, confirming no regression on the common/available happy path.
- **Unavailable / retryable / probe-error / checking / selection-preservation / consent-gate states:** this validation host has real qualifying hardware and cannot naturally exercise the unsupported-device path, so those invariants (including the new checking-progress UI, the toggle escape hatch during pending states, and the networking-consent gate never being bypassed while pending) are proven by the automated `LocalAiPageViewModelTests`/`LocalAiSetupUxContractTests` with controlled fake hardware probes and dispatcher fakes.

## Review

`autoreview` (Codex, gpt-5.6-sol, high reasoning) was run to convergence across this round's fix cycles, each producing a real, verified bug (device/model eligibility gating, wslconfig localization, CTS ownership races across two hardening passes, the step-3 dead end plus its hit-test-visibility follow-up, the pending-state consent-bypass gap, a checking-notification ordering gap, and the model-eligibility reconciliation gap) that was fixed and covered by a new or updated regression test, then re-reviewed clean of that finding. One reviewer-reported finding (a claimed C# definite-assignment compile error for the `eligibility` local on the device-ineligible path) was investigated and rejected: `eligibility` is explicitly initialized to `null` at declaration, and a from-scratch `obj`/`bin` rebuild of `OpenClaw.SetupEngine.UI` compiles with 0 warnings/errors, confirming the finding does not reproduce.

Not verified / blocked: full untargeted Tray suite (environmental hang, see above); WSL install E2E (would require actually installing a WSL distro on a shared host, skipped as out of scope for this validation pass; the setup-flow code paths are covered by the existing `SetupStepsTests`/`WslGlobalConfigManagerTests`/`LocalAiSetupUxContractTests`).

Supersedes #1204.

### DIFF
--- a/src/OpenClaw.SetupEngine.UI/Pages/CapabilitiesPage.xaml
+++ b/src/OpenClaw.SetupEngine.UI/Pages/CapabilitiesPage.xaml
@@ -123,6 +123,7 @@
                                          Visibility="Collapsed">
                                     <InfoBar.ActionButton>
                                         <HyperlinkButton x:Name="LocalAiUnavailableDetailsButton"
+                                                         x:Uid="Onboarding_LocalAi_UnavailableDetailsButton"
                                                          AutomationProperties.AutomationId="LocalAiUnavailableDetailsButton"
                                                          Content="See why"
                                                          Click="LocalAiUnavailableDetails_Click" />
--- a/src/OpenClaw.SetupEngine.UI/Pages/CapabilitiesPage.xaml
+++ b/src/OpenClaw.SetupEngine.UI/Pages/CapabilitiesPage.xaml
@@ -114,6 +114,36 @@
                                         <TextBlock Grid.Column="2" Text="Several GB" Style="{StaticResource CaptionTextBlockStyle}" Foreground="{ThemeResource TextFillColorTertiaryBrush}" FontFamily="Cascadia Mono, Consolas" VerticalAlignment="Center" />
                                     </Grid>
                                 </Border>
+                                <InfoBar x:Name="LocalAiUnavailablePanel"
+                                         AutomationProperties.AutomationId="LocalAiUnavailablePanel"
+                                         AutomationProperties.LiveSetting="Polite"
+                                         IsOpen="True" IsClosable="False" Severity="Informational"
+                                         Title="Local AI is not available"
+                                         Message="This PC does not meet one or more Local AI requirements."
+                                         Visibility="Collapsed">
+                                    <InfoBar.ActionButton>
+                                        <HyperlinkButton x:Name="LocalAiUnavailableDetailsButton"
+                                                         AutomationProperties.AutomationId="LocalAiUnavailableDetailsButton"
+                                                         Content="See why"
+                                                         Click="LocalAiUnavailableDetails_Click" />
+                                    </InfoBar.ActionButton>
+                                </InfoBar>
+                                <StackPanel x:Name="LocalAiAvailabilityRecoveryPanel"
+                                            Orientation="Horizontal"
+                                            Spacing="8"
+                                            VerticalAlignment="Center"
+                                            Visibility="Collapsed">
+                                    <ProgressRing x:Name="LocalAiAvailabilityProgressRing"
+                                                  Width="16"
+                                                  Height="16"
+                                                  IsActive="False"
+                                                  AutomationProperties.AccessibilityView="Raw" />
+                                    <HyperlinkButton x:Name="LocalAiRecheckAvailabilityButton"
+                                                     x:Uid="Onboarding_LocalAi_RecheckAvailabilityButton"
+                                                     AutomationProperties.AutomationId="LocalAiRecheckAvailabilityButton"
+                                                     Content="Recheck"
+                                                     Click="LocalAiRecheckAvailability_Click" />
+                                </StackPanel>
                                 <Border x:Name="LocalAiInstallReviewCard"
                                         AutomationProperties.AutomationId="LocalAiInstallReviewCard"
                                         Visibility="Collapsed"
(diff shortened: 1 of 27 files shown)


shanselman:
Pushed a new commit (`50f6a84c` -> `8c7199fa`) addressing the external Codex + Opus review round on this PR:

- Device-level eligibility gating for setup's `CapabilitiesPage`/`WelcomePage` (Codex HIGH), including a follow-up hardening to also reconcile a selected model that exists in the catalog but is `Unsupported` for this specific hardware (not just missing from the catalog), while never reconciling away a merely `EligibleButBusy` selection.
- Remaining `.wslconfig` inspection failure reason localized (Codex MEDIUM).
- Availability-probe `CancellationTokenSource` ownership hardened across three review passes: guaranteed exactly-once disposal on every path, then made the worker-thread release paths (rejected-enqueue, disposed/inactive) atomically coordinate with the UI-thread `CancelAvailabilityRefresh()` via `Interlocked.Exchange`/`CompareExchange` instead of racing it (Codex LOW -> hardened).
- `RecheckAvailability()` now enforces its own `CanRecheckAvailability` gate.
- Step-3 dead end when Local AI is enabled and availability is Checking/ProbeUnknown, fixed via a toggle escape hatch (not by ever letting Continue bypass eligibility or an undetermined WSL networking-consent decision) -- including a follow-up fix restoring `LocalAiOptionContent.IsHitTestVisible` so the re-enabled toggle is actually clickable, not just flagged enabled underneath a non-hit-testable container (Opus HIGH -> hardened).
- Hub InfoBar now distinguishes definitive-unavailable from checking/probe-unknown with distinct localized titles/severity, and shows explicit checking/recheck progress (progress ring + title/message) instead of silently hiding all chrome mid-probe; a follow-up fix also moved `StartAvailabilityRefresh()`'s notification to after the new probe token is assigned so a UI bound to `PropertyChanged` reliably observes `IsCheckingAvailability` the moment a recheck starts (Opus MEDIUM x2 -> hardened).
- `Debug.WriteLine` (compiled out in Release) replaced with `Trace.TraceWarning` for probe/localization failure diagnostics (Opus LOW).
- Welcome badge accessible-name suffix made idempotent across repeated detections, with a `LiveRegionChanged` announcement (forcing peer creation, since one may not exist yet) (Opus LOW).

Full details and the updated validation/proof are in the PR description above.

Requesting fresh review (ClawSweeper/Codex/Opus) on `8c7199fa`.

PR #1281, opened 2026-09-01
fix(local-ai): recommend Qwen3.8 with measured context profiles

### LINKED ISSUE #1280: Clarify recommendations when stronger models require reduced context
## Problem

Local AI setup can recommend the newest, highest-priority model by reducing its context length, even when an older or smaller model could run with a much larger context. That may be the right product choice, but the tradeoff needs to be explicit and intentional.

With the measured-conservative memory profiles currently being tested:

| CUDA-visible memory | Recommended model | Context / KV | Notable alternative |
|---:|---|---|---|
| 32 GiB | Qwen3.8 27B | 192K / Q8_0 | Qwen3.5 9B can run at 256K / F16 |
| 24 GiB | Qwen3.8 27B | 64K / F16 | Qwen3.5 9B can run at 256K / F16 |
| 16 GiB | Qwen3.5 9B | 128K / F16 | No other catalog model fits the complete recipe |

The 24 GiB case is the clearest example: choosing Qwen3.8 27B at 64K prioritizes model capability and version over the 9B model's 256K context. OpenClaw agent workloads can begin with a large system prompt and accumulate tool results, conversation history, and sub-agent context, so 64K versus 256K can materially affect runtime behavior.

## Clarification requested

- Confirm the intended recommendation priority between model capability or version, context capacity, and KV precision.
- Define the minimum context at which a model may be labeled `Recommended` for the agent harness.
- Decide whether setup should explain when a stronger model is recommended by trading away context.
- Consider separate presets such as best model quality and largest agent context if one global recommendation is misleading.
- Validate the final policy on real 24 GiB and 16 GiB GPUs, not only by constraining a 32 GiB device.

## Current implementation direction

The current selector evaluates complete, explicit recipes at 256K, 192K, 128K, and 64K with F16 then Q8_0 KV at each context. It prioritizes model version or capability first, then the largest context that fits, then the highest KV precision. Setup displays the chosen context and KV precision, but the broader product tradeoff should be agreed and documented.

joelagnel:
## Summary  - Recommend models by explicit generation priority, then select the largest fitting context and highest KV precision. - Add measured-conservative Qwen3.8 27B context profiles with F16 and Q8_0 KV fallbacks. - Persist the selected profile through setup and launch llama-server with its context and KV-cache settings. - Retire Qwen3.5 9B from new installs and reject 16 GiB GPUs, while keeping existing managed 9B installations launchable. - Show true binary GiB values.  Default capacity examples:  - RTX 5090 class (34,190,458,880 CUDA-visible bytes): Qwen3.8 27B, 192K, Q8_0. - 24 GiB: Qwen3.8 27B, 64K, F16. - 16 GiB: unsupported. Qwen3.5 9B is no longer selectable.  Fixes #1280.  ## Release blocker fixes (`74b008fe`)  Two release blockers found in the earlier head `28092144` are fixed in this branch.  **1. Implicit consent for the user-wide WSL change (security).** `28092144` deleted the affirmative consent checkbox and set `config.LocalAi.WslMirroredNetworkingConsent = config.LocalAi.Enabled`, so merely enabling Local AI silently authorized a user-wide `.wslconfig` mirrored-networking rewrite and a global `wsl.exe --shutdown` that stops every running distribution. The checkbox, the `_localAiNetworkingConsentRequired` gate, the reset-on-state-change helper, and the Continue gate are restored. Enabling Local AI alone no longer grants consent, and Continue stays blocked until the box is checked whenever the change is actually required. The entire consent surface in `CapabilitiesPage.xaml.cs` is byte-identical to the pre-regression state.  **2. Existing Qwen3.5 9B installs stranded on upgrade (compatibility).** `28092144` deleted the 9B entry outright. Because the launch path resolves `LocalModelCatalog.Find(manifest.ModelCatalogId)` and throws when it returns null, every user with an existing managed 9B installation would have been unable to start Local AI after upgrading. The entry is now a hidden legacy catalog record: excluded from `Models`, `ExplicitAlternatives`, and `Default`, so it can never be offered, recommended, or newly installed, but resolvable through `FindInstalled` on the installed-receipt, launch, and display paths only. Its pins are reproduced exactly as installed, and it exposes only the native-context F16 profile a pre-profile install could have recorded, so nothing is remapped and a receipt that could never have existed still fails closed.  ## Merge with current `main` (`569ebeec`)  Regular merge of `origin/main` at `09bc834f`, which includes #1253 (`fix(local-ai): bound GPU admission by dedicated device memory`), #1300, and #1302. No rebase or force push. Three conflicts, resolved as follows:  - `CapabilitiesPage.xaml.cs`: comment-only conflict; took main's CUDA-read wording. - `LocalAiInstallReconciler.cs`: kept **both** sides. This branch's five profile receipt checks (context plus the four KV precisions) and #1253's new `GpuIdsMatch` helper, which replaced the previous ordinal `SelectedGpuId` comparison. - `LocalInferenceQualificationTests.Evaluate_IgnoresLegacySharedMemoryFields`: the one semantic conflict. #1253 owns the memory semantics, so shared/unified memory stays ignored and the 8 GiB dedicated GPU is `Unsupported` with 8 GiB detected and free. #1281 owns the catalog, so the unsupported-fallback plan is the smallest *offered* model. `LocalInferenceSelector.SelectDefaultModelAndProfile` falls back to `Models.OrderBy(m => m.Weights.SizeBytes).First()`, and Qwen3.8 27B (16,464,440,224 bytes) is smaller than Qwen3.6 27B (17,106,773,120) and Qwen3.6 35B-A3B (22,663,387,424), so the expectation is Qwen3.8 27B.  This branch does not modify #1253's GPU-capacity contract.  ## Required proof pools  - `windows-11-arm64`: native ARM64 model selection and app launch behavior changed. - `windows-wsl-dgx-blackwell`: DGX/Blackwell model selection, setup, and fixed-prompt inference changed. - `windows-winui-interactive`: model details and the WSL consent prompt changed.  ## Validation  Current head: `8df3fb7c`  - `.\build.ps1`: passed, all projects and documentation validation succeeded. - `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj`: 3,905 passed, 32 environment-gated skips. - `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj`: 2,829 passed. - `dotnet test .\tests\OpenClaw.SetupEngine.Tests\OpenClaw.SetupEngine.Tests.csproj`: 1,030 passed. - `dotnet test .\tests\OpenClaw.Connection.Tests\OpenClaw.Connection.Tests.csproj`: 709 passed. - Rubber-duck review of the blocker fixes and of the merge resolution: no blocking issues.  Regression coverage added for both blockers:  - `LocalAiWslNetworkingConsentTests` proves that without consent the step returns `FailedTerminal` before applying anything: `ApplyMirroredNetworking` is never called, the `.wslconfig` bytes are unchanged, no backup is staged, and no shutdown is issued. A second case proves enabling Local AI alone does not satisfy the gate. - `LocalAiSetupUxContractTests` asserts the consent checkbox exists, that `WslMirroredNetworkingConsent` is never assigned from the enable toggle or a literal `true`, and that the Continue gate requires the checkbox. - `LocalAiPortLifecycleTests.Router_LaunchesRetiredQwen9BInstallAfterUpgrade` builds a manifest with the KV-precision fields stripped from disk, exactly as a pre-upgrade install would have been written, and proves it still resolves and launches. - `LocalAiPortLifecycleTests.Catalog_DoesNotOfferRetiredQwen9BForNewInstalls` proves the retired entry cannot re-enter selection. - `LocalAiPortLifecycleTests.Router_RejectsRetiredQwen9BReceiptWithUnsupportedProfile` proves compatibility did not become silent remapping.  ## Real behavior proof  **Consent gate, live WinUI at this head.** Setup was launched isolated on the consent step. Screenshot attached in the PR thread.  - Local AI on, consent unchecked: warning panel shown with the full prompt ("Local AI needs mirrored WSL networking. Setup will update your global .wslconfig and stop all running WSL distributions once. No distributions are deleted. Save any work in WSL first.") and the checkbox "I understand and allow this global WSL change and one-time shutdown." **"Install & set up" is disabled.** - Checking the box: accessibility tree diff shows `LocalAiNetworkingConsentCheckBox` `off -> on` and `PrimaryButton` `(disabled) -> enabled` in the same action. - Toggling Local AI off and back on: the checkbox returns to `off` and "Install & set up" is disabled again, so re-enabling never silently re-grants consent.  **Consent deny changes nothing.** The host `.wslconfig` SHA-256 was `5B5A9A4B...61542F0` with no running distributions before the run, and byte-identical with no running distributions after. The denied-consent path was exercised against a redirected temp config so the user's real global config was never a test subject.  **Selection and receipt behavior at this head**, from a standalone harness built against this branch after the merge:  ``` Offered models (3):   - qwen3.8-27b-mtp-ud-q4-k-m   priority=400  default=True   - qwen3.6-35b-a3b-mtp-q4-k-m  priority=300  default=False   - qwen3.6-27b-mtp-q4-k-m      priority=200  default=False Find(qwen3.5-9b) offered for new installs: False FindInstalled(qwen3.5-9b) resolvable for existing installs: True  RTX 5090 class (31.84 GiB): Eligible  model=qwen3.8-27b  profile=ctx-196608-q8_0  required=29.11 GiB 24 GiB class:               Eligible  model=qwen3.8-27b  profile=ctx-65536-f16    required=23.58 GiB 16 GiB class:               Unsupported (required 21.59 GiB)  Retired Qwen3.5 9B install, KV fields stripped from the on-disk manifest:   launch.ModelAlias     = qwen3.5-9b-mtp-q4-k-m   gateway primary model = llamacpp/qwen3.5-9b-mtp-q4-k-m   ctx-size = 262144 / cache-type-k = f16 / cache-type-v = f16   RESULT: existing Qwen3.5 9B install launches after upgrade.  Non-installable legacy profile:   Rejected: "The managed local AI context and KV cache receipt do not match a qualified catalog profile." ```  This reproduces the three capacity claims in the summary at the current head, after the #1253 merge.  Not verified / blocked:  - **Hardware blocker:** Qwen3.8 selection, download, launch, and fixed-prompt inference could not be recaptured at this head. The available validation host is a 16 GiB RTX 4080 SUPER, which this PR intentionally classifies as unsupported, so it cannot select or run the Qwen3.8 27B profile at all. This needs the `windows-wsl-dgx-blackwell` or RTX 5090 pool. - The earlier [Windows model-selection proof](https://github.com/openclaw/openclaw-windows-node/pull/1281#issuecomment-5515540905) was collected from integration head `38a172f1`. Its quoted WSL warning string is superseded: that concise message came from the removed-consent build, and the affirmative checkbox prompt shown above is what ships here.  ## Security and compatibility  - The user-wide `.wslconfig` mirrored-networking rewrite and the one-time global `wsl.exe --shutdown` require explicit affirmative consent. Enabling Local AI does not imply it, and setup cannot proceed without it when the change is needed. - Existing manifests referencing the retired Qwen3.5 9B model ID keep launching from their own pinned receipt. Model artifacts are never silently remapped, and the retired model can never be newly installed. - No new permissions or credential handling are introduced.


src/OpenClaw.Shared/Inference/Catalog/LocalModelCatalog.cs
```diff
@@ -107,19 +92,69 @@ public sealed record LocalModelInfo(
...
+}
 
 /// <summary>Immutable Hugging Face model pins offered by the Windows local inference flow.</summary>
 public static class LocalModelCatalog
 {
+    public const string Qwen38_27BModelId = "qwen3.8-27b-mtp-ud-q4-k-m";
     public const string Qwen35BModelId = "qwen3.6-35b-a3b-mtp-q4-k-m";
     public const string Qwen27BModelId = "qwen3.6-27b-mtp-q4-k-m";
     public const string Qwen9BModelId = "qwen3.5-9b-mtp-q4-k-m";
     public const int NativeContextTokens = 262_144;
```
larroy, line 145: can we refactor this and name it with t-shirt size: Context Large, Medium, Small, XS. Better with an enum or an array instead of defining all those variables explicitly. This suits better if we need to use a selector or at least makes the code easier to read by explicitly declaring the options and having a map to the context size length.

joelagnel: This is tricky to do. Weaker model might have larger context or higher parameters count. I prefer something like Model name - version number - parameter count. Let's look into this after initial merge as product team might have a say on the specific strings for this and other similar strings, and I don't want to gate merging this last minute. Thanks

joelagnel:
@clawsweeper re-review

joelagnel:
Joel: will add proof

joelagnel:
## Windows model-selection proof

Captured on Windows 11 ARM64 with an NVIDIA RTX Spark N1X during the validated Local AI integration run.

Local AI install review showing Qwen 3.8 27B recommended

The screen shows:

- Local AI enabled.
- 45.4 GiB CUDA-visible memory.
- 40.3 GiB required by the selected profile.
- Qwen 3.8 27B selected and marked Recommended.
- A 16.5 GB model artifact.

During the same run, fixed-prompt inference completed with 32 generated tokens and GPU verification passed with 66/66 layers and 15,738,286,899 bytes in CUDA model buffers.

joelagnel:
@clawsweeper re-review

PR #1334, opened 2026-09-05
fix(setup): remove post-install gateway port ownership heuristic

Elliot-Construct:
## Summary

Remove the post-install `ss -tlnp` process-name ownership heuristic from `StartGatewayStep`.

`openclaw gateway install --force` can already start/restart the managed gateway service. The subsequent start step should therefore use the idempotent `openclaw gateway start` path and existing health polling instead of rejecting a listener purely because `ss` does not expose a process name.

This avoids false `Port 18789 is already in use by another process` failures when the listener is the gateway service just installed by setup.

## Regression

The first commit intentionally adds a failing regression test only. It asserts that normal startup does not run the `ss -tlnp` ownership probe and still invokes `openclaw gateway start` followed by health checking.

## Intended lifecycle

`install -> gateway start -> health check`

This supersedes the narrower process-name workaround discussed in #547, because `ss` may expose `node`, `openclaw`, or no process identity at all.

### DIFF
--- a/src/OpenClaw.SetupEngine/StartGatewayStep.cs
+++ b/src/OpenClaw.SetupEngine/StartGatewayStep.cs
@@ -33,25 +33,6 @@ public sealed class StartGatewayStep : SetupStep
         var pathCmd = ctx.WslPathPrefix;
         var action = restart ? "restart" : "start";
 
-        if (!restart)
-        {
-            var portCheck = await ctx.Commands.RunInWslAsync(
-                distro, $"ss -tlnp 2>/dev/null | grep ':{ctx.Config.GatewayPort}\\b' || true",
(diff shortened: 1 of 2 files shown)


Elliot-Construct:
Also I was messaging one of the maintainers in the contributor channel in discord and all of a sudden I got kicked out of the server? Not sure why? I was specifically answering questions he had. @ngutman

karkarl:
Closing as superseded. #1349 preserved the diagnosis and replaced this PR's removal of ownership verification with a safer PID-based service ownership check. #1349 has now been superseded by #1382, which transplants that fix onto current main with strengthened diagnostics, coverage, and real setup E2E proof.

Issue #1413, opened 2026-09-14
Node Sandbox is unreachable on stable Windows: #1327 refuses the appcontainer-dacl tier that #784/#786/#787 validated, and the bundled wxc-exec has no BFS compiled in


yjw0817:
## Summary

On a fully updated Windows 11 25H2 machine, Node Sandbox cannot be turned on. The UI refuses it:

> **Node Sandbox unavailable** — Node Sandbox can't be turned on because this PC does not provide MXC BaseContainer without host DACL augmentation. This PC reports `appcontainer-dacl`. Host DACL augmentation is required.
>
> Agent-started commands will keep using the host execution path until MXC is available.

So agent-started Windows commands run **uncontained on the host**, and the user has no way to change that.

What makes this worth filing: the `appcontainer-dacl` tier was implemented, runtime-proven end to end on this exact build family, and merged — #784 / #786 / #787 — and then #1327 narrowed the admission policy to BaseContainer-only, which makes that merged work unreachable on every Windows client below build 26600.

## Environment

| | |
|---|---|
| Windows Hub | `2026.7.1-4` (latest release) |
| OS | Windows 11 Pro 25H2, build **26200.8037**, Client, fully updated |
| CPU | Intel Core i7-7700 |
| Bundled MXC | `%LOCALAPPDATA%\OpenClawTray\tools\mxc\x64\wxc-exec.exe`, version **0.8.0+7dac1a9** |
| `C:\Windows\System32\processmodel.dll` | **does not exist** |

## Probe output

`wxc-exec.exe --probe`:

```json
{
  "tier": "appcontainer-dacl",
  "needsDaclAugmentation": true,
  "warnings": [
    "BaseContainer tier not selected, and AppContainer + BFS is not compiled into this binary; falling back to AppContainer + DACL",
    "AppContainer + DACL tier selected: AppContainer processes may be unable to read metadata of the system-drive root (e.g. `cmd.exe`, `pwsh.exe`, `node.exe` startup stats of `C:\\`). Run `wxc-host-prep prepare-system-drive` (elevated) to grant the minimal metadata ACEs.",
    "AppContainer + DACL tier selected: AppContainer processes may be unable to open the NUL device (`\\Device\\Null`), which the kernel resets to an AppContainer-hostile default at every boot. Run `wxc-host-prep prepare-null-device` (elevated) to reapply the documented security descriptor."
  ],
  "probes": {
    "baseContainerApiPresent": false,
[...]

PR #1415, opened 2026-09-14
fix(sandbox): explain unsupported Windows containment

yjw0817:
Related: #1413

## What Problem This Solves

When the MXC probe reports a real but unsupported tier such as `appcontainer-dacl`, the Sandbox page should state the actual result: Node Sandbox is unavailable on this Windows build, and the user should check Windows Update.

Previously, that probe result had `IsAppContainerAvailable=true`, so the page did not classify it as a Windows containment limitation even though `CanRunSystemRunSandbox=false`. The rejection dialog used lower-level BaseContainer and DACL wording without the existing Windows Update guidance.

## Why This Change Was Made

OpenClaw admits `system.run` containment only when MXC reports BaseContainer without host DACL augmentation. Other tiers remain unavailable by design.

This patch uses the authoritative `CanRunSystemRunSandbox` result when classifying a resolvable, non-errored probe. The unavailable InfoBar and the rejection dialog now reuse the existing localized Windows-version message and Windows Update guidance.

The patch does not change tier admission, saved preferences, fallback behavior, strict blocking, or command execution.

## User Impact

On a host reporting `appcontainer-dacl` or another non-admitted tier:

- Node Sandbox remains off and unavailable.
- The page says the Windows version does not support sandboxing yet.
- The warning states that commands run on the host without sandbox protection.
- The user is told to install the latest Windows updates, with Windows Insider builds mentioned as an optional path to newer capabilities.

## Evidence

Current-head isolated app with a deterministic `appcontainer-dacl` probe result:

Node Sandbox unavailable dialog explaining that the Windows version does not support sandboxing yet and recommending Windows Update

The inspected live dialog text was:

```text
Your Windows version doesn't support sandboxing yet

OpenClaw Node Sandbox requires MXC BaseContainer without host DACL augmentation.
This PC reports appcontainer-dacl. Host DACL augmentation is required.

Commands will run on the host without sandbox protection while sandboxing is unavailable. Sandboxing requires a recent Windows build with the AppContainer primitives shipped. Install the latest Windows updates (or join the Windows Insider Program for the newest builds) to enable containment.
```

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Docs or instructions
- [x] Tests or validation
- [ ] Security hardening
- [ ] Chore or infrastructure

## Scope

- [x] Tray or WinUI UX
- [ ] Windows node capability
- [ ] Local MCP or `winnode`
- [ ] Gateway, connection, or pairing
- [ ] Setup or onboarding
- [x] Tests, CI, or docs

## Required proof pools

- `windows-winui-interactive`: the Sandbox unavailable state and rejection dialog changed.

## Validation

Exact maintainer patch commit: `234ad9e7`.

| Command | Result |
|---|---|
| `.\build.ps1` | Passed |
| `dotnet test .\tests\OpenClaw.Shared.Tests\OpenClaw.Shared.Tests.csproj --no-restore` | 4,000 passed, 32 skipped, 0 failed |
| `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore` | 2,976 passed, 0 failed |
| `dotnet test .\tests\OpenClaw.Tray.Tests\OpenClaw.Tray.Tests.csproj --no-restore --filter FullyQualifiedName~SandboxPage` | 3 passed, 0 failed |
| `.\scripts\validate-mxc-e2e.ps1` | Blocked before containment testing. WSL failed to attach the fresh isolated `ext4.vhdx` with `Access is denied`; 6 dependent tests failed and teardown completed. No MXC pass is claimed. |
| `git diff --check` | Passed |

One initial full Shared run hit `McpHttpServerTests.Dispose_DuringInFlightHandler_DoesNotSurfaceObjectDisposedException`. The exact test passed immediately on retry, and the complete required Shared suite then passed.

## Real Behavior Proof

- Environment: native Windows x64 Debug app, isolated tray data.
- Probe fixture: `tier=appcontainer-dacl`, `needsDaclAugmentation=true`, emitted by a temporary deterministic test executable through the documented `OPENCLAW_WXC_EXEC` override. The fixture was not committed.
- Exact flow: launched the current build directly to `openclaw://hub/sandbox --post-setup-restart`, inspected the page with Microsoft WinApp CLI, and toggled Sandbox on.
- Observed page state: `Node Sandbox is off: host execution`; unavailable InfoBar title `Your Windows version doesn't support sandboxing yet`; controls disabled.
- Observed dialog state: the same Windows-version title, the DACL probe reason, explicit uncontained host fallback, and Windows Update guidance.
- Screenshot link verified: uploaded successfully to GitHub and embedded above.
- Not verified / blocked: real Gateway to Windows node MXC execution was not reached because the isolated WSL distro VHDX could not attach on this machine. Execution behavior is unchanged by this patch.

## Security Impact

- New permissions or capabilities? No
- Secrets or tokens handling changed? No
- New or changed network calls? No
- Command or tool execution surface changed? No
- Data access scope changed? No

- Backward compatible? Yes
- Config or environment changes? No
- Migration needed? No

## Review Conversations

- [ ] I replied to or resolved every bot review conversation addressed by this PR.
- [ ] I left unresolved only conversations that still need maintainer judgment.


yjw0817:
Before (released 2026.7.1-4): toggle off, intent discarded
image

After (this branch): toggle stays on, "unavailable: host fallback"
image

yjw0817:
Thanks, both corrections are right and I have fixed the PR rather than argued with them. New head: `b4d5d29`. The PR body is updated accordingly.

## 1. "Execution behavior is unchanged" was wrong

Confirmed in `src/OpenClaw.Shared/Mxc/MxcCommandRunner.cs`: the custom-env denial sits **before** the availability check, so it only applies while the preference is on.

```csharp
if (!settings.SystemRunSandboxEnabled) { ... return await RunHostFallbackAsync(...); }

// Custom env changes the execution boundary. Until MXC can enforce it
// in-container, sandbox-enabled requests must not bypass policy through
// the MXC-unavailable compatibility fallback.
if (request.Env is { Count: > 0 }) return DenyCustomEnvUnsupported();

if (!_isSandboxAvailable()) { ...host fallback... }
```

So on an unavailable host, `system.run` with custom environment variables is now refused where a page visit previously made it run uncontained. That is a real behavior change and the body should not have claimed otherwise.

I would argue it is the correct direction: that comment states the guard's purpose, and the old normalization was the thing defeating it, by turning the preference off so the request never reached the check. Restoring the guard is the point rather than a side effect. But it is a compatibility change, so it belongs in the maintainer decision, not in a "no runtime impact" claim. It is now stated in the PR body, and the new dialog tells the user directly: "Requests that set custom environment variables are refused while the sandbox is on."

## 2. "Users can turn the preference back on" was wrong, and is now true

You are right that `RejectSandboxEnableWhenUnavailableAsync` made already-off profiles unrecoverable on an unavailable host, which left the fix half-done: new profiles keep intent, existing victims stay stuck.

`b4d5d29` addresses that. Turning the sandbox on while containment is unavailable now saves the preference and then explains the limitation, instead of reverting the toggle. Nothing else moves: the page keeps the unavailable status card and InfoBar, sub-controls stay dimmed because `UpdateControlsEnabledState()` gates on availability, and `SystemRunBlockHostFallbackWhenMxcUnavailable` keeps its fail-closed path untouched.

Recovery proof, isolated profile seeded with `"SystemRunSandboxEnabled": false` to simulate a profile the old behavior turned off:

1. Open This Computer, Sandbox. Turn the toggle on.
2. Dialog: "Node Sandbox is on, but this PC cannot contain commands yet".
3. Close the app and read `settings.json`: `"SystemRunSandboxEnabled": true`.

The page then shows "Node Sandbox unavailable: host fallback" with the toggle on and the red InfoBar open, which is the honest state: the preference is the user's, the capability is the host's.

## 3. The UX contract reversal is yours to decide

Noted that #825 chose normalization deliberately, to avoid implying containment was active. I am not claiming that concern was wrong. The difference is where it is expressed: with this PR the page still says containment is unavailable in the status card, the InfoBar, and now a dialog, but it says it without editing the user's stored preference. If you would rather keep the old contract, the right call is to close this PR rather than land a weaker compromise, and I would still suggest the persisted write be dropped even if the toggle keeps rendering off.

## 4. Validation and the mxc pool

Updated in the body. Two failures, both environmental, both stated rather than hidden:

- `Shared.Tests` `Mxc.MxcExecutorTests.RunAsync_DecodesFragmentedPreamblesWithoutOverflow` reproduces on a clean tree with this branch stashed.
- `Tray.Tests` `NativeSpeechStackRuntimeTests.TrayBuildOutput_NativeTtsStack_LoadsWithAppLocalVCRuntime` fails with its own remediation message: the VS "C++ Redistributable Update" component is not installed on this host, so the build copied the older fallback runtime. Not touched by this change.

On `windows-wsl-mxc`: I have kept the declaration and I am reporting it blocked rather than substituting adjacent coverage. This host probes `tier: appcontainer-dacl`, so `scripts/validate-mxc-e2e.ps1` cannot produce a real containment run here, and per AGENTS.md an `-AllowSkip` run must not be reported as merge validation. The containment path itself is untouched by this PR; what changes is which preference value reaches `MxcCommandRunner`, which the custom-env note above describes exactly.

yjw0817:
Recovery on this branch: profile seeded with SystemRunSandboxEnabled=false,
turned the toggle on, preference saved as true.
image

yjw0817:
b4d5d29 recovery dialog: turning the sandbox on while containment is
unavailable saves the preference and explains, instead of refusing.
image

shanselman:
I pushed a small maintainer patch rather than replacing your PR. It keeps the existing Sandbox-off and host-fallback contract, then narrows the fix to the #1413 case: any resolvable MXC tier that OpenClaw cannot admit, including appcontainer-dacl, now receives the existing Windows-version and Windows Update guidance in both the InfoBar and rejection dialog. I also updated the PR title/body and attached current-head isolated UI proof. Thanks for identifying the confusing unavailable behavior.\n\n@clawsweeper re-review

shanselman (approved):
The maintainer patch narrows this to the intended UX correction: non-admitted but resolvable MXC tiers remain unavailable and receive clear Windows Update guidance. Current-head UI proof and CI are sufficient.

yjw0817:
Thanks for taking the time to patch the branch instead of replacing it — and the narrower scope is the right call. Keeping the Sandbox-off and host-fallback contract while fixing only the misleading explanation is a cleaner fix than what I proposed.

If a BaseContainer-capable build ever needs testing on a pre-26600 host, I still have this machine set up for it (25H2 26200.8037, probing appcontainer-dacl) and I'm happy to run whatever would be useful.

